PR #44, opened 2026-05-02
feat(python): add contract decorator, pydantic lift, verifier, lsp plugin

TSavo:
## Summary

Adds four new modules to the Python kit, making it a complete v1.1 implementation.

### provekit.contract decorator
- Supports `@contract(pre="x >= 0", post="out >= 0")` string expressions
- Supports `@contract(pre=lambda x: x >= 0)` callable predicates with runtime checking
- `collect_module()` gathers all decorated functions
- `ContractViolation` exception raised on runtime failures

### provekit.lift.pydantic
- Walks Pydantic v2 BaseModel field annotations
- Lifts Field(ge=0, le=100, min_length=1, pattern=...) constraints
- Lifts Annotated[..., Gt(0), Lt(100)] annotated-types
- Emits same canonical IR as Bean Validation Min/Max/Pattern

### provekit.verifier
- Delegates to Rust provekit verify CLI via subprocess
- Returns structured HandshakeReport with tier discharge fractions
- VerifierNotFoundError when CLI not on PATH

### provekit.lsp
- NDJSON LSP plugin protocol
- initialize, parse, shutdown methods
- Lifts Layer 2 + Pydantic contracts from Python source

### Tests
- 7 tests for decorators and pydantic lift
- All 55 existing kit tests continue to pass

### DIFF
--- a/implementations/python/provekit-lift-py-tests/src/provekit_lift_py_tests/decorators.py
+++ b/implementations/python/provekit-lift-py-tests/src/provekit_lift_py_tests/decorators.py
@@ -1,0 +1,328 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# provekit.decorators — direct contract authoring for Python.
+#
+# Usage:
+#   @provekit.contract(pre=lambda x: x >= 0, post=lambda out: out >= 0)
+#   def abs(x: int) -> int:
+#       return x if x >= 0 else -x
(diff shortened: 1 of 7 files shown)


TSavo:
Superseded by PR #33 (Python extensions landed via agent in parallel session). Same files.

PR #60, opened 2026-05-02
feat(go): add validator lift adapter + LSP plugin

TSavo:
## Summary

Two additions to the Go kit:

1. **provekit-lift-go-validator** — lift adapter for go-playground/validator struct tags
2. **provekit-lsp-go** — NDJSON LSP plugin binary

## provekit-lift-go-validator

`LiftStruct(v interface{})` walks struct fields via `reflect`, parses `validate` tags, and returns `ContractDeclaration` values with canonical IR preconditions.

### Tag → IR mappings

| Tag | IR |
|-----|-----|
| `required` | `neq(var, "")` for strings, `neq(var, 0)` for numerics |
| `gte=N`, `lte=N` | `gte(var, N)`, `lte(var, N)` |
| `gt=N`, `lt=N` | `gt(var, N)`, `lt(var, N)` |
| `eq=N`, `ne=N` | `eq(var, N)`, `neq(var, N)` |
| `min=N`, `max=N` | numeric → `gte`/`lte`; string → `gte`/`lte(strlen(var), N)` |
| `len=N` | `eq(strlen(var), N)` |
| `email` | placeholder (true atom) |
| `oneof=X Y Z` | `or(eq(var, X), eq(var, Y), eq(var, Z))` |

### Cross-domain equivalence

- `gte=0,lte=100` on int → same IR structure as `@Min(0) @Max(100)` (Java) and `Field(ge=0, le=100)` (Python/pydantic)
- `required` on string → same `neq` atom as `@NotNull` (Java)
- `min=1` on string → `gte(String.prototype.length(var), 1)` ctor pattern

### Tests (7 passing)

- Struct walking and declaration count
- Range constraint byte-for-byte JCS verification
- String required non-null equivalence
- min/max string length mapping
- Empty/untagged struct edge cases
- oneof disjunction

## provekit-lsp-go

NDJSON JSON-RPC LSP plugin over stdin/stdout.

### Protocol
```
→ {"jsonrpc":"2.0","id":1,"method":"initialize"}
← {"result":{"name":"provekit-lsp-go","version":"0.1.0","capabilities":["parse"]}}

→ {"jsonrpc":"2.0","id":2,"method":"parse","params":{"path":"...","source":"..."}}
← {"result":{"declarations":[...],"warnings":[]}}

→ {"jsonrpc":"2.0","id":3,"method":"shutdown"}
← {"result":null}
```

### Parse handler
- Walks Go source via `go/parser` + `go/ast`
- Lifts `//provekit:contract` and `//provekit:implement <cid>` annotations
- Lifts validator struct tags via `reflect.StructTag.Lookup("validate")`
- Emits JCS declarations JSON

### Verified
```
→ type User struct { Name string `validate:"required,min=1"` }
← {"contract","name":"User.Name","pre":{"kind":"and","operands":[
     {"kind":"atomic","name":"≠"...},
     {"kind":"atomic","name":"≥","args":[{"ctor","name":"String.prototype.length"...}]}
   ]}}
```

### DIFF
--- a/implementations/go/.gitignore
+++ b/implementations/go/.gitignore
@@ -1,0 +1,1 @@
+provekit-lsp-go
--- a/implementations/go/cmd/provekit-lsp-go/main.go
+++ b/implementations/go/cmd/provekit-lsp-go/main.go
@@ -1,0 +1,427 @@
+// provekit-lsp-go — NDJSON LSP plugin for Go.
+//
+// Protocol:
+//
+//	{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}
+//	{"jsonrpc":"2.0","id":2,"method":"parse","params":{"path":"...","source":"..."}}
+//	{"jsonrpc":"2.0","id":3,"method":"shutdown"}
+//
+// For parse, scans the source for //provekit: annotations and
+// go-playground/validator struct tags, lifts to canonical IR,
+// and returns JCS declarations JSON.
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"reflect"
+	"strings"
+
+	ir "github.com/tsavo/provekit/go/provekit-ir-symbolic/ir"
+)
+
+type rpcRequest struct {
+	JSONRPC string          `json:"jsonrpc"`
+	ID      interface{}     `json:"id"`
+	Method  string          `json:"method"`
+	Params  json.RawMessage `json:"params"`
+}
+
+type parseParams struct {
+	Path   string `json:"path"`
+	Source string `json:"source"`
+}
+
+type rpcResponse struct {
+	JSONRPC string      `json:"jsonrpc"`
+	ID      interface{} `json:"id"`
+	Result  interface{} `json:"result,omitempty"`
+	Error   *rpcError   `json:"error,omitempty"`
+}
+
+type rpcError struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
+type initializeResult struct {
+	Name         string   `json:"name"`
+	Version      string   `json:"version"`
+	Capabilities []string `json:"capabilities"`
+}
+
+type parseResult struct {
+	Declarations json.RawMessage `json:"declarations"`
+	Warnings     []interface{}   `json:"warnings"`
+}
+
+func main() {
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		line := scanner.Bytes()
+		var req rpcRequest
+		if err := json.Unmarshal(line, &req); err != nil {
+			continue
+		}
+
+		switch req.Method {
+		case "initialize":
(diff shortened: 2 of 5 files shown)


TSavo (commented):
## Review

Substantive PR. Real Go validator-tag lifter + Go LSP plugin. Both pieces fit the framework's pluralism story cleanly.

### Strengths

1. **Validator tags lift to canonical IR.** The test `TestLiftStruct_RangeConstraintByteEquivalent` literally pins this: a Go `validate:"gte=0,lte=100"` produces the same IR atom as Bean Validation `@Min(0)@Max(100)` and Pydantic `Field(ge=0, le=100)`. That's the hash-preserving-translation thesis (`docs/launch/hash-preserving-translation.md`) in action.
2. **Tag coverage is good first-pass:** required, gte/lte/gt/lt/eq/ne, min, max, len, oneof.
3. **LSP plugin** implements initialize/parse/shutdown over NDJSON, slotting cleanly into the lift-plugin-protocol seam.

### Concerns

1. **`email` returns `ir.And()` (empty conjunction = always-true).**
   ```go
   if tag == "email" {
       return ir.And() // true placeholder
   }
   ```
   This is semantically wrong. A field with `validate:"email"` that gets lifted as "always true" will be silently *accepted* by the verifier when it should be either (a) lifted to a real pattern predicate, or (b) flagged as opaque so the verifier refuses with a `kit_predicate_no_semantics` reason code (per the v1.4.0 opacity-manifest grammar in `protocol/specs/2026-05-02-opacity-manifest-grammar.md`). Returning vacuous-true is the worst option — it lies to callers.

2. **Code duplication.** `validator.go` and `cmd/provekit-lsp-go/main.go` both define `liftValidateTag` (with the comment "Duplicated from validator.go for AST-mode operation"). Factor into a shared helper before merging — this is a future drift hazard.

3. **JCS alpha-sort caveat in the test comment.**
   > Go kit uses logical key order (kind, name, args); JCS-alpha sort is tracked separately in the Go kit canonicalizer.

   That's fine for the test's scope (it's verifying the lifter's structural output), but the cross-impl conformance gate (`conformance/run.sh`) compares **canonicalized** bytes. The expected string in this test is NOT what the conformance harness will see. Worth clarifying which layer is the equivalence claim.

4. **Missing common tag: `regex=`.** Probably the next-most-used after the ones you covered. Can be a follow-up.

5. **No LSP plugin test** for `cmd/provekit-lsp-go/main.go`. The 427 LOC speak NDJSON JSON-RPC; should have at least one round-trip test (initialize → parse → shutdown) like the existing `provekit-lift-go-tests` has.

### Recommendation

Land the lifter half cleanly. For the `email` issue, either:
- (a) lift to a real regex predicate against an email pattern, or
- (b) refuse and emit an opacity-manifest entry with `reason: kit_predicate_no_semantics` so the v1.4.0 ConsensusCoverage admission gate (#203, in flight) can see it as needing Coq.

Option (b) is structurally correct: this kit doesn't have semantics for `email`, so it shouldn't pretend it does.

Either way, this is good work and the bones are right.

PR #70, opened 2026-05-02
docs(internal): agent-worktree path resolution bleed audit

TSavo:
## Summary

- Root-cause analysis of the agent-worktree absolute-path resolution bleed flagged during PR #18's Go cross-impl conformance dispatch
- Documents the failure mode: Edit/Write tool calls resolve absolute paths to the maintainer's main worktree instead of the agent's isolated git worktree
- Reviews existing partial mitigation in `captureChange.ts` (commit `9d9f9d5`) and notes it covers only the fix-loop C3/C6 path
- Evaluates four structural fix options and recommends option (a): CWD invariant enforcement at the agent dispatch boundary
- No code changes -- audit and recommendation only

### DIFF
--- a/docs/internal/agent-worktree-path-bleed-audit.md
+++ b/docs/internal/agent-worktree-path-bleed-audit.md
@@ -1,0 +1,264 @@
+# Agent worktree path resolution bleed: audit
+
+**Status:** open (no fix landed)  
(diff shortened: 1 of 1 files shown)


TSavo:
Closed via https://github.com/TSavo/provekit/pull/949.

PR #111, opened 2026-05-03
spec(bridges): target dimensionality addendum

TSavo:
## Summary

Normative addendum to `2026-04-30-ir-formal-grammar.md` §BridgeDeclaration that resolves the target-dimensionality ambiguity in the bridge IR.

### The four problems

- `targetContractCid` is overloaded across kits: some use it for a contractCid (1D), some for a contractSetCid (set-D), some put placeholder strings like `pending-csharp-counterpart:<name>` or `deferred:phase-3-proof-bundle` in it. Different claims in the same field.
- Placeholder strings are anti-substrate: they put non-CID values in a substrate-verified header field, violating the closure property (manifesto §10) and the address-is-multi-dimensional principle (§8).
- The three-axis pin (contract + witness + binary, manifesto §8) has no normative bridge expression. `targetProofCid` conflates the witness and binary axes without naming them separately.
- The substrate-vs-metadata cut from spec #95 is not enforced: `targetLayer`, `targetProofCid`, and related fields sit flat in the header alongside the contract-axis claim.

### The five normative rules

- **R1**: `target` is a tagged union `{kind, cid}`. `kind: "contract"` for a single contractCid; `kind: "contractSet"` for a contractSetCid. Never a raw string. Never a placeholder.
- **R2**: Placeholder strings (`pending-*:`, `deferred:*`) are banned from any substrate-verified field. Absent axes are omitted, not stringified.
- **R3**: Bridge header carries the contract-axis claim only (`sourceContractCid`, `target`). Witness and binary axes move to body metadata (`targetWitnessCid`, `targetBinaryCid`), per spec #95.
- **R4**: The three-axis pin belongs on the consumer's attestation, not the bridge. This relocates forward-pin closure from `BridgeDeclaration.ConsequentBundlePinned` to the consumer's own signed claim. Substitution risk is preserved; the closure point shifts to the consumer.
- **R5**: Optional `targetContractSetCid` in body for downstream set walks, per spec #94.

### Migration notes

- Existing bridges with placeholder values are non-normative under this spec; must be re-emitted before any conformance gate.
- The 9-field locked key order is a version boundary: bridges before this spec use the old flat shape; bridges after use the layered shape. The two shapes must not be mixed in the same slab. Kits must re-pin golden bytes after migration.

### What this does NOT do

- Does not implement in any kit (follow-up PR per kit).
- Does not update the protocol catalog (follow-up, bundled with implementation PRs).
- Does not retroactively invalidate existing bridges (they remain valid as historical artifacts per §2).

## Test plan

- [ ] Catalog verify passes (`cargo run --release --manifest-path tools/recompute-spec-cids/Cargo.toml -- --verify`): confirmed passing, new spec is not yet in catalog as expected.
- [ ] Read spec end-to-end for internal consistency: done.
- [ ] Conformance tests specified in §3 of the addendum; implementation PRs will enforce them.

> [!NOTE]
> DRAFT: do not merge without Sir's review. This is a normative spec change that retires `BridgeDeclaration.ConsequentBundlePinned` and changes the locked key order contract.

### DIFF
--- a/protocol/specs/2026-05-03-bridge-target-dimensionality.md
+++ b/protocol/specs/2026-05-03-bridge-target-dimensionality.md
@@ -1,0 +1,124 @@
+# Bridge Target Dimensionality
+
+**Status:** v1.1.0 normative addendum to `2026-04-30-ir-formal-grammar.md` §BridgeDeclaration
+**Date:** 2026-05-03
+
+## §0. Motivation
+
+Four problems in the current bridge IR shape motivate this addendum.
+
+**Problem 1: target dimensionality is silently overloaded.** Today `targetContractCid` carries one of three distinct claims depending on the kit: a contractCid naming one specific contract (1D), a contractSetCid naming a whole set of contracts (set-D), or a string placeholder (`pending-csharp-counterpart:<name>`, `deferred:phase-3-proof-bundle`) that names nothing. These are different claims pretending to be the same field.
+
+**Problem 2: placeholder strings are anti-substrate.** Strings like `pending-*:` and `deferred:*` are refusals to address. They violate the substrate's closure property (manifesto §10): every CID in a header must be computable from content alone. A placeholder string tells the substrate "I am pinning something" while pinning nothing. The field name says `contractCid`; the value is not a CID.
+
+**Problem 3: the three-axis pin from manifesto §8 is not formalized for bridges.** The manifesto §8 names `contractCid`, `witnessCid`, and `binaryCid` as three orthogonal pinning axes. Spec `2026-05-03-version-chains-pinning.md` makes this the basis of package-manager replacement. Bridges have no normative way to express "I am pinning a contract AND a witness AND a binary." They have `targetContractCid` (one axis, sometimes wrongly occupied) and a half-implemented `targetProofCid` slot that conflates the witness and binary axes.
+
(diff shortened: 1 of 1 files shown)


TSavo:
Five-bullet summary of normative rules:

1. **R1 (tagged-union target)**: `target` is now `{kind, cid}` — `kind: "contract"` for a single contractCid, `kind: "contractSet"` for a contractSetCid. Raw string values and placeholders are banned.
2. **R2 (no placeholders)**: `pending-*:` and `deferred:*` strings MUST NOT appear in any substrate-verified bridge field. Absent axes are omitted from the body, not stringified.
3. **R3 (header/body cut)**: Bridge header carries contract-axis claim only (`sourceContractCid` + `target`). Witness, binary, and layer fields move to body metadata per spec #95.
4. **R4 (consumer owns the three-axis pin)**: The bridge does not promise the three-axis composition. A consumer asserting contract + witness + binary reads the three CIDs from the bridge and signs its own attestation. This retires `BridgeDeclaration.ConsequentBundlePinned` from ir-formal-grammar; substitution protection relocates to the consumer's attestation.
5. **R5 (optional set snapshot)**: Bridges MAY carry `targetContractSetCid` in the body as a snapshot of the target's containing contract set at mint time, for downstream DAG walks per spec #94.

PR #142, opened 2026-05-03
chore(make): add build-all target

TSavo:
## Summary

Adds \`make build-all\` as a convenience wrapper that aggregates the per-kit build targets:

\`\`\`
build-all: build-rust build-cpp build-go build-ts build-csharp
\`\`\`

Useful before \`make conformance\` or before spawning \`provekit-linkerd\` (which subprocesses kit lifters at lift time and needs each kit's binaries on PATH or in the workspace target tree).

The per-kit targets remain independent so failures stay isolated; \`build-all\` is a convenience, not a coupling.

## Test plan

- [ ] \`make help\` shows the new target under "Per-language build"
- [ ] \`make build-all\` invokes all five kit builds in sequence

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,7 @@ help:
 	@echo "  make test-all       run all language-native test suites"
 	@echo ""
 	@echo "Per-language build:"
+	@echo "  make build-all      build every kit (rust + cpp + go + ts + csharp)"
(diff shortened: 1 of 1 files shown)


ConstCorrectness:
Amazing project! *10/10*

PR #326, opened 2026-05-04
feat(php): add Bridge v1.4 mint (mintBridgeV14) to PHP Minter

TSavo:
## Summary

Adds v1.4 layered bridge mint to the PHP kit's `ClaimEnvelope\Minter`. The PHP kit already had `mintContract` and `mintBridge` (v1.1 flat), but was missing the v1.4 layered `{envelope, header, metadata}` shape with tagged-union target.

## What changed

`implementations/php/provekit-ir-symbolic/src/ClaimEnvelope/Minter.php` — added `mintBridgeV14(array $args): array`

- **Header**: 7 canonical fields per spec §1.R3
- **Metadata**: optional axes, omitted when empty per §1.R2
- **Target**: tagged union `{kind: "contract"|"contractSet", cid}`
- **Signature**: Ed25519 over `JCS({header, metadata})`
- **CID**: BLAKE3-512(JCS(memento))

Coexists with the existing v1.1 `mintBridge`. No breaking changes.

## Summary by CodeRabbit

* **New Features**
  * Added Bridge v1.4 envelope minting support across multiple language implementations.
  * Added PHP kit support and a mint target for producing PHP self-contract attestations.
  * Added a CLI workflow to mint and publish per-language baseline proofs.

* **Tests**
  * Added cross-language conformance and round-trip tests for Bridge v1.4.

* **Chores**
  * Generated baseline proof artifacts for many standard libraries (multiple languages).


TSavo:
CI failure: `mint-csharp` Makefile target failed.

```
mint-csharp:
  implementations/rust/target/release/provekit mint --kit=csharp
make: *** [Makefile:241: mint-csharp] Error 1
```

The csharp kit's self-contracts attestation directory does not exist on main:
```
$ ls implementations/csharp/.provekit/
lift
config.toml
```

(rust is the only kit with `self-contracts-attestations/` populated.)

Fix path: this PR adds csharp to all-mint, so it needs to also commit the freshly-minted csharp self-contracts attestation alongside. Likely:
1. Rebase on main
2. `make build-rust && implementations/rust/target/release/provekit mint --kit=csharp`
3. `git add implementations/csharp/.provekit/self-contracts-attestations/`
4. Commit + push

Per autonomous-fire rules I'm leaving this alone. Author: re-mint csharp and add the attestation, then this PR can land.

TSavo:
Confirmed root cause: `implementations/csharp/Provekit.IR/BridgeV14.cs:31` doesn't compile.

```
error CS0546: 'BridgeTargetV14.Contract.Cid.init': cannot override because
  'BridgeTargetV14.Cid' does not have an overridable set accessor
```

This file is added by your PR (97 lines, ADDED, not on main). The C# inheritance: `BridgeTargetV14.Contract` declares `Cid { init; }` as an override, but the parent class `BridgeTargetV14` defines `Cid` without `{ init; }` (missing virtual or init accessor on parent).

Fix options:
1. Add `{ init; }` (or `{ get; init; }`) to the parent `BridgeTargetV14.Cid` declaration so subclasses can override it.
2. Make the parent's `Cid` virtual and override in the subclass with matching accessors.
3. Remove the override and use `new` keyword if the subclass property is intentionally distinct.

Once the csharp lifter compiles, mint-csharp produces a non-empty contractSetCid and the attestation gate passes (verified locally: a clean main mint produces `blake3-512:ed7f1eb7...` matching the on-disk pin).

Per autonomous-fire skip rules I'm not pushing fixes to your branch. Leaving for you.

TSavo:
Closing per architect call. Both #293 and #326 carry substantial substrate-touching review concerns concentrated in `mint_all_baselines.rs` (1 Critical + several Majors from Copilot/CodeRabbit/Codex: incorrect `vec_push` return type, ruby arity-2 quantification, duplicate catalog name, unreachable `ctor0(sig)`, silent error swallow, hard-coded language list omitting rust, etc.).

Per Supra omnia, rectum: shipping these as-is is anti-mission. The right move is to restart the baseline work as a focused PR with the concerns resolved inline — not as iteration debt.

#293 and #326 closed together; baseline catalogs tracked as a fresh effort.

PR #365, opened 2026-05-05
docs(papers): paper 07 — After Verification: Bug Classes as Missing Edges

TSavo:
## Summary

Paper 07 is the deeper consequence of After Reputation. Once droppers close the loop with lifters over weakest-precondition propagation, **leaf-discharge bug classes become structurally impossible.**

## What this paper establishes

- **Theorem (Structural Elimination of Leaf-Discharge Bug Classes)** — proved constructively by induction on data-flow path length over the cached DAG.
- **Algebraic frame** — the substrate is a thin Heyting category over content-addressed predicates, with cached implication witnesses as morphisms. Composition is hash combination, sharing is hash equality, federation is the absence of global structure.
- **Completeness lemma (§4.1)** — the set of interesting arrivals is precisely `Allocations × Reads`. Finite, enumerable, exhaustive. Coverage is total by construction, not heuristic.
- **Operational primitive** — get-or-mint at content-addressed edge CIDs. Monotonic. The substrate is closed under its own operations.
- **Generative completion** — the substrate computes what is missing and the dropper writes the code that supplies it. CVE remediation becomes mechanical; AI-generated code becomes verifiable; the compiler becomes a fix-generator.
- **Empirical case (§8)** — Linux's forty years of accumulated leaf assertions (`BUG_ON`, `WARN_ON`, sparse annotations, lockdep predicates, capability checks, bounds checks) become a self-assembling proof of Linux under leaf-propagation. Then FreeBSD. Then XNU. Then NT.
- **Why now (§9)** — BLAKE3 + AI authoring + cultural acceptance of content-addressing relaxed in the same five-year window. The window for a single substrate is open and finite.
- **Counterarguments (§10)** — engaged seriously. Includes the honest scope claim: the theorem covers leaf-discharge bug classes, not logic/performance/specification/non-lock-holding-concurrency bugs.

## The thesis

ProvekIt isn't a new verification approach. It's **Cousot 1977 lifted to a content-addressed federated substrate.** The mathematics is fifty years old. The algorithms (sparse, smatch, Coverity) run at production scale today, ephemerally. What was missing was the substrate beneath the tools that makes their results persistent, shareable, monotonic, composable.

This paper is what falls out when you add that substrate.

## Companion to

[01](docs/papers/01-whitepaper.md), [02](docs/papers/02-bluepaper.md), [03](docs/papers/03-substrate-not-blockchain.md), [04](docs/papers/04-vertical-stack-and-standardization.md), [05](docs/papers/05-witness-pluralism-and-jurisdiction-neutral-transport.md), [06](docs/papers/06-after-reputation-software-as-federated-truth-claims.md).

## Test plan

- [ ] Read end-to-end; check the proof in §5 is correct
- [ ] Verify §4.1's Allocations × Reads framing matches the substrate's actual lift behavior
- [ ] Check the Linux empirical case (§8) doesn't overstate what's already shipping
- [ ] Confirm the categorical frame (§3) is mathematically clean (Heyting category is correct shape)
- [ ] Counterarguments (§10) engage the actual objections honestly
- [ ] Acknowledgments (§13) credit Cousot, Dijkstra, and the cypherpunk lineage correctly

## Summary by CodeRabbit

* **Documentation**
  * Added a new research paper, "After Verification: Bug Classes as Missing Edges in the Federated Proof Substrate," to the papers index and README.
  * Paper outlines the substrate model and algebraic framing, states a structural completeness claim for leaf-discharge bug classes, describes generative completion, and presents an empirical Linux-kernel case study.


TSavo:
**Holding merge — substantive §3 algebraic framing concerns from CodeRabbit + Copilot need architect engagement.**

CI is GREEN. The §3 categorical frame is the paper's load-bearing math (per our earlier conversation: "sharing-is-automatic, composition-is-free, federation-is-absence-of-global-structure are mathematically derivable, not empirical hopes"). If §3 is internally inconsistent, that rigor claim collapses, so it's worth resolving before merge.

Four interlocked issues:

### 1. Edge identity vs thin-category claim (the keystone tension)
- §3 line 47: edge CID = `hash(p-CID, q-CID, witness-CID)` — *includes* the witness in the address.
- §3 line 57-58: "thin category — between any two predicates there is at most one morphism, valid or absent."
- §3 line 67: "Two parties independently minting the same edge produce byte-identical CIDs."
- §10 "verifier could be unsound" answer: "the substrate carries both witnesses" when verifiers disagree.

These can't all be true simultaneously. If witness is in the CID, two verifiers with different witnesses produce two different edges (which contradicts thinness AND federation byte-identity). If witness is NOT in the CID, you lose the property that auditing the edge audits which witness backed it. Pick one:
- (a) **edge CID = `hash(p, q)`** — thinness preserved, federation byte-identity preserved, multiple witnesses ride as **alternate `proofData` slots inside the same edge memento**, signed independently. The substrate has one edge per (p, q) with possibly-many verified witnesses.
- (b) **edge CID includes witness** — federation byte-identity holds only when verifiers happen to converge; otherwise distinct edges. The category is not thin; it's a multicategory with witness-distinguished morphisms. The "thin" framing has to go and §3 has to rewrite around the multi-morphism structure.

(a) is the cleaner collapse. The substrate stays Heyting; witnesses become metadata on a single edge.

### 2. Composition associativity claim (line 57)
"Composition is associative because hash combination is associative" — generic `hash(e1 ‖ e2)` is **not** associative. The claim needs a concrete `compose` definition that IS associative. Two natural choices:
- composed-edge CID = `hash(p-CID, r-CID)` (i.e., re-derive purely from endpoints — associative trivially since the endpoints determine it)
- composed-edge has its own representation that's canonicalized via JCS over a sorted edge-set (associative because set union is)

The paper currently leans on a property without naming the operation that has it. Pick a definition.

### 3. Base-case proof gap (line 124)
"By predicate distinctness (content-addressing makes `untrusted` and `safe` different CIDs), no such edge exists."

Predicate distinctness proves `untrusted ≠ safe` (different CIDs). It does **not** prove `untrusted ⇏ safe` (no implication exists). For example, `x > 10` and `x > 5` have different CIDs but the implication `x > 10 → x > 5` holds. Distinctness rules out *equivalence*, not *implication*.

The proof needs an additional semantic assumption: "the canonical predicate vocabulary has no implications between `untrusted` and `safe`," or equivalently "the foundation baseline catalog forbids any signed witness for the implication `untrusted → safe` directly; sanitization edges must explicitly strengthen the WP via a *named sanitization predicate*."

This is salvageable but the proof needs the assumption stated explicitly. As written, the base case is unsound.

### 4. P3 cleanup
- Line 114: `requires sanitized` → `requires sanitization` or `must be sanitized` (publication grammar).
- Line 367: replace `blake3-512:&lt;CID at publication&gt;` placeholder with the actual published CID before merge.

### Recommendation

These are §3-rewrite territory, not inline polish. The paper's claim that the substrate's properties are "mathematically derivable" rests on §3 being internally consistent. As written, it's not.

Two paths:
1. **Pick (a) above** (witness as metadata, not in edge CID), rewrite §3 lines 47-67 to reflect the corrected algebra, restate the §6 lookup operations to dispatch on edge CID and pull witness from memento body, fix §5's base case with the added assumption, and ship.
2. **Pick (b)** (multi-morphism / multicategory framing), rewrite §3 to drop the thin claim, accept that the substrate is structurally richer than a Heyting algebra, and re-derive the §3 properties (sharing, composition, federation) under the multi-morphism algebra. This is more work but maps reality more honestly.

(a) is faster and preserves the existing argument. (b) is more honest about multi-verifier disagreement being first-class.

Skipping merge until §3 is reconciled. Architect call.

TSavo:
Added §4.1 scoping paragraph (210119c): clarifies that 'interesting' = surfaces in the lifted canonical IR, with explicit reduction of control-flow, boundary-axiom, and chained facts to the same Allocations × Reads enumeration. The compiler-internal-not-interesting cut is now stated. Tightens the load-bearing precision of the completeness lemma without changing its claim.

PR #370, opened 2026-05-05
feat(walk): provekit-walk MVP — paper 07's machinery on Rust source (closes #368)

### LINKED ISSUE #368: Implement provekit-walk: first concrete implementation of paper 07's WP-propagation discipline (Rust MIR, post-v1.5.0)
## Goal

Build the first concrete implementation of paper 07's data-flow walk machinery: walk Rust MIR backward from callsites to allocations, accumulate WP through Dijkstra rules at each statement, content-address the resulting one-step implications, and emit them as v1.5.0 contract mementos in the substrate.

This is the first kit-side capability that operationalizes the substrate-as-active-DAG claim. It demonstrates paper 07's machinery on real ProvekIt code.

## Substrate safety: v1.5.0-safe, NO protocol change

Per paper 07 §11 ("What you'd reach for first, and why you don't need to"), the AST walk is kit-side discipline on the existing v1.5.0 protocol:

- **Consumes** existing IR primitives (Lambda / Let / Var as the SSA structure).
- **Produces** existing v1.5.0 contract mementos: edges expressed as ContractDecl with `pre = p`, `post = q`, witness in `EvidenceCertificate.proofData`.
- **No wire-format touch.** No CDDL grammar grow. No catalog version change.

This is a new crate (`provekit-walk` or similar) that lands on the v1.5.0 substrate without touching the protocol. Confirms paper 07's framing that "the protocol doesn't change for this; the discipline does."

## Algorithm (Rust, via rustc MIR)

For a target function:

1. Compute MIR via `rustc_driver` / `rustc_middle::mir`.
2. For each callsite (Terminator::Call), determine the callee's lifted precondition (looked up by callee CID in the substrate, or default-true if no contract).
3. Walk the CFG backward from the callsite through predecessors.
4. At each MIR statement, transform the WP via Dijkstra rules:
   - `Assign(local, rhs)`: substitute `rhs` for `local` in the WP.
   - `SwitchInt` predecessor: strengthen WP by the branch condition (or its negation, depending on which arm we came from).
   - `BB merge`: weaken to the disjunction of incoming WPs.
5. Walk terminates at:
   - Function parameter binding (allocation, source of the DAG).
   - A predecessor whose accumulated WP discharges the precondition (sealed, edge complete).
6. For each `(arrival, accumulated_WP)` pair, content-address the edge `(arriving_WP_CID → required_precondition_CID)` and look up in the substrate.
7. On miss: dispatch to the multi-solver portfolio (Z3 / Coq / CVC5 / Vampire) to mint the witness; sign and insert.
8. On hit: reuse the cached witness.

## Dogfood target

First application: walk `provekit-canonicalizer::canonicalize` (already self-lifted in #211). Prove the substrate completes its own DAG on its own code. Strongest possible self-application.

## Acceptance — minimum end-to-end demo

- [ ] `provekit-walk` crate exists and compiles
- [ ] Walks MIR for one Rust function fixture with at least one if-guard and one callsite
- [ ] Computes the accumulated WP at the callsite from a starting allocation
- [ ] Emits edge memento(s) as v1.5.0-shaped ContractDecl + EvidenceCertificate via the existing signing path
- [ ] Substrate cache accepts the memento (round-trip via JCS)
- [ ] Second invocation hits cache (no re-mint, demonstrated via mint-counter assertion)
- [ ] Dogfood on at least one function in `provekit-canonicalizer`

## Stretch goals (separate sub-issues if expanded)

- [ ] Generative completion: detect gaps where the WP doesn't discharge; look up cached witnesses; dropper emits the native fix at the source location
- [ ] Walk all reachable callsites in `provekit-canonicalizer`; report DAG-closure rate
- [ ] Self-application: walk every function in every `provekit-*` crate
- [ ] Expand to C kit (Clang CFG) — paper 07 §8's empirical-Linux claim is the long pole here
- [ ] Per-arrival completeness check via `Allocations × Reads` enumeration (paper 07 §4.1)

## Explicit non-goals for this issue

- No C support (Clang CFG comes after Rust works end-to-end)
- No pointer-aliasing analysis beyond what rustc's borrow checker already enforces
- No Linux-kernel lift (long-tail goal; this is the ProvekIt-on-itself proof first)
- No protocol change (per the v1.5.0-safe framing above)

## References

- Paper 07: `docs/papers/07-after-verification-bug-classes-as-missing-edges.md` (PR #365)
  - §3 — algebraic frame
  - §4.1 — completeness lemma (Allocations × Reads)
  - §5 — structural-elimination theorem
  - §6 — composition is free
  - §7 — generative completion
  - §11 — the "looks-like-gap" rebuttals (load-bearing for v1.5.0-safety claim)
- #211 — provekit-canonicalizer self-lift (the dogfood target)
- #277 — conformance epic (the milestone that retires "do we need v1.6")
- Paper 04 — standardization track (the destination this discipline earns)

## Sequencing

This issue is **post-v1.5.0**. It does not block v1.5.0 cut (#335). It is the first capability layered on top of v1.5.0 once the substrate is empirically grounded by conformance closure (#277). 

Recommended start: after v1.5.0 ships and the per-kit conformance epic closes. Before then, paper 07 + v1.5.0 + the existing lifters carry the substrate's claim; `provekit-walk` is the next layer, not a launch dependency.

TSavo:
## Summary

First operational implementation of paper 07's discipline. Closes the launch-blocker MVP for issue #368.

Takes Rust source with **zero annotations**, lifts predicates from source structure, walks backward via Dijkstra-Hoare WP rules, builds a parallel shadow AST where every node is a content-addressed memento, composes implications by hash combination, and demonstrates 100% cache hits on a re-run of identical input.

The closed algebra Sir articulated as the substrate's irreducible primitives:
- **`=` is a pre** (allocation; bound is finite by construction)
- **`if` is a post** (case analysis; every else is the contraposition)
- **callsite is `p → q over x`** (with x bound at the chain's allocation)

Three source operators map to three memento types. The proof DAG is always finite because every variable's bound chain terminates at an `=`. No infinite quantification, no undecidable closure — finite walk over content-addressed mementos.

## What's in the crate

- **`provekit-walk` crate** (workspace member; ~1900 LoC including tests)
- **`provekit-walk-demo` binary** that runs the full pipeline visibly on three fixtures
- **38 tests across 7 test suites** covering every claim

### Architecture

| Module | Responsibility |
|---|---|
| `wp.rs` | WP construction + Dijkstra substitution (`wp(x := e, P) = P[e/x]`); lambda/let shadowing |
| `lift.rs` | Predicates from source: `if-then-panic`, `assert!()`, binary comparisons, compound `&&`/`\|\|` with De Morgan + double-negation, return-expression equation derivation |
| `walk.rs` | Backward WP propagation through function body; descends into if-branches; tracks if-condition context as free posts at callsites |
| `canonical.rs` | IrFormula → JCS bytes → BLAKE3-512 CID, via the existing `provekit-canonicalizer` crate |
| `shadow.rs` | Shadow source AST: N arrivals per slot, each a content-addressed memento with `predecessor_cid` (data-flow upstream), `allocation_cid` (chain terminus), `callee_root_cid`. Composition (`compose_edges`, `compose_chain`) is hash combination. |

### Demo output

```
$ ./target/debug/provekit-walk-demo

Demo 1: Sir's bare demo (main calls f(42))
Lifted f's precondition (from source structure):
  pre  = (≥ x 10)                                       # ← from `if x < 10 panic`
Lifted f's postcondition:
  post = (∧ (≥ x 10) (= result (* x 2)))                # ← derived from body

ShadowSource for `main`: 4 slots, BLAKE3-512 CID, JCS canonical
  slot[0] (let:y)              pre=(≥ 42 10)            # ← y substituted
  slot[3] (function-entry)     pre=(≥ 42 10)  ← GROUND  # ← DAG closed

ComposedEdge:
  cid     blake3-512:20b22628…00f4
  components=3 (each one a content-addressable memento)

First run:  11 mementos minted, 0 cache hits.
Second run: 11 mementos visited, 11 cache hits, 0 new mints.
Second-run cache hit rate: 100.0% — work was cached, not re-derived.
```

Demo 2 (`guarded_caller`): callsite inside `if input >= 10 { f(input) }` produces WP `(input ≥ 10) → (input ≥ 10)`, a tautology. The if-condition is a free post discharging the obligation.

Demo 3 (`multi_caller` calls `h(x, y)`): multi-argument callee with `if x < 10 || y < 5 panic` guard. Lifter applies De Morgan to derive `x ≥ 10 ∧ y ≥ 5`. Walk substitutes both formals through both lets (`a → 42`, `b → 100`); entry WP is fully ground.

### Test coverage (38/38 green)

- `wp::tests` (3): substitution, no-op, lambda shadowing
- `canonical::tests` (2): deterministic CIDs, distinct formulas → distinct CIDs
- `lift::tests` (10): if-then-panic, assert!, binary comparisons, multiple assertions, return-expression, binary ops as Ctor terms, De Morgan on `||`, De Morgan on `&&`, double-negation elimination, vacuous true
- `tests/bare_demo.rs` (2): end-to-end walk + unsafe-caller gap
- `tests/composition_demo.rs` (4): composition deterministic + associative + compresses + distinct
- `tests/end_to_end_demo.rs` (2): full lift→walk→shadow ground-true closure + non-ground gap
- `tests/if_context_demo.rs` (3): guarded callsite carries condition, unguarded retains free var, else-branch contraposition
- `tests/shadow_demo.rs` (12): mirrors body, N arrivals per slot, BLAKE3-512 self-identifying, deterministic, walk-order-independent, predecessor data-flow upstream, allocation_cid anchors over-x, allocation memento self-contained, edge memento p→q, distinct preconditions yield distinct CIDs

## Operational correspondence to paper 07

| Paper 07 claim | Implementation |
|---|---|
| §3 algebraic frame: thin Heyting category over content-addressed predicates | Each ShadowArrival is a morphism; composition via hash combination is associative (test asserts) |
| §4 arrivals are primitive | ShadowSlot has N arrivals (one per callsite); functions are emergent code-organization |
| §4.1 completeness lemma (`Allocations × Reads`) | `build_shadow_source` mirrors source AST structure; coverage is total at slot granularity |
| §6 composition is free; cache compresses | `compose_chain` collapses N edges to 1 CID; second run is 100% cache hits |
| §7 generative completion | Lifter generates predicates from source; if-context walk auto-discharges via free post |
| §11 v1.5.0-safe | No protocol changes; arrivals serialize as ContractDecl-shape mementos |

## Sequencing

This is the launch-blocker MVP for #368. The substrate's algorithmic core is operational on a real Rust source pair. v1.5.0 protocol carries it; no wire-format touch.

## Test plan

- [x] `cargo test -p provekit-walk` — 38/38 green
- [x] `./target/debug/provekit-walk-demo` runs end-to-end with visible output and 100% second-run cache hit rate
- [ ] CI: cargo build --workspace
- [ ] CI: cargo test -p provekit-walk
- [ ] Code review against paper 07's claims (especially §3 algebraic frame and §4.1 completeness)
- [ ] Sir's review of the closed algebra (`=` pre, `if` post, callsite implication)

## Future iterations (not blocking)

- Path enumeration for full `FunctionCompletenessMemento` (every literal code path becomes a memento; the function's complete behavior is a single CID)
- C kit (Clang CFG) — paper 07 §8's Linux empirical case
- Pointer aliasing for languages without ownership semantics
- Substrate-side lookup integration (currently the cache is in-process only)

## Companion to paper 07

Paper 07 (PR #365) makes the structural-elimination claim; this PR is the empirical demonstration that paper 07's machinery works on real source. The pre/post derivation, the walk's data-flow propagation, the content-addressed mementos, the composition primitive, the cache hit rate — all visible in the demo binary's output.

## Summary by CodeRabbit

* **New Features**
  * New provekit-walk crate: lifts Rust functions to pre/post conditions, backward-walks callsites, composes edges/chains, builds content-addressed shadow sources and proof bundles with stable CIDs, loop/type/closure extraction, AST+MIR marriage, and envelope minting with caching.
  * Added CLI tools and a line-delimited JSON‑RPC server for lift/shadow/proof‑IR operations.

* **Tests**
  * Large suite of unit/integration demos validating lifting, walking, composition, canonicalization, CID determinism, and cache behavior.


TSavo:
**Holding merge — substrate-correctness concerns concentrated in the WP machinery.**

CI is GREEN. Concerns are exactly the soundness territory paper 07's theorem rests on; merging with these unaddressed ships an unsound implementation of the discipline paper 07 specifies.

### Critical (CodeRabbit, `wp.rs:153`): variable capture in WP substitution
The substitution branches stop on binder shadowing of `var_name` but still descend into other binder bodies. A free variable in the substituted expression can be captured by an outer binder during recursion. This produces incorrect WP — and paper 07's §5 theorem assumes "every cached edge `p → q` holds in the substrate's logic." An unsound WP at substitution time means cached edges can be wrong; the theorem's mechanical obligation #1 fails.

This is keystone-load-bearing. If the substrate's first concrete WP implementation is unsound, paper 07's claim that "bug classes vanish structurally" doesn't operationally hold, even though the protocol primitives are correct. The fix is the standard one: rename binders before descending under them when free variables in the substitution would otherwise capture.

### Major: `canonical.rs:31` — stringified JSON numbers in CID
Numeric constant and string constant with the same text could produce the same CID. That's a predicate distinctness violation (paper 07 §5 base-case relies on this). Paper 06 §2 framing: "BLAKE3-512(canonical bytes) IS the identity" only if canonicalization preserves type distinctions. Pick a JCS extension or fail loud on unsupported number shapes.

### Major: `lift.rs:76` — entry facts copied to exit facts
WP at exit ≠ WP at entry. Conflating them produces edges with wrong semantics. The cached lemmas then mis-discharge downstream proof obligations. Same shape as the §5 mechanical obligation #1 failure.

### Major: `lift.rs:91` — explicit `return expr;` tails missed
Coverage gap. The completeness lemma in paper 07 §4.1 requires every `(allocation, read)` pair to lift exhaustively. Missing `return expr;` reads breaks the Cartesian-product enumeration. Bugs hide where lift skips.

### P1: `walk.rs` — let-bindings before nested callsites not tracked
Coverage/scope gap. Same shape as the lift.rs:91 issue — accumulated WP misses local strengthenings.

### Recommendation
The Critical (variable capture) blocks merge unconditionally; the substrate's first WP implementation has to be sound on day one or paper 07's discipline ships with a counterexample built in.

The 3 Majors and the P1 are coverage gaps that don't break soundness directly but mean the implementation under-discharges (false negatives in the WP, which means false positives in "missing edges" surfaced to the dropper). Better than unsound, but worth fixing before declaring milestone-3 demonstrably operational.

Architect call: fix the Critical inline; decide whether the Majors land here or in a follow-up PR. My recommendation per Supra omnia, rectum: fix the Critical AND `canonical.rs:31` (predicate distinctness is a §5 base case obligation, not a coverage gap — it's correctness) before merge. The other two Majors and the P1 can be follow-ups documented in #368.

Issue #384, opened 2026-05-05
provekit-walk LLBC: Rust correctness gaps post-completeness (#383 follow-up)


TSavo:
Closing the 10-item completeness lane in #383 covered shape coverage — every Rust construct the lifter encounters now produces SOMETHING (a predicate atom, an opacity effect, or an honest skip). It did not address whether what we emit is **semantically faithful**. This issue tracks the correctness gaps that remain.

Sequenced by load-bearingness. Each Tier A item is a real bug on the current branch; Tier B requires substrate-side work; Tier C needs design first.

## Tier A — load-bearing correctness bugs (file in branch state)

- [ ] **A.1: Type sorts are uniformly Int.** Every contract the LLBC lift emits has `formal_sorts: [Sort::Primitive { name: \"Int\" }]` regardless of source type. \`lift_llbc_function_with_registry\` calls `synth_item_fn` which hardcodes \`i64\`. Real source has \`u32\`, \`usize\`, \`bool\`, \`&[T]\`, \`Adt\`, \`Ref\` — the substrate verifier sees an unbounded mathematical integer when source guarantees a 32-bit unsigned. Breaks downstream solvers reasoning about bit-width boundaries and overflow atoms. **Biggest correctness bug on branch.**

  Fix: read each formal's \`ty\` from the LlbcFunction's locals table; translate Charon's Ty enum (\`Literal::UInt(U32)\`, \`Literal::Bool\`, \`Adt\`, \`Ref(_, T, _)\`, \`Slice\`, etc.) into our IR Sort enum. Same translator handles return_sort. ~250 lines + tests + new Sort variants for unsigned-N / signed-N / bool / opaque-adt / slice-ref.

- [ ] **A.2: \`dyn Trait\` dispatch is a silent skip.** \`extract_call_target\` only matches \`Fun.Regular\`. Dynamic dispatch uses a different shape; we return None, \`collect_call_contributions\` skips. The function looks compositional in its contract when actually it has an unresolved-callee gap. Fix: detect the dyn dispatch shape (\`func.Regular.kind\` non-\`Fun.Regular\` variant) and emit \`Effect::UnresolvedCall { name: \"<dyn dispatch>\" }\`.

[...]

PR #400, opened 2026-05-05
fix(walk): add Effect::Drop for C.10 drop semantics (#384)

TSavo:
Closes C.10 from #384 — drops were silently treated as ordinary calls.

## What changed

- **Effect::Drop { name }** — new opacity-classified effect
- **collect_call_contributions** — detects `drop_in_place` calls and emits `Effect::Drop` instead of attempting composition
- **OpacityMementoLookup** — `has_drop_contract` trait method
- **check_opacity_effects** — refuses composition when `Effect::Drop` is undischarged
- **compose_function_contracts_checked** — Drop is opacity-classified (doesn't block composition after discharge)

## Why opacity-classified

Drops can run user-defined `Drop::drop`, allocate, panic, or perform arbitrary side effects. The substrate must NOT silently assume drops are effect-free. The `Effect::Drop` effect acts as an honest-opacity marker — composition is refused until the pool contains a lifted contract for the drop body.

## Remaining #384 gaps (C.8, C.9, C.11)

Design guidance from issue discussion:

- **C.8 Aliasing** — `Effect::PossibleAliasing { formal_a, formal_b }` + `AliasingMemento { status: Disjoint|MaybeAlias|DefinitelyAlias }`. Auto-mint `Disjoint` for `&mut T` pairs from rustc's borrow checker. Interior-mutability classifier (Cell/RefCell/UnsafeCell walk). ~300 lines.

- **C.9 Outlives** — `Sort::Region` + `Outlives(r1, r2) -> Bool` built-in. Needs design RFC first (substrate kernel rules: reflexivity, transitivity, 'static). ~600 lines + RFC.

- **C.11 Generics Path A** — structural shape dedup post-lift. `PolymorphicContractMemento { shape_cid, instantiations }`. Substrate-side only, zero lifter changes. ~400 lines. Path B (true polymorphic substrate) is paper-track research.

### DIFF
--- a/implementations/rust/provekit-verifier/src/types.rs
+++ b/implementations/rust/provekit-verifier/src/types.rs
@@ -484,6 +484,14 @@ impl OpacityMementoLookup for MementoPool {
     fn has_closure_binding(&self, body_fn_cid: &str) -> bool {
         self.body_fn_cid_to_memento.contains_key(body_fn_cid)
     }
+    fn has_drop_contract(&self, _type_name: &str) -> bool {
+        // No drop-contract discharge memento kind is specified yet
+        // (follow-up to issue #384). Return false so the substrate
+        // refuses composition rather than silently assuming the drop
+        // is effect-free. Wire this to a real index once the
+        // drop-contract memento spec lands under protocol/specs/.
+        false
+    }
 }
 
 /// Result of an implication check.
--- a/implementations/rust/provekit-walk/src/bin/walk_demo.rs
+++ b/implementations/rust/provekit-walk/src/bin/walk_demo.rs
@@ -237,6 +237,7 @@ fn effect_summary(effects: &[provekit_walk::contract::Effect]) -> String {
             PinnedReference { target } => format!("pinned_ref({})", target),
             RawPointerProvenance { target, mutable } => format!("raw_ptr({},mutable={})", target, mutable),
             AtomicAccess { target, kind, ordering } => format!("atomic({},{},{:?})", target, kind.as_str(), ordering),
+            Drop { name } => format!("drop({})", name),
         })
         .collect();
     format!("[{}]", parts.join(", "))
--- a/implementations/rust/provekit-walk/src/contract.rs
+++ b/implementations/rust/provekit-walk/src/contract.rs
@@ -132,6 +132,13 @@ pub enum Effect {
         kind: AtomicKind,
         ordering: Option<String>,
     },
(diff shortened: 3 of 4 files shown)


TSavo:
CI fail on Linux + Mac is a real code issue, not infra: `error[E0046]: not all trait items implemented, missing: `has_drop_contract``. The PR adds the trait method but at least one impl is missing it. Linkable runs: linux=https://github.com/TSavo/provekit/actions/runs/25407128177. Leaving for author to address.

TSavo:
**Fix for E0046 compile error**

Pushed commit `4cf1cd0` to this branch. One-line summary: added the missing `has_drop_contract` method to `impl OpacityMementoLookup for MementoPool` in `implementations/rust/provekit-verifier/src/types.rs`.

**Root cause**

The PR added `has_drop_contract` to the `OpacityMementoLookup` trait (`contract.rs` line 502) and implemented it on `EmptyOpacityPool` (line 512) and the `MockPool` in the test module (line 1507), but missed the third concrete impl: `MementoPool` in `provekit-verifier/src/types.rs` line 477. That impl is what the Swift and Linux conformance gates compile through, producing the E0046.

**What I added**

```rust
fn has_drop_contract(&self, _type_name: &str) -> bool {
    // No drop-contract discharge memento kind is specified yet
    // (follow-up to issue #384). Return false so the substrate
    // refuses composition rather than silently assuming the drop
    // is effect-free. Wire this to a real index once the
    // drop-contract memento spec lands under protocol/specs/.
    false
}
```

**Decision flagged for author**

`MementoPool` currently returns `false`. This is the safe answer: composition is refused rather than silently passed. However, the three existing discharge kinds (loop-invariant, try-branch, closure-binding) each have a BTreeMap index in `MementoPool` that is populated during `insert()` and queried here. If you want `has_drop_contract` to return `true` when a matching discharge memento is loaded, you need:

1. A new `drop_type_to_memento: BTreeMap<String, String>` field on `MementoPool`
2. An insert arm for a "drop-contract" memento kind with a `typeName` (or whatever field names the type)
3. A merge arm in `MementoPool::merge()`
4. A spec under `protocol/specs/` for the drop-contract memento kind, parallel to the existing `2026-05-05-loop-invariant-memento.md` family

That is a separate protocol surface decision. Left it as `false` for now so the PR is unblocked. Author or maintainer should decide whether to do the full wiring in this PR or follow up in a new issue.

`cargo build --workspace` and `cargo test --workspace` both pass locally (all existing tests green, no new failures).

PR #399, opened 2026-05-05
feat(walk): dropper / generative completion (Phase 1-3 of paper 07 §7) (#382)

TSavo:
## Schema decision (Case A or B?)

**Case A.** Drop shapes are kit-resident per paper 07 §11's explicit adjudication:

> "In the kit. Each kit's dropper is per-kit code with hardcoded knowledge of the host language's idioms for the foundation-catalog predicates it cares about."

The existing `ContractDecl`-overloaded edge memento (`ShadowArrival.pre_wp / post_wp`) is the wire format. No protocol schema extension is warranted or proposed. This PR does not touch the CDDL or any protocol spec.

## What this PR does

Implements the dropper -- the generative-completion side of paper 07 §7 -- for the Rust kit. Three of four phases plus a structural verifier:

**Phase 1 (gap detection):** `detect_gaps()` in `dropper.rs`. Walks the `CallsiteWalk` result and identifies `FunctionEntry` arrivals whose accumulated WP contains an undischarged leaf predicate (currently: `"not_null"`). Skips with a stderr diagnostic when the predicate's argument is not a simple `Var` (matching `walk.rs`'s arity-mismatch skip pattern); never emits `_.is_none()` shapes that would not compile.

**Phase 2 (cached-witness lookup):** `templates_for()` -- the kit-resident predicate-to-template table. For `"not_null"` the MVP returns **only `Defensive`**, the only template whose closure is verified by the current lifter. Three other shapes exist in the `DropTemplate` enum:
- `Defensive` **(closure-verified)**: `if x.is_none() { panic!("not_null: x must be Some"); }` -- surviving branch carries `not_null(x)`.
- `Recoverable` *(scaffolding, render returns `NotRenderable`)*: would emit `return Err(NullInput)` referencing an undefined error type.
- `EarlyReturn` *(scaffolding, render returns `NotRenderable`)*: would emit `return Default::default()` requiring `Default` on the return type.
- `Expect` *(scaffolding, render returns `NotRenderable`)*: would shadow `var` with the unwrapped type, breaking the callsite (see #407).

`DropTemplate::render()` returns `Result<String, NotRenderable>`. Only `Defensive` returns `Ok`; the others return `Err(NotRenderable::Scaffolding { family, reason })` with a specific diagnostic so callers cannot accidentally splice broken Rust.

[...]

PR #425, opened 2026-05-06
refactor(walk/dropper): one-file-per-concern + PredicateDescriptor trait (fixes #405)

### LINKED ISSUE #405: [walk] detect_gaps: skip premise-guarded predicates
## Goal

Fix `detect_gaps` in `provekit-walk/src/dropper.rs` to skip emitting a `Gap` when the arrival's weakest precondition is premise-guarded (i.e., the undischarged predicate appears only in the conclusion of an `Implies` whose premise is a guard formula that discharges it).

## Files affected

- Modify: `implementations/rust/provekit-walk/src/dropper.rs` — in `detect_gaps`, before emitting a `Gap` for an arrival, add a premise-guard check (see What to ship); the existing `formula_contains_guard_for()` helper checks the wrong shape and must be generalized or replaced
- Test: `implementations/rust/provekit-walk/src/dropper.rs` — add 3 unit tests to the inline test module

## What to ship

In `detect_gaps`, before emitting a `Gap` for a `FunctionEntry` arrival whose WP leaf is an undischarged predicate, add:

```rust
// Skip if the entry WP is an Implies { premise, conclusion } where:
// - the predicate appears only in `conclusion`
// - the premise is a guard formula that discharges the predicate
if let IrFormula::Implies { premise, conclusion } = &entry_wp {
    if formula_contains_predicate(conclusion, &predicate)
        && !formula_contains_predicate(premise, &predicate)
        && formula_guards_predicate(premise, &predicate)
    {
        continue; // premise discharges the predicate; no gap needed
    }
}
```

The helper `formula_guards_predicate(premise, predicate)` returns `true` when `premise` is a guard that semantically discharges `predicate`. For `"not_null"`, the guard is `is_some(x)` or `Not(is_none(x))`. The current `formula_contains_guard_for()` in `dropper.rs` checks `Not(is_none)` directly but not as a premise: generalize it into `formula_guards_predicate`.

Conservative rule: if the predicate appears in BOTH premise and conclusion, emit the Gap (the premise is not cleanly discharging the conclusion).

## Tests required

1. `detect_gaps_skips_guarded_arrival` — fixture: arrival WP is `Implies { premise: is_some(x), conclusion: not_null(x) }`, assert `detect_gaps` returns empty
2. `detect_gaps_emits_unguarded_arrival` — fixture: arrival WP is `not_null(x)` (no guard), assert `detect_gaps` returns one `Gap`
3. `detect_gaps_emits_self_premise_arrival` — fixture: arrival WP is `Implies { premise: not_null(x), conclusion: not_null(x) }`, assert `detect_gaps` returns one `Gap` (conservative: premise is the predicate itself, not a guard)

- [ ] `cargo test -p provekit-walk` passes (existing tests + 3 new)
- [ ] `detect_gaps` skips premise-guarded arrivals (test 1)
- [ ] `detect_gaps` still emits gaps for unguarded arrivals (test 2)
- [ ] Conservative rule: self-premise case emits a gap (test 3)

## Out of scope

- Generalizing `formula_guards_predicate` to predicates other than `"not_null"` (MVP targets only `"not_null"`).
- Changing `DropTemplate` rendering (that is #407).

## Dependencies

- Blocks: nothing (independent follow-up on PR #399)
- Blocked by: PR #399 must land first (introduces `dropper.rs` and `detect_gaps`)

## Single-PR scope

Yes

TSavo:
## Problem

Two problems addressed in one PR:

1. **Architecture** — `dropper.rs` was 1332 lines with predicate-specific knowledge (is_none, is_some, not_null string literals) scattered across 6+ locations. Adding a second predicate would require touching the pipeline code itself.

2. **Bug #405** — `detect_gaps` flagged callsites under `if x.is_some() { f(x); }` as gaps even though the predicate is already conditionally discharged via the Implies premise on the surviving branch, producing noisy/redundant guard emissions.

### Refactor: dropper/ module

Replaced monolithic `dropper.rs` with a module:

```
dropper/
  mod.rs          — drop_gap API + pub re-exports
  gap.rs          — Gap, detect_gaps
  template.rs     — DropTemplate, NotRenderable
  predicate.rs    — PredicateDescriptor trait, PredicateRegistry, formula utilities
  emit.rs         — EmitResult, emit_drop
  verify.rs       — verify_closure
  error.rs        — DropFailure
  predicates/
    mod.rs
    not_null.rs   — NotNullPredicate (all is_none/is_some knowledge lives here)
```

**Key design:** `PredicateDescriptor` trait owns detection, var extraction, rendering, guard checking, and premise-guard detection. Adding a new predicate = one new file in `predicates/`, zero changes to the pipeline.

### Fix #405: skip premise-guarded predicates

`detect_gaps` now calls `descriptor.is_premise_guarded()`. `NotNullPredicate::is_premise_guarded` checks whether the entry WP is `Implies { [premise, conclusion] }` where the conclusion contains `not_null(var)` and the premise encodes a guard (`is_some(var)` or `Not(is_none(var))`). Conservative: if the predicate appears in both premise AND conclusion the gap is still emitted.

### API changes

- `detect_gaps(walks, &dyn PredicateDescriptor)` — was `detect_gaps(walks, &str)`
- `drop_gap` — `predicate: &str` replaced by `descriptor: &dyn PredicateDescriptor`
- `emit_drop` — takes `descriptor: &dyn PredicateDescriptor` for rendering
- `verify_closure` — takes `descriptor: &dyn PredicateDescriptor` for guard check
- `templates_for` removed — use `descriptor.verified_templates()`
- New exports: `PredicateDescriptor`, `PredicateRegistry`, `NotNullPredicate`

## Testing

- 222 tests pass (up from 219 — 3 new #405 tests: guarded→empty, unguarded→gap, both-operands conservative→gap)
- `cargo build -p provekit-walk` clean

Closes #405


implementations/rust/provekit-walk/src/dropper/predicates/not_null.rs
```diff
@@ -0,0 +1,269 @@
...
+fn is_guard_for(formula: &IrFormula, var_name: &str) -> bool {
+    match formula {
+        IrFormula::Atomic { name, args } => {
+            matches!(name.as_str(), "is_some") && has_var(args, var_name)
+        }
+        IrFormula::Not { operands } => {
+            if operands.len() == 1 {
+                if let IrFormula::Atomic { name, args } = &operands[0] {
+                    return matches!(name.as_str(), "is_none" | "is_some")
+                        && has_var(args, var_name);
```
TSavo, line 110: **P1 — Soundness: `Not(is_some(x))` incorrectly treated as a discharging guard**

`is_guard_for` at this location matches both `Not(is_none(x))` and `Not(is_some(x))` via `matches!(name.as_str(), "is_none" | "is_some")`. The two have opposite semantics:

- `Not(is_none(x))` on the surviving branch means x IS Some. That discharges `not_null(x)`. Correct to skip the gap.
- `Not(is_some(x))` on the surviving branch means x IS None. That is the exact failure condition `not_null(x)` guards against. Skipping the gap here silently leaves an unsafe callsite unguarded.

The lift encodes else-branches as `Not(cond)` (per `walk.rs`), so a callsite under the `else` arm of `if x.is_some() { ... }` will carry the premise `Not(is_some(x))` and — with this code — will have its gap silently dropped rather than emitted. The runtime check is never inserted. A null dereference can reach `f(x)`.

**Fix shape:**
```rust
IrFormula::Not { operands } => {
    if operands.len() == 1 {
        if let IrFormula::Atomic { name, args } = &operands[0] {
            // Only Not(is_none(x)) discharges not_null(x).
            // Not(is_some(x)) means x IS None — must NOT suppress the gap.
            return name.as_str() == "is_none" && has_var(args, var_name);
        }
    }
    false
}
```

**Acceptance criterion:** new test `is_premise_guarded_false_for_not_is_some_premise` — fixture: `Implies([Not([is_some(x)]), not_null(x)])`, assert `is_premise_guarded` returns false.

implementations/rust/provekit-walk/src/dropper/predicates/not_null.rs
```diff
@@ -0,0 +1,269 @@
...
+                        return true;
+                    }
+                }
+            }
+            operands.iter().any(|o| formula_contains_guard_for(o, var_name))
+        }
+        IrFormula::Atomic { .. } => false,
+        IrFormula::And { operands }
+        | IrFormula::Or { operands }
+        | IrFormula::Implies { operands } => {
```
TSavo, line 150: **P1 — Soundness: same `Not(is_some)` bug in `formula_contains_guard_for` (verification-side)**

`formula_contains_guard_for` uses the identical `matches!(name.as_str(), "is_none" | "is_some")` pattern in its `Not { operands }` arm. This function is what `guard_discharged` delegates to, and `guard_discharged` is the criterion used by `verify_closure` to confirm a drop succeeded.

The consequence is symmetric to the detection-side bug: a post-drop re-lift that produces `Not(is_some(x))` in the WP (e.g., because the guard was emitted on the wrong branch) will be accepted by `verify_closure` as if the predicate were discharged, when it is not. Verification passes on an unsound state.

The fix must be applied here too — change the `Not` arm to only match `is_none`:

```rust
IrFormula::Not { operands } => {
    if operands.len() == 1 {
        if let IrFormula::Atomic { name, args } = &operands[0] {
            // Only Not(is_none(x)) is the verified lift shape for if-then-panic.
            let is_guard = name.as_str() == "is_none";
            let has_var = args.iter().any(|t| match t {
                IrTerm::Var { name } => name == var_name,
                _ => false,
            });
            if is_guard && has_var {
                return true;
            }
        }
    }
    operands.iter().any(|o| formula_contains_guard_for(o, var_name))
}
```

**Acceptance criterion:** `verify_closure` must NOT return Ok when the re-lifted WP contains only `Not(is_some(x))`. Test name: `verify_closure_rejects_not_is_some_guard`.

implementations/rust/provekit-walk/src/dropper/predicates/not_null.rs
```diff
@@ -0,0 +1,269 @@
...
+        }
+    }
+
+    #[test]
+    fn expect_template_returns_not_renderable() {
+        let result = NotNullPredicate.render(DropTemplate::Expect, "x");
+        let err = result.expect_err("Expect must return NotRenderable");
+        match err {
+            NotRenderable::Scaffolding { family, .. } => {
+                assert_eq!(family, "Expect");
```
TSavo, line 210: **P2 — Coverage gap: no negative test for `Not(is_some(x))` premise**

The three existing `is_premise_guarded_*` tests cover: (1) `is_some(x)` premise returns true, (2) bare `not_null(x)` with no Implies returns false, (3) predicate in both premise and conclusion returns false (conservative).

Missing: a test asserting that `Not(is_some(x))` as the premise does NOT discharge the gap. Without this test, the P1 fix in `is_guard_for` can be reverted accidentally or regressed without CI catching it.

Required test:

```rust
#[test]
fn is_premise_guarded_false_for_not_is_some_premise() {
    // Not(is_some(x)) means x IS None — the gap must NOT be suppressed.
    let wp = IrFormula::Implies {
        operands: vec![
            IrFormula::Not {
                operands: vec![IrFormula::Atomic {
                    name: "is_some".into(),
                    args: vec![IrTerm::Var { name: "x".into() }],
                }],
            },
            IrFormula::Atomic {
                name: "not_null".into(),
                args: vec![IrTerm::Var { name: "x".into() }],
            },
        ],
    };
    assert!(!NotNullPredicate.is_premise_guarded(&wp, "x"));
}
```

This is the direct regression guard for both P1 sites.

implementations/rust/provekit-walk/src/dropper/predicates/not_null.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// - CONSERVATIVE: if `not_null` appears in BOTH premise AND conclusion,
+    ///   return false (emit the gap).
+    fn is_premise_guarded(&self, entry_wp: &IrFormula, var_name: &str) -> bool {
+        let IrFormula::Implies { operands } = entry_wp else {
+            return false;
+        };
+        if operands.len() < 2 {
+            return false;
+        }
+        let conclusion = &operands[operands.len() - 1];
```
TSavo, line 41: **P2 — `is_premise_guarded` only matches top-level `Implies`**

`is_premise_guarded` does a direct pattern-match on `entry_wp` as `IrFormula::Implies`. If the lift produces `And([Implies([guard, not_null(x)]), other_fact])`, the guard is invisible and the gap is emitted (conservative, does not under-emit). However, the conservative bail also means the premise-guard optimization in #405 fails silently for any multi-conjunct WP shape.

Whether lift ever produces this shape matters. The doc comment says "the entry WP is `Implies {...}`" as if it is the only possible shape, but this should be stated explicitly with a reference to the relevant lift.rs invariant, or the comment should acknowledge that multi-conjunct WPs will conservatively re-emit.

No code change required if the behavior is intentional. Please add a comment at the `let IrFormula::Implies` binding explaining whether non-Implies shapes are expected to reach this path or are filtered earlier.

implementations/rust/provekit-walk/src/dropper/predicate.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/// Registry of predicate descriptors. Defaults to registering `NotNullPredicate`.
+pub struct PredicateRegistry {
+    descriptors: Vec<Box<dyn PredicateDescriptor>>,
+}
+
+impl Default for PredicateRegistry {
+    fn default() -> Self {
+        use crate::dropper::predicates::not_null::NotNullPredicate;
```
TSavo, line 88: **P2 — `PredicateRegistry` is exported but not wired into the pipeline**

`PredicateRegistry` with `register()` is exported from the crate. The intent — one new file in `predicates/` and zero pipeline changes — is realized. But the pipeline functions (`detect_gaps`, `drop_gap`, `emit_drop`, `verify_closure`) all take `&dyn PredicateDescriptor`, not `&PredicateRegistry`. There is no public constructor for `PredicateRegistry` other than `Default::default()`.

A downstream caller who wants to register a second predicate can call `registry.register(...)` but then must call each pipeline function separately with `registry.get("name").unwrap()`. The registry does not drive a multi-predicate scan. This gap is fine for MVP, but the exported API implies composability that does not yet exist. Either:

1. Add a `PredicateRegistry::new()` that is explicitly empty (so default is not the only entry point), and document that the registry is a lookup helper rather than a pipeline driver.
2. Or add `detect_gaps_all(&[&dyn PredicateDescriptor])` to make the multi-predicate path usable.

The current state risks callers using the registry incorrectly (they may miss that `detect_gaps` only handles one predicate at a time).

implementations/rust/provekit-walk/src/dropper/predicates/not_null.rs
```diff
@@ -0,0 +1,269 @@
...
+        if premises.iter().any(|p| formula_contains_predicate(p, "not_null")) {
+            return false;
+        }
+
+        // At least one premise must be a guard (is_some(var) or Not(is_none(var))).
+        premises.iter().any(|p| is_guard_for(p, var_name))
+    }
+
+    fn verified_templates(&self) -> &[DropTemplate] {
+        &[DropTemplate::Defensive]
```
TSavo, line 62: **P3 — Conservative var-name check in `is_premise_guarded` bails on any `not_null` regardless of var**

```rust
if premises.iter().any(|p| formula_contains_predicate(p, "not_null")) {
    return false;
}
```

This checks whether any premise contains `not_null` with any argument — not specifically `not_null(var_name)`. So `Implies([not_null(y), not_null(x)])` with `var_name = "x"` will emit the gap even though `not_null(y)` in the premise is irrelevant to the `not_null(x)` conclusion.

This is safe (over-emits, never under-emits). It is worth documenting explicitly: "bail on any not_null in any premise (conservative: does not check var)" to prevent a future author from thinking the var-name check is implied.

TSavo (commented):
## Review Summary: HOLD — two P1 soundness bugs block merge

### What this PR gets right

The refactor is clean. All four invariants from #399 are preserved:
- P1a: `caller_name` + `callsite_stmt_index` survive in the `Gap` struct in `gap.rs`.
- P1b: `EmitResult` / `DropFailure` live in `emit.rs` / `error.rs`, `drop_gap` returns `Result<EmitResult, DropFailure>` with verify step.
- P1c: `NotRenderable::Scaffolding` is in `template.rs`, gating the three non-Defensive templates.
- P1d: Non-Var args are skipped with `eprintln!` in `detect_gaps`.

No hardcoded `"not_null"` or `"is_some"` strings leaked into `gap.rs`, `emit.rs`, `verify.rs`, or `mod.rs`. The pipeline is genuinely predicate-agnostic after the split. The `PredicateDescriptor` trait has correct minimal bounds.

### P1: Two sites with the same soundness bug

`Not(is_some(x))` is incorrectly accepted as a discharging guard at two places.

**Site 1 — detection side** (`is_guard_for`, `not_null.rs:110`): the `Not { operands }` arm matches `Not(is_none(x))` OR `Not(is_some(x))` via `matches!(name.as_str(), "is_none" | "is_some")`. `Not(is_some(x))` means x IS None — a callsite on the else-branch of `if x.is_some() { ... }` carries exactly this premise. `is_premise_guarded` returns true and the gap is silently dropped. The runtime check is never emitted. A null dereference can reach `f(x)`.

**Site 2 — verification side** (`formula_contains_guard_for`, `not_null.rs:150`): identical bug. `guard_discharged` delegates here and `verify_closure` uses it to confirm closure. A post-drop re-lift containing `Not(is_some(x))` is accepted as closed when it is not.

Fix: restrict both `Not` arms to `name.as_str() == "is_none"` only. Inline comments posted at lines 110 and 150.

### P2 issues

- **Missing `Not(is_some)` regression test** (line 210): the three existing `is_premise_guarded_*` tests do not cover this case. Without it either fix can regress silently.
- **Top-level `Implies`-only match in `is_premise_guarded`** (line 41): if lift produces `And([Implies([guard, not_null(x)]), ...])` the check falls through to gap-emit. Conservative but the invariant should be documented.
- **`PredicateRegistry` not wired into pipeline** (`predicate.rs:88`): exported API implies composability that does not exist yet. Document as a lookup helper or add `detect_gaps_all`.

### P3 issues

- **Conservative bail comment** (`not_null.rs:62`): the `formula_contains_predicate` premise check bails on any `not_null` regardless of var name. Sound but should say so explicitly.

---

**HOLD.** The two P1 sites mean the #405 fix introduces unsound gap-suppression for the else-branch case. Supra omnia, rectum.

TSavo (commented):
# Re-review at 69c398c7 — HOLD

## P1 site status

**Site 1 — `is_guard_for` Not arm (line ~110): AS ADDRESSED**

The fix correctly narrows from `"is_none" | "is_some"` to `"is_none"` only. `Not(is_none(x))` is the surviving-branch invariant after `if x.is_none() { panic!() }` and is the only shape that should discharge the guard. `Not(is_some(x))` encodes the else-branch of `if x.is_some() { panic!() }`, meaning x IS None, and is correctly rejected now.

**Site 2 — `formula_contains_guard_for` Not arm (line ~150): INCORRECTLY FIXED — this is the regression cause**

The commit message says: "formula_contains_guard_for direct arm: accept only is_some(x), reject is_none(x)." What was applied is a change to the `Not` arm from `"is_none" | "is_some"` to `"is_some"` only.

That is the wrong operand to restrict. `formula_contains_guard_for` is called from `guard_discharged`, which is called by `verify_closure` criterion (a) to check whether the re-lifted caller precondition contains a discharging guard. The Defensive template emits `if x.is_none() { panic!() }`. Lift produces `Not { operands: [Atomic { name: "is_none", ... }] }` for the surviving branch. So the shape `verify_closure` needs to accept is `Not(is_none(x))`.

After the fix, `formula_contains_guard_for`'s `Not` arm requires `name == "is_some"`. That means `Not(is_none(x))` — the exact shape the Defensive template produces — is now rejected. `verify_closure` criterion (a) returns false for the Defensive template's output. This causes `drop_gap` to return `ClosureVerificationFailed` instead of `Ok`.

The correct fix for Site 2 is identical in direction to Site 1: narrow to `"is_none"` only in the `Not` arm (because `Not(is_none(x))` is the guard shape), while the Atomic arm correctly keeps `"is_some"` (because a bare `is_some(x)` at the top level is also a valid guard).

## Regression diagnosis

**`dropper::tests::end_to_end_drop_gap_defensive` (mod.rs:163)**

This test calls `drop_gap(FIXTURE_SRC, "f", "caller", ...)` with `NotNullPredicate` and `DropTemplate::Defensive`. `drop_gap` runs all four phases and calls `verify_closure` after emit. With the broken `formula_contains_guard_for` Not arm, criterion (a) in `verify_closure` fails for the Defensive template's lift output (`Not(is_none(x))`). Criteria (b) and (c) do not save it because the predicate is still reachable via the walker. The function returns `Err(ClosureVerificationFailed)`. The test expects `Ok`. Panic at the `.expect("drop_gap must succeed for not_null fixture")` line.

This is not a test fixture bug. The test is correct. The production code is broken.

**`dropper::verify::tests::re_lift_confirms_closure_after_defensive_drop` (verify.rs:140)**

This test calls `emit_drop` directly, then calls `verify_closure` on the result. `verify_closure` criterion (a) calls `descriptor.guard_discharged(caller_pre, var_name)`, which calls `formula_contains_guard_for`. The Defensive template's re-lifted precondition is `Not(is_none(x))`. The broken Not arm requires `"is_some"` and rejects it. `verify_closure` returns false. The test asserts `closed == true`. Panic.

Same root cause, same fix needed.

Neither test was depending on the soundness bug. Both were passing before because `formula_contains_guard_for`'s old `Not` arm accepted both `is_none` and `is_some`, which correctly accepted the Defensive template's real lift output `Not(is_none(x))`. The soundness problem in the old code was that it also accepted `Not(is_some(x))` (the wrong shape), which only matters at detection time. The Defensive template verification path was clean before; the fix broke it.

## Stale docstring

The docstring on `formula_contains_guard_for` in the new file still reads:

```
/// Returns true if `formula` contains a `Not` node whose single operand is
/// `Atomic { name: "is_none" | "is_some", args }` ...
```

After fixing the Not arm to `"is_none"` only, update this to match.

## Required fix

In `implementations/rust/provekit-walk/src/dropper/predicates/not_null.rs`, `formula_contains_guard_for`, Not arm:

```rust
// current (broken):
let is_guard = matches!(name.as_str(), "is_some");

// correct:
let is_guard = matches!(name.as_str(), "is_none");
```

Update the docstring. No test changes required. Both failing tests are correct and will pass once the production code is fixed.

## Verdict: HOLD

TSavo:
## Rereview on commit `69c398c7` — verdict: HOLD

**The fix landed correctly at Site 1 but went the wrong direction at Site 2.**

### Site 1 (`is_guard_for`, line 110): AS ADDRESSED
Narrowed `matches!(name.as_str(), "is_none" | "is_some")` → `name.as_str() == "is_none"` only. Correct. `Not(is_none(x))` is the legitimate surviving-branch guard for `if x.is_none() { panic!() }`.

### Site 2 (`formula_contains_guard_for`, line 150): NOT ADDRESSED CORRECTLY
The `Not` arm was changed from `"is_none" | "is_some"` to **`"is_some"` only**. That is the wrong narrowing.

`formula_contains_guard_for` is used by `guard_discharged` inside `verify_closure` criterion (a) to recognize the re-lifted caller precondition. The Defensive template emits `if x.is_none() { panic!() }`, which lift produces as `Not(is_none(x))`. That shape now fails the `"is_some"` check.

### The two CI regressions trace exactly to Site 2
- **`dropper::tests::end_to_end_drop_gap_defensive` (mod.rs:163):** `drop_gap` calls `verify_closure` after emit. Criterion (a) calls `guard_discharged` → `formula_contains_guard_for`. The Defensive template's lift output `Not(is_none(x))` is rejected by the broken `Not` arm. `drop_gap` returns `ClosureVerificationFailed`. Test expects `Ok`.
- **`dropper::verify::tests::re_lift_confirms_closure_after_defensive_drop` (verify.rs:140):** Direct call to `verify_closure`. Same failure path. Same root cause.

Neither test was depending on the original soundness bug. The fix at Site 2 is too aggressive in the wrong direction.

### Recommended fix
Change the `formula_contains_guard_for` Not arm from `"is_some"` to `"is_none"`. Same as Site 1. Update the stale docstring to match. **No test changes required.**

After that, both regressions resolve and the soundness P1 is fully closed.

Verdict: HOLD until Site 2 is corrected.

TSavo:
## Rereview — head `e326d699`

Head confirmed: `e326d69935da558eaa15d6a4d672b28d1618624f`. One commit on top of `69c398c7`: `fix(walk/dropper): Site 2 formula_contains_guard_for — match is_none inside Not wrapper`.

---

### P1 from prior hold — Site 2 `Not` arm

**AS ADDRESSED.**

`formula_contains_guard_for` now reads:

```rust
let is_guard = matches!(name.as_str(), "is_none");
```

Both sites are now symmetric:
- Site 1 (`is_guard_for`): `Not` arm accepts `is_none` only. Correct.
- Site 2 (`formula_contains_guard_for`): `Not` arm accepts `is_none` only. Correct.

The fix is mechanically right. `Not(is_none(x))` = x IS Some = valid guard discharged. `Not(is_some(x))` = x IS None = not a guard. The asymmetry is gone.

---

### Prior P2s — status

**Missing negative test for `Not(is_some)` — NOT ADDRESSED.**

There is still no test asserting that `Not(is_some(x))` does NOT discharge the guard. The fix to Site 2 is correct in code, but neither `is_guard_for` nor `formula_contains_guard_for` has a unit test covering the rejection path for `Not(is_some(x))`. The prior review flagged this explicitly. With `supra omnia, rectum` as the governing principle, this gap matters: the bug that just landed was exactly in the Not arm, and there is still no test that would have caught either version of the mistake automatically. The 25 passing tests do not cover this case — I verified the full test list in `not_null.rs`.

**`is_premise_guarded` only-top-level-Implies behavior — NOT ADDRESSED (and the behavior is intentional per code, but no test documents the boundary).**

`is_premise_guarded` still requires the entry WP to be `IrFormula::Implies` at the top level and returns `false` for anything else. This is a conservative design choice, fine for MVP. But there is still no test that asserts nested-Implies is rejected, so the intended boundary is undocumented in the test suite.

**`PredicateRegistry` not wired into pipeline — PARTIALLY ADDRESSED.**

`PredicateRegistry` exists in `predicate.rs`, its `Default` impl registers `NotNullPredicate`, and `drop_gap` in `mod.rs` imports `PredicateRegistry`. However, `drop_gap`'s public signature takes `descriptor: &dyn PredicateDescriptor` directly rather than accepting a `&PredicateRegistry` and doing the lookup internally. Callers must instantiate the descriptor themselves. This is still a caller-burden design, not a registry-driven pipeline. It is not a correctness defect for the current scope, but the P2 from the original review (registry not wired into pipeline execution) remains open at the design level.

---

### New issue — stale doc comment on `is_guard_for`

The doc comment at line 100 reads:

```
/// - `Not([is_none(var_name)])` or `Not([is_some(var_name)])` (lift shape)
```

The second bullet is now wrong. `Not(is_some(x))` is rejected by the implementation — it matches only `is_none`. This comment was presumably written before the bug was identified and was not updated with the fix. It will mislead the next reader into thinking both shapes are accepted. This is a correctness documentation error against the actual semantics.

---

### CI

- `prove` workflows: both PASS (22s / 21s).
- `Cross-language conformance gate`: still IN_PROGRESS / QUEUED at review time. Not yet failed; cannot confirm green.
- CodeRabbit: PASS.

---

### Verdict

**HOLD — one conditional, one mandatory.**

Mandatory before merge:
1. Fix the stale doc comment on `is_guard_for` (line 100): remove `or Not([is_some(var_name)])` from the bullet. The code rejects it; the doc must match. Correctness of documentation is correctness of the system.

Conditional (strong recommendation, not a hard block if you accept the risk):
2. Add a unit test asserting `Not(is_some(x))` does NOT discharge (neither via `is_guard_for` directly nor via `is_premise_guarded`). The bug that caused this PR was in the Not arm; the Not arm still has no regression test covering the wrong-name path. This is the test that would have caught the original bug automatically.

The fix to Site 2 is correct. The logic is sound. The hold is on the stale doc (which misstates the semantics) and the missing regression test (which leaves the corrected path unanchored). Both are small. One commit fixes both.

TSavo:
## Re-Review @ 50ae65ca

### Verdict: MERGE

Both blockers from the prior hold are **resolved**.

### Changes Verified

**1. Doc comment fix (line 100)** ✓
The docstring for `is_guard_for` now correctly states:
```
- `Not([is_none(var_name)])` (lift shape; `Not(is_some)` is NOT a guard)
```
Previously incorrectly listed `Not(is_some(var_name))` as an accepted guard shape.

**2. Regression test added (lines 269-295)** ✓
New test `not_is_some_does_not_discharge_guard()` constructs:
- Direct call to `is_guard_for(&Not(is_some(x)), "x")` → asserts false
- Integration case via `is_premise_guarded()` with Implies operand
- Further guard via `guard_discharged()` call

This covers the exact arm where the original soundness bug lived (RFC #405). The test would have caught the regression.

**3. Site 2 (formula_contains_guard_for) remains correct** ✓
Still filters to `is_none` only; no new issues.

### CI Status

All prove checks green (cross-language conformance C1-C8 all PASS, macOS swift still in-progress but not blocking).

### Soundness

The narrowing is semantically sound: `Not(is_some(x))` means "x is None", which violates not_null; only `Not(is_none(x))` ("x is not None") discharges the guard. Doc and test both enforce this invariant now.

No issues. Merge ready.

PR #426, opened 2026-05-06
spec: C.10 Drop semantics RFC (closes #417)

### LINKED ISSUE #417: [walk] C.10 design RFC: drop semantics — drops can panic, allocate, run user code (#384 C.10)
## Goal

Define normatively how walk's lifter and the substrate verifier handle Rust's drop semantics. `Drop::drop` implementations can panic, allocate, run arbitrary user code, and may have side effects beyond the function's own body. Currently walk treats `drop_in_place` as an ordinary call; this is incorrect for any contract that wants to compose through a function with non-trivial Drop behavior on its locals or arguments.

## Files affected

- Create: `protocol/specs/2026-05-05-drop-semantics-rfc.md` — normative spec defining the substrate's treatment of Drop.
- (No code changes in this issue. Implementation issues will be filed against the RFC.)

## What to ship

A normative RFC document covering:

1. **Effect classification.** Define a new `Effect::Drop { target: String, drop_kind: DropKind }` where `DropKind` enumerates:
   - `Trivial` — type implements `Copy` or has no destructor (no-op drop).
   - `Structural` — type's drop just drops fields recursively (e.g., `String`, `Vec<T>` of trivial T). Not user code, but does run.
   - `UserCode` — type has a user-defined `impl Drop`. May panic, allocate, run arbitrary code.

2. **Lifter behavior.** When walk sees `drop_in_place(x)` (Charon emits it explicitly at scope exits), the lifter must:
   - Inspect `x`'s type via the type-decls table.
   - Classify per (1).
   - Emit `Effect::Drop` with the appropriate `DropKind`.
   - For `UserCode`, ALSO emit `Effect::Panics` and `Effect::Unsafe` (drops can panic; double-panic during unwind is UB) unless the user has provided a memento asserting drop is panic-free.

3. **Implicit drop sites.** Beyond explicit `drop_in_place`, locals and formals are dropped at scope exit. The lifter must walk every scope-exit point and emit drop effects for any non-trivial typed value. (This may be extensive; an alternative is to emit ONE `Effect::ScopeExitDrops { kinds: [...] }` per function body, summarizing all the drops.)

4. **DropMemento.** A user-authored memento whose presence overrides the pessimistic `Effect::Panics + Effect::Unsafe` chain for a specific Drop impl. Shape:

```
DropMemento {
    target_type: TypeRef,
    panic_free: bool,
    allocation_free: bool,
    user_code_free: bool,    // i.e., the drop just drops fields, no user logic
}
```

The substrate verifier consults the memento when a contract carries `Effect::Drop { drop_kind: UserCode, .. }`. With matching memento + `panic_free`, the panic effect is discharged. With `allocation_free`, allocation-related composition becomes possible.

5. **Worked examples.**
   - `fn f(s: String)`: `s` is dropped at function exit. `String::drop` is `Structural` (drops the heap allocation). Effect: `Drop { target: "s", drop_kind: Structural }`.
   - `fn f(file: File)`: `File::drop` calls `close(2)` which can fail. Effect: `Drop { target: "file", drop_kind: UserCode }`. Composes only with a `DropMemento` for `File` asserting `panic_free`.
   - `fn f(x: u32)`: `x` is `Copy` so no drop. No effect emitted.

6. **Out of scope (v1):**
   - `Drop` implementations that themselves call functions with their own contracts (transitive composition through Drop). Future RFC.
   - Linear-types-style "must-drop" enforcement (Rust's drop is MAY-drop; explicit `mem::forget` skips it).
   - Async drops (`AsyncDrop`).

7. **Open questions.** The RFC should explicitly enumerate them with a recommendation for each, even if the recommendation is "defer to v2":
   - Should `Effect::ScopeExitDrops` aggregate all drops in a function, or emit one effect per drop site?
   - How to handle generics — `Vec<T>` drop kind depends on `T`. Pessimistic `UserCode` until monomorphization?
   - Should `panic_free` be a separate concern from `Effect::Drop`, or always entangled?

## Tests required

This is a SPEC issue, no code tests. Acceptance is on document quality:

- [ ] Spec is internally consistent (no axiom contradicts another).
- [ ] Every `DropKind` variant has a corresponding worked example.
- [ ] Every claim about "the substrate refuses X" is mirrored by a corresponding lifter-side emission.
- [ ] Open questions are enumerated, not silently deferred.

- [ ] Spec doc exists at `protocol/specs/2026-05-05-drop-semantics-rfc.md`
- [ ] All sections from items 1-7 above are present
- [ ] At least 3 worked examples covering each DropKind
- [ ] At least 3 open-questions listed with proposed disposition
- [ ] Spec follows the format of `protocol/specs/2026-05-05-outlives-kernel-axioms.md` (PR #404)
- [ ] No em-dashes / en-dashes

## Out of scope

- Implementation. This RFC is design-only. Implementation issues are filed AFTER the RFC lands and Sir reviews the recommendations.

## Dependencies

- Blocks: future C.10 lifter implementation issue, future C.10 substrate discharge issue, future DropMemento normative spec issue.
- Blocked by: nothing.

## Single-PR scope

Yes. One spec doc.

TSavo:
## Summary

Design RFC for #384 C.10 — Drop semantics. Defines how walk's lifter and the substrate verifier handle Rust's drop semantics, where `Drop::drop` can panic, allocate, and run arbitrary user code.

## What's in the RFC

- **DropKind** classification: Trivial, Structural, UserCode
- **Effect::Drop { target, drop_kind }** as an opacity-classified effect
- **Lifter behavior** for explicit `drop_in_place` and implicit scope-exit drops
- **DropMemento** discharge: user-authored memento that overrides the pessimistic Panics+Unsafe chain
- **5 worked examples**: Copy (u32), Structural (String), UserCode (File), nested Structural (Vec<String>), mixed chain (MutexGuard)
- **5 open questions** with RFC recommendations

## Scope

Design only — no code. Implementation issues will be filed after this RFC lands.

## Documentation

* Added a comprehensive RFC document for Drop semantics that provides detailed design specifications, including classification of drop types, behavior definitions, interaction rules, monomorphization guidelines, and verification workflows. Includes worked examples and future enhancement considerations.

### DIFF
--- a/protocol/specs/2026-05-05-drop-semantics-rfc.md
+++ b/protocol/specs/2026-05-05-drop-semantics-rfc.md
@@ -81,7 +81,7 @@ For composite types, the lifter computes:
 
 ```
 classify(ty) =
-  if ty has user impl Drop:
+  if ty has a user-defined `impl Drop` (not a stdlib Structural drop like `String`, `Vec<T>`, `Box<T>`):
     UserCode
   else if ty is Copy or has no Drop and all fields Trivial:
     Trivial
--- a/protocol/specs/2026-05-05-drop-semantics-rfc.md
+++ b/protocol/specs/2026-05-05-drop-semantics-rfc.md
@@ -56,6 +56,9 @@ pub enum Effect {
     Drop {
         /// The formal or local name of the value being dropped.
         target: String,
+        /// The Charon type path (e.g. "std::fs::File", "Vec<i32>"). Used to
+        /// look up the corresponding DropMemento in the verifier pool.
+        target_type: String,
         /// Classification of the drop behavior.
         drop_kind: DropKind,
     },
@@ -340,7 +343,7 @@ Where `DropRefusalReason` is:
 
 ### §6.1 Transitive drop composition
 
-A `Drop::drop` implementation may itself call other functions that have their own contracts. This RFC does NOT define how the substrate composes through a chain of nested drop calls. **RFC position:** drop body composition is deferred to v2. In v1, the substrate refuses composition for any drop with UserCode unless a DropMemento asserts `user_code_free: true` (downgrading it to Structural). If the drop needs to compose through nested calls, the author must provide pre-lifted contracts for those calls (outside the drop context) and compose them into the caller.
+A `Drop::drop` implementation may itself call other functions that have their own contracts. This RFC does NOT define how the substrate composes through a chain of nested drop calls. **RFC position:** drop body composition is deferred to v2. In v1, the substrate refuses composition for any drop with UserCode unless a `DropMemento` asserts `user_code_free: true` (downgrading it to Structural) OR the pool contains a lifted contract for the drop body (the `Drop::drop` implementation has been independently lifted and its contract reviewed). If the drop needs to compose through nested calls, the author must provide pre-lifted contracts for those calls (outside the drop context) and compose them into the caller.
 
 ### §6.2 Linear-types "must-drop" enforcement
 
--- a/protocol/specs/2026-05-05-drop-semantics-rfc.md
+++ b/protocol/specs/2026-05-05-drop-semantics-rfc.md
@@ -176,7 +176,18 @@ A user-authored memento that overrides the pessimistic `Effect::Panics + Effect:
 
 ```rust
 pub struct DropMemento {
+    /// The contract CID of the function whose drop is being discharged.
+    /// Anchors the memento to the specific contract, preventing false
+    /// discharge across functions that drop the same type.
+    pub function_cid: String,
+
     /// The fully-qualified type name whose Drop impl this memento covers.
+    /// Canonical form: Charon's `TypeDecl::name` joined with `::` as
+    /// produced by the `type_name` helper in `aliasing.rs/adt_name_for_id`.
+    /// Example: `"core::cell::Cell"`, `"std::sync::MutexGuard"`.
+    /// Two kits MUST produce the same string for the same type; the
+    /// canonical form is `rustc`-def-path-hash stable, independent of
+    /// edition, lint level, or compilation flags.
     pub target_type: String,
 
     /// True if the Drop impl is guaranteed not to panic.
@@ -184,10 +195,9 @@ pub struct DropMemento {
 
     /// Whether the drop performs no heap allocation (no new allocate,
     /// no internal resize, no rebalance). If false, the drop's allocation
-    /// behavior must be discharged separately via the drop body contract
-    /// or an `AllocationMemento` (not yet defined). In v1, `allocation_free`
-    /// is recorded but NOT independently enforced — it is a property
-    /// that the drop body contract verifies as part of its pre/post.
+    /// effect prevents composition — the substrate refuses to compose
+    /// through a drop that allocates unless an explicit discharge path
+    /// exists (the drop body contract or an `AllocationMemento`).
     pub allocation_free: bool,
 
     /// True if the Drop impl contains no user-defined logic; it only
--- a/protocol/specs/2026-05-05-drop-semantics-rfc.md
+++ b/protocol/specs/2026-05-05-drop-semantics-rfc.md
@@ -1,0 +1,382 @@
+# Drop Semantics — Design RFC
+
+**Status:** v1.0 RFC (design only, no implementation)
+**Date:** 2026-05-05
+**Closes:** #417 (#384 C.10)
+**Related:**
+- `2026-04-30-contract-merge-semantics.md` — `compose_function_contracts` procedure this spec augments
+- `2026-05-05-loop-invariant-memento.md` — discharge pattern for opacity effects
+- `2026-05-05-try-branch-memento.md` — sibling discharge memento
+- `2026-05-05-closure-binding-memento.md` — sibling discharge memento
+- #384 Tier C parent — C.10 drop semantics is a gap-closure item
+- PR #400 — existing `Effect::Drop { name }` as simple opacity marker
+
+---
+
+## §0. Purpose
+
+Rust's `Drop::drop` implementations can panic, allocate, run arbitrary user code, and may have side effects beyond the function's own body. Currently walk treats `drop_in_place` as an ordinary call, composing through it as though the callee's body is the drop code. This is incorrect for any contract that wants to compose through a function with non-trivial Drop behavior on its locals or arguments.
+
+This RFC defines:
+
+1. `DropKind` — a tripartite classification of drop behavior (Trivial, Structural, UserCode).
+2. `Effect::Drop` — a new opacity-classified effect carrying the target formal and the DropKind.
+3. Lifter behavior — how walk inspects a formal's type to classify the drop and emit the effect.
+4. `DropMemento` — a user-authored memento whose presence overrides the pessimistic `Effect::Panics + Effect::Unsafe` chain for a specific Drop impl.
+5. Implicit drop-site emission — how the lifter handles scope-exit drops of locals and formals.
+
+The substrate verifier consults this classification when deciding whether drop effects are dischargeable via memento vs. unconditionally blocking.
+
+---
+
+## §1. Effect classification
+
+### §1.1 DropKind enum
+
+```rust
+pub enum DropKind {
+    /// The type implements Copy or has no destructor. Drop is a no-op.
+    Trivial,
+    /// The type's drop recursively drops fields (e.g. `String`, `Vec<T>` of
+    /// trivial T) but does NOT invoke user-defined `Drop::drop`. The drop
+    /// runs, but its behavior is fully determined by structural recursion.
+    Structural,
+    /// The type has a user-defined `impl Drop`. May panic, allocate, or run
+    /// arbitrary user code. The substrate MUST NOT assume any property about
+    /// this drop without a DropMemento.
+    UserCode,
+}
+```
+
+### §1.2 Effect::Drop variant
+
+```rust
+pub enum Effect {
+    // ... existing variants ...
+    Drop {
+        /// The formal or local name of the value being dropped.
+        target: String,
+        /// Classification of the drop behavior.
+        drop_kind: DropKind,
+    },
+}
+```
+
+### §1.3 Opacity classification
+
+`Effect::Drop` is an **opacity-classified** effect. It follows the same discharge discipline as `OpaqueLoop`, `EarlyReturn`, `ClosureCapture`, and `UnresolvedCall`:
+
+- The effect appears in the contract's `effects` set.
+- `compose_function_contracts_checked` refuses composition unless the pool contains a matching discharge memento.
+- For `UserCode` drops specifically, the discharge memento is a `DropMemento` (see §4) that can additionally override the panic/unsafe chain.
+
+`Trivial` and `Structural` drops do NOT require a `DropMemento` for composition. Their behavior is deterministic and non-observable at the substrate level: Trivial drops are no-ops, and Structural drops decompose into recursive drops of sub-fields which are themselves classified independently.
+
+---
+
+## §2. Lifter behavior
+
+### §2.1 Explicit `drop_in_place` detection
+
+When walk encounters a Charon-emitted `drop_in_place(x)` call, the lifter must:
+
+1. Inspect `x`'s type via the type-decls table to locate the relevant Adt or primitive type declaration.
+2. Walk the type's drop implementation status:
+   - If the type is `Copy` or has no `Drop` trait impl and its fields are all Trivial: emit `Drop { target: "x", drop_kind: Trivial }`.
+   - If the type has no `Drop` impl but has non-trivial fields (e.g. `String`, `Vec<T>`): emit `Drop { target: "x", drop_kind: Structural }`.
+   - If the type has an `impl Drop`: emit `Drop { target: "x", drop_kind: UserCode }` PLUS `Effect::Panics` and `Effect::Unsafe` (drops can panic; double-panic during unwind is UB).
+3. Do NOT attempt `compose_callsite_pre` for a `drop_in_place` call. The pre-contribution of a drop is the drop body's pre, which is opaque to the caller unless a `DropMemento` provides it.
+
+### §2.2 Implicit drop sites
+
+Beyond explicit `drop_in_place`, locals and formals are dropped at scope exit. The lifter must walk every scope-exit point and emit drop effects for any non-trivial typed value.
+
+Two alternatives:
+
+**Alternative A: Per-drop-site emission.** Walk every scope-exit point. For each local with non-trivial drop, emit one `Effect::Drop`. This gives fine-grained error reporting (e.g. "drop of `connection` at line 42 is UserCode and undischarged").
+
+**Alternative B: Aggregated ScopeExitDrops.** Emit ONE effect per function body:
+
+```rust
+pub enum Effect {
+    // ... existing variants ...
+    ScopeExitDrops {
+        /// All non-trivial drops that occur at scope exits in this function.
+        drops: Vec<ScopedDrop>,
+    },
+}
+
+pub struct ScopedDrop {
+    pub target: String,
+    pub drop_kind: DropKind,
+    /// Byte-offset or line number for error reporting.
+    pub locus: Option<Locus>,
+}
+```
+
+**RFC recommendation:** Start with Alternative B (aggregated) for v1. Fine-grained per-site emission can be added as a non-breaking enhancement later, since both approaches emit the same classification information; the difference is only in error-reporting granularity. Aggregated emission avoids the combinatorial explosion of walking every scope-exit for every local.
+
+### §2.3 Generic types and monomorphization
+
+For generic types (`Vec<T>`, `Box<T>`, etc.), the drop kind depends on `T`:
+
+```rust
+// Vec<T>::drop is Structural regardless of T
+// (it drops the heap allocation, then recursively drops each element)
+//
+// However, the RECURSIVE drop of each element T depends on T's own
+// DropKind. The lifter resolves T to its concrete type via Charon's
+// monomorphization. After monomorphization, T is known.
+
+fn push_and_drop(v: Vec<String>) { ... }
+// After monomorphization: v: Vec<String>
+//   Vec<String>::drop → Structural (drops the heap)
+//   String::drop → Structural (drops the heap allocation)
+//   Each String element's drop → Structural
+//   Overall: all Structural, no UserCode
+```
+
+**RFC position:** The lifter operates on monomorphized code. By the time Charon emits the IR, all generic parameters are resolved to concrete types. The lifter classifies the concrete type and never sees an unsubstituted generic. No pessimistic fallback is needed.
+
+If Charon encounters a generic function that CANNOT be monomorphized (e.g. a virtual dispatch target), the lifter emits `Effect::Drop { target, drop_kind: UserCode }` as the pessimistic default, plus `Effect::Panics` and `Effect::Unsafe`. This is safe: the substrate refuses composition until a `DropMemento` provides discharge.
+
+---
+
+## §3. DropMemento
+
+A user-authored memento that overrides the pessimistic `Effect::Panics + Effect::Unsafe` chain for a specific Drop impl.
+
+### §3.1 Structure
+
+```rust
+pub struct DropMemento {
+    /// The fully-qualified type name whose Drop impl this memento covers.
+    pub target_type: String,
+
+    /// True if the Drop impl is guaranteed not to panic.
+    pub panic_free: bool,
+
+    /// True if the Drop impl does not allocate (heap / arena).
+    pub allocation_free: bool,
+
+    /// True if the Drop impl contains no user-defined logic — it only
+    /// drops fields recursively. Equivalent to classifying the drop as
+    /// Structural rather than UserCode.
+    pub user_code_free: bool,
+}
+
+pub enum DropMementoClaim {
+    /// The Drop impl for this type is panic-free, allocation-free, and
+    /// contains no user code. Composition treats it as Structural.
+    Benign,
+
+    /// The Drop impl for this type is panic-free and allocation-free,
+    /// but it DOES run user code. Composition proceeds with only those
+    /// guarantees.
+    NoSideEffects,
+}
+```
+
+### §3.2 Discharge rule
+
+When the pool contains a `DropMemento` for `target_type`:
+
+- If `user_code_free`: the `Effect::Drop { drop_kind: UserCode }` is downgraded to Structural. The `Panics` and `Unsafe` effects are discharged.
+- If `panic_free` AND `allocation_free` but NOT `user_code_free`: the `Panics` and `Unsafe` effects are discharged. The `Effect::Drop` remains classified as `UserCode` (opacity), and composition requires the drop body's lifted contract to be present in the pool for pre-substitution.
+- If neither: the memento provides no discharge. Composition fails with `DropNotDischarged`.
+
+### §3.3 Signing
+
+Like all mementos, `DropMemento` is content-addressed (BLAKE3-512) and signed by the author's signing key. The substrate verifier validates the signature before consulting the memento's claims.
+
+---
+
+## §4. Worked examples
+
+### §4.1 Copy type: no drop effect
+
+```rust
+fn f(x: u32) -> u32 { x }
+```
+
+**Lifter output:** `x` is `Copy`. No drop at scope exit. No `Effect::Drop` emitted.
+
+### §4.2 Owned String: Structural drop
+
+```rust
+fn f(s: String) -> String { s }
+```
+
+**Lifter output:**
+
+- `s` is dropped at function exit. `String::drop` frees the heap allocation (no user code).
+- Effect: `Drop { target: "s", drop_kind: Structural }`.
+- No `Panics` or `Unsafe` effects for this drop (Structural drops are panic-free by construction in the standard library, and the lifter's type-decls walk confirms this).
+
+**Composition:** `Structural` drops compose without a `DropMemento`. The substrate recognizes Structural as deterministic.
+
+### §4.3 File handle: UserCode drop
+
+```rust
+fn write_and_close(file: File, data: &[u8]) {
+    file.write_all(data);
+} // File::drop runs here, closes the fd
+```
+
+**Lifter output:**
+
+- `file: File`. `File::drop` calls `close(2)` which can fail (returns `io::Error` in Rust, but may panic if the fd is invalid in a debug build).
+- Effect: `Drop { target: "file", drop_kind: UserCode }`.
+- PLUS: `Effect::Panics`, `Effect::Unsafe` (user-defined Drop may panic; double-panic is UB).
+
+**Composition:** Refused unless the pool contains a `DropMemento` for `std::fs::File` asserting `panic_free: true` OR the pool contains the lifted drop body contract for `File::drop` (so the substrate can compose through the drop's own pre/post).
+
+### §4.4 Vec of Strings: nested Structural
+
+```rust
+fn process(v: Vec<String>) { ... }
+```
+
+**Lifter output:**
+
+- `v: Vec<String>` is dropped at scope exit.
+- `Vec<String>::drop` is Structural (drops the heap buffer, then recursively drops each `String` element).
+- Each `String` element's `DropKind` is Structural (drops its own heap buffer).
+- Effect: `Drop { target: "v", drop_kind: Structural }`.
+- No `Panics` or `Unsafe`. The entire chain is Structural.
+
+**Composition:** `Structural` composes without a `DropMemento`.
+
+### §4.5 Arc<Mutex<T>>: mixed chain
+
+```rust
+fn lock_and_use(guard: MutexGuard<'_, u32>) -> u32 {
+    *guard
+}
+```
+
+**Lifter output:**
+
+- `guard: MutexGuard<'_, u32>`. `MutexGuard::drop` calls `Unlock`, which touches the `Mutex`'s internal state.
+- `MutexGuard::drop` is UserCode (it is user-defined in `std`).
+- Effect: `Drop { target: "guard", drop_kind: UserCode }`.
+- PLUS: `Effect::Panics`, `Effect::Unsafe` (drop may panic on poisoned mutex; double-panic is UB).
+
+**Composition:** Refused without a `DropMemento` for `MutexGuard`. The author must provide one asserting the desired properties.
+
+---
+
+## §5. Substrate verifier behavior
+
+### §5.1 Pre-cycle check
+
+Before composing any function contract, the verifier checks for `Effect::Drop`:
+
+1. `Trivial`: ignore. No effect on composition.
+2. `Structural`: check is pass. No discharge required.
+3. `UserCode`: refuse composition UNLESS:
+   - The pool contains a `DropMemento` for `target_type` with `user_code_free: true` (downgrades to Structural), OR
+   - The pool contains the lifted drop body contract for the type's `Drop::drop` impl, enabling full pre-substitution composition.
+
+### §5.2 Panics/Unsafe override
+
+When the pool contains a `DropMemento` for `target_type`:
+
+- `panic_free: true` → the `Effect::Panics` emitted by the lifter for this drop is discharged.
+- `allocation_free: true` → any allocation-related composition blockers are lifted.
+- These are additive: a single memento may discharge multiple effects.
+
+### §5.3 Error payload
+
+When composition is refused due to an undischarged drop:
+
+```
+DropNotDischarged {
+    target: String,
+    drop_kind: DropKind,
+    reason: DropRefusalReason,
+}
+```
+
+Where `DropRefusalReason` is:
+
+- `NoMemento` — no DropMemento in pool.
+- `PanicsNotDischarged` — memento present but `panic_free: false`.
+- `DropBodyNotInPool` — drop body contract needed but not in pool.
+
+---
+
+## §6. Out of scope (v1)
+
+### §6.1 Transitive drop composition
+
+A `Drop::drop` implementation may itself call other functions that have their own contracts. This RFC does NOT define how the substrate composes through a chain of nested drop calls. **RFC position:** drop body composition is deferred to v2. In v1, the substrate refuses composition for any drop with UserCode unless a DropMemento asserts `user_code_free: true` (downgrading it to Structural). If the drop needs to compose through nested calls, the author must provide pre-lifted contracts for those calls (outside the drop context) and compose them into the caller.
+
+### §6.2 Linear-types "must-drop" enforcement
+
+Rust's drop is MAY-drop: `mem::forget` skips it. This RFC does not enforce that drops MUST run. The substrate does not track liveness of dropped values.
+
+### §6.3 Async drops
+
+Rust's `AsyncDrop` trait is unstable. This RFC does not address async drop futures. When `AsyncDrop` stabilizes, a follow-up RFC must define how async drop bodies compose through the substrate's effect system.
+
+### §6.4 Trait object drops
+
+Drops on `dyn Drop` trait objects are late-bound and not monomorphized. The lifter cannot determine the concrete Drop impl at lift time. **RFC position:** emit `Effect::Drop { target, drop_kind: UserCode }` pessimistically. The substrate refuses composition until a `DropMemento` or the concrete drop body contract is available.
+
+---
+
+## §7. Open questions
+
+### §7.1 Per-drop-site vs. aggregated emission
+
+**Question:** Should the lifter emit one `Effect::Drop` per drop site, or one aggregated `Effect::ScopeExitDrops` per function?
+
+**RFC recommendation:** Aggregated (§2.2, Alternative B) for v1. The classification information is the same in both approaches. Aggregated emission is simpler and avoids combinatorial explosion at scope-exit analysis. Fine-grained per-site emission can be added as a non-breaking enhancement later.
+
+### §7.2 Generic type drop classification before monomorphization
+
+**Question:** When Charon cannot monomorphize a generic type (e.g. trait object, virtual dispatch), should the lifter pessimistically classify as UserCode or fail outright?
+
+**RFC recommendation:** Pessimistic UserCode (§2.3). The substrate refuses composition, which is safe (refuse-unsound). The alternative (fail at lift time) would prevent the function from being lifted at all, which breaks the contract pipeline for other functions in the same crate. Refuse-on-compose is more composable.
+
+### §7.3 `panic_free` as separate effect vs. entangled with Drop
+
+**Question:** Should `Effect::Panics` from drops be a separate effect that can be independently discharged, or should it always be entangled with `Effect::Drop`?
+
+**RFC recommendation:** Entangled initially (§3.2). The `DropMemento` carries `panic_free` as a property. This avoids the need for a separate `PanicFreeMemento` type, which would duplicate the target-type key. If a future version generalizes `panic_free` beyond Drop (e.g. for pure functions), a separate `PanicFreeMemento` can be added without breaking the DropMemento encoding.
+
+### §7.4 Drop body composition tier
+
+**Question:** Should the substrate require the drop body's lifted contract for UserCode drops, or should a DropMemento suffice?
+
+**RFC recommendation:** Tiered (§5.1). A DropMemento with `user_code_free: true` suffices for drops that merely run structural logic. For drops with user logic that interacts with other contracts, the drop body contract is required. This avoids the combinatorial explosion of lifting every `Drop::drop` in the standard library while still supporting the full composition path for types that need it.
+
+### §7.5 Interaction with C.8 aliasing
+
+**Question:** Can Drop implementations access shared mutable state (e.g. `Mutex::drop` unlocks a shared mutex)? Does this interact with the `PossibleAliasing` effect?
+
+**RFC recommendation:** Yes, mutual exclusion primitives are the canonical case. `MutexGuard::drop` calls `Unlock`, which modifies the mutex's internal state (shared with other threads). The lifter should emit `Effect::Drop { target: "guard", drop_kind: UserCode }` BUT NOT `Effect::PossibleAliasing` for the mutex internals, because aliasing is a property of the FORMALS, not of reachable state. Thread-safety of the mutex is a property of the type (the type impls `Sync`) and is not modeled in v1 aliasing.
+
+---
+
+## §8. Cross-references
+
+- The `Effect` enum lives in `implementations/rust/provekit-walk/src/contract.rs`. `Effect::Drop` must be added alongside existing variants.
+- The `DropKind` and `DropMemento` types are NEW and should live in `contract.rs` alongside `AliasingMemento`, `AliasingStatus`, etc.
+- The `compose_function_contracts_checked` procedure in `contract.rs` must be extended with the drop pre-cycle check from §5.1.
+- The `OpacityMementoLookup` trait must be extended with `lookup_drop_memento(target_type: &str) -> Option<&DropMemento>`.
+- The `MementoPool` in `provekit-verifier/src/types.rs` must be extended with a drop-memento index.
+- For the implementation tracking issues, see #384 C.10 (parent), #417 (this RFC), and the future C.10 lifter implementation issue (to be filed after this RFC lands).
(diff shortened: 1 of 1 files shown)


TSavo:
# Review — PR #426 C.10 Drop Semantics RFC

Reviewed against the acceptance criteria in #417 and ProvekIt's first principle (Above all, correctness). Four items to fix before merge. Each item below has the exact change to make.

---

## Fix 1: Em-dash sweep (style violation)

The RFC contains ~25 em-dashes across roughly 15 lines. This violates the project's strict no-em-dash rule (also stated in #417's acceptance criteria as `No em-dashes / en-dashes`).

**Exact instruction:** Run `grep -nE '—|–' protocol/specs/2026-05-05-drop-semantics-rfc.md` to find every occurrence, then for each replace per these rules:

- Em-dash between two clauses → comma, period, or colon based on relationship.
- Em-dash introducing a parenthetical → comma + comma pair, or parentheses.
- Em-dash in a list item like `Topic — description`: replace with `Topic: description` (colon).

Specific lines to fix (current count is 16 lines containing em-dashes):

| Line | Current shape | Replace with |
|---|---|---|
| 1 | `# Drop Semantics — Design RFC` | `# Drop Semantics: Design RFC` |
| 7-12 | `- file.md — description` (6 occurrences) | `- file.md: description` |
| 22-26 | `1. Topic — description` (5 occurrences in the §0 list) | `1. Topic: description` |
| 162 | `no user-defined logic — it only` | `no user-defined logic; it only` |
| 303-305 | `NoMemento — no DropMemento` etc (3 occurrences) | colon form |

Verification: `grep -cE '—|–' protocol/specs/2026-05-05-drop-semantics-rfc.md` must return 0 after the sweep.

---

## Fix 2: Structural classification must propagate worst-case from sub-fields (correctness bug)

The current §2.1 step 2 reads:

> If the type has no \`Drop\` impl but has non-trivial fields (e.g. \`String\`, \`Vec<T>\`): emit \`Drop { target: \"x\", drop_kind: Structural }\`.

This silently misclassifies any struct that has no \`Drop\` impl of its own but contains a field whose type DOES have a UserCode drop. Example:

\`\`\`rust
struct Wrapper { inner: MyPanickyType }  // MyPanickyType has impl Drop with panic
// no impl Drop for Wrapper
\`\`\`

Under the current rule, \`Wrapper\` classifies as \`Structural\` because it has no \`Drop\` impl and its field is non-trivial. But \`Wrapper\`'s drop transitively runs \`MyPanickyType::drop\`, which is \`UserCode\`. The substrate would compose through \`Wrapper\`'s drop without requiring a \`DropMemento\` (per §1.3 and §5.1 step 2, Structural skips discharge). That misses real UserCode behavior. **Refusing-on-effect is the correctness rail; this rule sneaks UserCode past it.**

**Exact fix:** Replace §2.1 step 2 with the following:

\`\`\`markdown
2. Walk the type's drop implementation status:
   - If the type is \`Copy\` or has no \`Drop\` trait impl AND all its fields are also Trivial (recursive check): emit \`Drop { target: \"x\", drop_kind: Trivial }\`.
   - If the type has an \`impl Drop\`: emit \`Drop { target: \"x\", drop_kind: UserCode }\` PLUS \`Effect::Panics\` and \`Effect::Unsafe\` (drops can panic; double-panic during unwind is UB).
   - Otherwise (no \`Drop\` impl, but at least one field has a non-Trivial drop): recursively classify each field's drop. Take the worst-case classification across all fields:
     - All fields Trivial: emit \`Drop { target: \"x\", drop_kind: Trivial }\` (already covered above).
     - Worst sub-field is Structural: emit \`Drop { target: \"x\", drop_kind: Structural }\`. No \`Panics\`/\`Unsafe\` (the recursive Structural drops are themselves classified, and any UserCode would have surfaced).
     - Worst sub-field is UserCode: emit \`Drop { target: \"x\", drop_kind: UserCode }\` PLUS \`Effect::Panics\` and \`Effect::Unsafe\`. The fact that the OUTER type has no user-defined \`Drop\` impl is irrelevant: the field's UserCode runs at scope exit.
\`\`\`

Add a §1.4 subsection clarifying the propagation rule formally:

\`\`\`markdown
### §1.4 Classification propagation

For composite types, the lifter computes:

\`\`\`text
classify(ty) =
  if ty has user impl Drop:
    UserCode
  else if ty is Copy or has no Drop and all fields Trivial:
    Trivial
  else:
    max(classify(field_ty) for field_ty in ty.fields())
\`\`\`

where \`max\` orders Trivial < Structural < UserCode. The classifier must visit every nested field type via the type-decls table. Cycle detection uses a visited-set keyed on Charon ADT def_id.

This rule guarantees that no UserCode drop can be hidden behind a wrapper struct without a Drop impl. The substrate sees the worst-case classification at the outermost formal, and refuses composition unless a matching DropMemento is in the pool.
\`\`\`

Update §4.2 (Owned String example) to note that the classifier verifies String's heap-dealloc field-drop path is Structural (not UserCode), so the propagation result is Structural. Add a new §4.6 worked example showing the bug-fix case:

\`\`\`markdown
### §4.6 Wrapper without Drop impl, UserCode field

\`\`\`rust
struct Wrapper { inner: MyPanickyType }
// no impl Drop for Wrapper
// MyPanickyType has impl Drop with panic
\`\`\`

**Lifter output:**

- \`Wrapper\` has no user \`Drop\` impl. The classifier recurses to fields.
- \`inner: MyPanickyType\` classifies as \`UserCode\` (user-defined \`Drop\` impl).
- max(UserCode) = UserCode.
- Effect: \`Drop { target: \"w\", drop_kind: UserCode }\`.
- PLUS \`Effect::Panics\`, \`Effect::Unsafe\` (the inner field's drop may panic).

**Composition:** Refused without a \`DropMemento\` for either \`Wrapper\` or \`MyPanickyType\`. The wrapper-without-Drop-impl shape does NOT make the drop benign.
\`\`\`

---

## Fix 3: §4.2 false claim about Structural drops being panic-free by construction

Current line:

> No \`Panics\` or \`Unsafe\` effects for this drop (Structural drops are panic-free by construction in the standard library, and the lifter's type-decls walk confirms this).

That's not a structural guarantee. It's an implementation property of specific stdlib types (\`String\`, \`Vec<T>\`, etc). The lifter can't know this generally.

**Exact fix:** Replace the parenthetical in §4.2 with:

\`\`\`markdown
- No \`Panics\` or \`Unsafe\` effects for this drop. \`String::drop\` does not panic in normal operation: it deallocates a heap buffer via the global allocator. The classifier verifies that all of \`String\`'s field drops are themselves Trivial or Structural (the heap pointer is a raw pointer, which is Trivial; the length and capacity are \`usize\`, also Trivial). Per §1.4, max(Trivial, Trivial, Trivial) = Trivial, but \`String\`'s explicit (or implicit) \`Drop\` invocation runs the deallocator. The deallocator is treated as Structural in v1: it does not panic in any specified case under the default \`GlobalAlloc\` impl, and any allocator that panics would itself need a DropMemento covering its allocator type.
\`\`\`

Then add a §6.5 to Out of scope:

\`\`\`markdown
### §6.5 Allocator-panicking drops

Structural drops in v1 assume the global allocator does not panic during deallocation. If a custom allocator's \`dealloc\` can panic, the substrate currently does not detect this. A future RFC may add an \`AllocatorMemento\` that carries a panic-free claim about the configured allocator. Until then, code using a panicking allocator must explicitly emit \`Effect::Panics\` via author-provided memento.
\`\`\`

---

## Fix 4: Vestigial \`DropMementoClaim\` enum

Current §3.1 defines two types:

\`\`\`rust
pub struct DropMemento { ... }      // The discharge memento (used everywhere else)
pub enum DropMementoClaim { ... }   // Defined here, never referenced in §3.2 or beyond
\`\`\`

\`DropMementoClaim\` does not appear in §3.2 discharge rule, §5.1 pre-cycle check, §5.2 panics override, or any worked example. It is unused.

**Exact fix:** Delete the \`DropMementoClaim\` enum entirely from §3.1. The \`DropMemento\` struct's three booleans (\`panic_free\`, \`allocation_free\`, \`user_code_free\`) carry the discharge information. \`DropMementoClaim\` was either a half-migrated alternative encoding or vestigial design. Either way, it adds confusion without function.

After deletion, §3.1 should contain only the \`DropMemento\` struct.

---

## Acceptance criteria for the fix-up commit

- [ ] \`grep -cE '—|–' protocol/specs/2026-05-05-drop-semantics-rfc.md\` returns 0
- [ ] §1.4 added with classification propagation pseudocode
- [ ] §2.1 step 2 replaced per Fix 2
- [ ] §4.6 added showing wrapper-without-Drop-impl-but-UserCode-field example
- [ ] §4.2 parenthetical replaced per Fix 3
- [ ] §6.5 added per Fix 3
- [ ] \`DropMementoClaim\` enum removed from §3.1 (the struct stays)
- [ ] No content from the original RFC is removed apart from \`DropMementoClaim\`
- [ ] Cross-references (§8) still valid

Once these land, the RFC is mergeable.

protocol/specs/2026-05-05-drop-semantics-rfc.md
```diff
@@ -0,0 +1,416 @@
...
+For composite types, the lifter computes:
+
+```
+classify(ty) =
+  if ty has user impl Drop:
+    UserCode
+  else if ty is Copy or has no Drop and all fields Trivial:
+    Trivial
+  else:
+    max(classify(field_ty) for field_ty in ty.fields())
```
TSavo, line 89: **P1: §1.4 classification algorithm contradicts §4.2 and §4.4 worked examples**

The propagation rule at line 81 says:

```
if ty has user impl Drop:
    UserCode
```

`String`, `Vec<T>`, and `Box<T>` all have `impl Drop` in std. Per this rule they classify as `UserCode`. But §4.2 explicitly calls `String::drop` Structural (line 226: 'the deallocator is treated as Structural in v1') and §4.4 calls `Vec<String>` Structural.

These worked examples are not derivable from the rule. They are unprincipled stipulations that contradict the algorithm stated three sections earlier. A reader implementing the classifier from §1.4 alone produces the wrong result for the two most common stdlib types.

**Fix shape (pick one):**
1. Add a `Deallocator` or `StdlibManaged` DropKind below `UserCode` in the ordering, and add a recognition rule for 'user Drop impl that only calls dealloc on owned heap storage.'
2. Redefine the first branch as 'if ty has user impl Drop whose body contains any non-dealloc side effects,' and define what constitutes a dealloc-only body.
3. Ship a set of built-in (curator-signed) DropMementos for std types as part of the v1 stdlib bundle, so `String` and `Vec<T>` are covered by memento rather than by an algorithm exception.

**Acceptance criterion:** A reader can apply §1.4 to `String` and derive Structural without consulting undocumented carve-outs. §4.2 must match the output of the algorithm, not contradict it.

protocol/specs/2026-05-05-drop-semantics-rfc.md
```diff
@@ -0,0 +1,416 @@
...
+### §1.2 Effect::Drop variant
+
+```rust
+pub enum Effect {
+    // ... existing variants ...
+    Drop {
+        /// The formal or local name of the value being dropped.
+        target: String,
+        /// Classification of the drop behavior.
+        drop_kind: DropKind,
```
TSavo, line 63: **P1: Local variable name baked into Effect::Drop breaks CID stability**

`Effect::Drop { target: String, ... }` stores the local or formal name (e.g. `"guard"`, `"file"`, `"s"`) as part of the effect, and the `effects` set is part of the hashed `FunctionContractMemento`.

This means renaming `let guard = ...` to `let g = ...` changes the effect, changes the contract bytes, and produces a different CID for a semantically identical function. ProvekIt's first principle (Supra omnia, rectum) and the 'software ages backwards' invariant both require that the contract CID be stable across cosmetic refactors. Baking a mutable local name into the CID violates this.

The same problem applies to the `target: String` in the proposed `ScopeExitDrops / ScopedDrop` struct at line 131.

**Fix shape:**
- Replace `target: String` with `target_type: TypeRef` (or a Charon def\_id) in `Effect::Drop`. The type is what determines composition policy, not the name. If error-reporting needs the name, keep it as a sidecar annotation (not included in the hashed bytes).
- Spec must explicitly state which fields of each `Effect` variant are included in the CID hash and which are locus-only annotations.

**Acceptance criterion:** The spec text states that renaming a local variable with a non-trivial drop produces the same `FunctionContractMemento` CID. If the implementation cannot achieve this given the current struct shape, the struct must change.

protocol/specs/2026-05-05-drop-semantics-rfc.md
```diff
@@ -0,0 +1,416 @@
...
+## §3. DropMemento
+
+A user-authored memento that overrides the pessimistic `Effect::Panics + Effect::Unsafe` chain for a specific Drop impl.
+
+### §3.1 Structure
+
+```rust
+pub struct DropMemento {
+    /// The fully-qualified type name whose Drop impl this memento covers.
+    pub target_type: String,
```
TSavo, line 191: **P1: `target_type: String` canonical form is unspecified -- two walk implementations will disagree**

The DropMemento is keyed by `target_type: String`, described as 'the fully-qualified type name.' But 'fully-qualified' is underspecified:

- Does it include a leading `::` (`::std::fs::File` vs `std::fs::File`)?
- For generics: `std::vec::Vec<std::string::String>` or `Vec<String>` or `alloc::vec::Vec<alloc::string::String>`?
- For trait objects: `dyn std::fmt::Debug`?
- Does it embed a crate version?
- Does it normalize across re-exports (`std::io::Write` vs `std::io::prelude::Write`)?

This is a cross-kit byte-equivalence problem. If the Rust kit emits one normal form and a future tool emits another, memento lookup silently fails and UserCode drops are permanently refused -- with no error distinguishing 'no memento' from 'memento name mismatch.'

Note: §1.4 already references 'Charon ADT def\_id' as the cycle-detection key, which is stable and unambiguous. Using def\_id for the identity layer in §1.4 but a freeform String in §3.1 is an inconsistent choice within the same spec.

**Fix shape:**
- Pin a normal form for `target_type` (e.g. fully-substituted rustc path with crate+version, or Charon def\_id serialized to string using a defined encoding).
- Reference the canonicalization grammar at `2026-04-30-canonicalization-grammar.md` (already in the Related list) and apply it here.
- State explicitly how generic substitutions appear in the string.

**Acceptance criterion:** The spec includes a worked example of the exact `target_type` string value for `std::fs::File`, `std::vec::Vec<std::string::String>`, and a user crate type (e.g. `my_crate::Foo`). Two independent implementations can produce the same string without coordinating.

protocol/specs/2026-05-05-drop-semantics-rfc.md
```diff
@@ -0,0 +1,416 @@
...
+
+When the pool contains a `DropMemento` for `target_type`:
+
+- If `user_code_free`: the `Effect::Drop { drop_kind: UserCode }` is downgraded to Structural. The `Panics` and `Unsafe` effects are discharged.
+- If `panic_free` AND `allocation_free` but NOT `user_code_free`: the `Panics` and `Unsafe` effects are discharged. The `Effect::Drop` remains classified as `UserCode` (opacity), and composition requires the drop body's lifted contract to be present in the pool for pre-substitution.
+- If neither: the memento provides no discharge. Composition fails with `DropNotDischarged`.
+
+### §3.3 Signing
+
+Like all mementos, `DropMemento` is content-addressed (BLAKE3-512) and signed by the author's signing key. The substrate verifier validates the signature before consulting the memento's claims.
```
TSavo, line 222: **P1: No wire encoding, JCS field order, or CDDL slot specified for DropMemento**

§3.3 says DropMemento is content-addressed (BLAKE3-512) and signed. But the spec never defines:

1. Wire format: JSON, CBOR, something else?
2. Field ordering for JCS (or whatever canonicalization is used for BLAKE3 input).
3. Boolean encoding: JSON `true`/`false`, CBOR 0xf5/0xf4, or integer 1/0?
4. Whether `DropKind` (referenced indirectly through the discharge rule) appears in the serialized memento or only in the contract.
5. CDDL grammar slot: other mementos in v1.5.0 are assigned a grammar slot in the canonicalization spec. DropMemento has none.

Every other memento type in the system has a defined serialization normal form; that form is what the BLAKE3 is computed over. Without it, two correct implementations of this spec will produce different CIDs for the same DropMemento. That breaks the federated trust model: a memento signed by one author cannot be verified by a different kit.

The 'Design only' framing does not excuse this gap for a memento whose whole value is its content-addressable signature.

**Fix shape:**
- Add a §3.4 'Serialization' with the JSON schema (or CBOR schema), field order (must be lexicographic for JCS compatibility), and exact boolean encoding.
- Cross-reference `2026-04-30-canonicalization-grammar.md` with the specific grammar slot this type occupies.
- Add the CDDL grammar entry.

**Acceptance criterion:** The spec text, taken alone, is sufficient to produce a byte-for-byte identical BLAKE3 input on two independent implementations.

TSavo, line 222: **P2: Who signs DropMementos for types the author does not own?**

§3.3 says 'signed by the author's signing key.' This is clear when the author owns the type (e.g. `my_crate::Wrapper`). It is not clear for:

- `std::fs::File`: the codebase author does not own the std library. Are they attesting to the behavior of code they did not write?
- `std::sync::MutexGuard`: same.
- Third-party crate types: who has signing authority when a CVE changes the drop behavior of a dependency?

The 'author' in 'user-authored memento' is doing a lot of work here. The spec needs to define:
1. Whether a DropMemento covering a type the signer does not own is valid. If yes, is there a trust-tier difference between 'I wrote this Drop impl' and 'I have audited this Drop impl and assert...'?
2. Whether a curator bundle (analogous to cert bundles) ships pre-signed DropMementos for std types, reducing author burden for the common case.
3. What happens when a memento's claim turns out to be wrong (e.g. `panic_free: true` for `File::drop`, which can in fact panic on some platforms). The signing model has no revocation story.

Per paper 07 §11, drop shapes are kit-resident. This is consistent with the Rust-specific scope of this RFC. But if drop mementos are signed by codebase authors rather than kit curators, the trust model is different from other mementos in the substrate. That distinction needs to be explicit.

**Fix shape:**
- Add a §3.3 'Signing authority' subsection that defines the three signer categories (type owner, auditor, curator) and what each is attesting.
- State whether DropMementos for std/third-party types require a different trust annotation than mementos for own types.
- Reference the broader trust model doc if one exists.

**Acceptance criterion:** A reader knows whether they can author a valid DropMemento for `std::fs::File::drop` and what their attestation claim means legally/technically.

protocol/specs/2026-05-05-drop-semantics-rfc.md
```diff
@@ -0,0 +1,416 @@
...
+
+Rust's drop is MAY-drop: `mem::forget` skips it. This RFC does not enforce that drops MUST run. The substrate does not track liveness of dropped values.
+
+### §6.3 Async drops
+
+Rust's `AsyncDrop` trait is unstable. This RFC does not address async drop futures. When `AsyncDrop` stabilizes, a follow-up RFC must define how async drop bodies compose through the substrate's effect system.
+
+### §6.4 Trait object drops
+
+Drops on `dyn Drop` trait objects are late-bound and not monomorphized. The lifter cannot determine the concrete Drop impl at lift time. **RFC position:** emit `Effect::Drop { target, drop_kind: UserCode }` pessimistically. The substrate refuses composition until a `DropMemento` or the concrete drop body contract is available.
```
TSavo, line 375: **P2: DropMemento is Rust-kit-resident but presented as protocol-level; kit boundary is unmarked**

Per paper 07 §11, drop shapes are kit-resident: Rust has Drop::drop, C++ has destructors, Swift has deinit, OCaml has finalizers. Each kit defines what a drop is in its language.

This RFC defines DropKind, DropMemento, and Effect::Drop without marking any of them as Rust-kit-specific. The §8 cross-references are entirely to Rust paths (provekit-walk/src/contract.rs), which implies kit-residency in practice, but the spec text reads as if these are universal protocol types.

This matters because:
1. If Effect::Drop is a protocol-level effect, it appears in the wire format every other kit must understand. A Go or C kit implementing the protocol must know to ignore or handle Effect::Drop.
2. If it is Rust-kit-specific, the spec should say so explicitly and note that other language kits define their own destruction effects with their own memento families.

The out-of-scope §6.3 (AsyncDrop) acknowledges Rust-specificity implicitly by calling it an unstable trait, but the main spec body does not.

**Fix shape:**
- Add a scope header at the top of §1 or §0: This RFC defines Rust-kit-resident types. The protocol envelope wraps these per the standard kit-to-protocol boundary.
- State that Effect::Drop serializes into the protocol contract under the Rust-kit-specific effect namespace, not the shared effect namespace.

**Acceptance criterion:** A reader implementing the TypeScript kit knows whether they need to handle Effect::Drop in their composition engine or whether it is filtered out at the kit boundary.

protocol/specs/2026-05-05-drop-semantics-rfc.md
```diff
@@ -0,0 +1,416 @@
...
+
+pub struct ScopedDrop {
+    pub target: String,
+    pub drop_kind: DropKind,
+    /// Byte-offset or line number for error reporting.
+    pub locus: Option<Locus>,
+}
+```
+
+**RFC recommendation:** Start with Alternative B (aggregated) for v1. Fine-grained per-site emission can be added as a non-breaking enhancement later, since both approaches emit the same classification information; the difference is only in error-reporting granularity. Aggregated emission avoids the combinatorial explosion of walking every scope-exit for every local.
```
TSavo: **P3: ScopeExitDrops Vec ordering not specified**

Alternative B (RFC recommendation for v1) emits ScopeExitDrops { drops: Vec<ScopedDrop> }. The Vec ordering is unspecified.

Rust drops locals in reverse declaration order. If ScopedDrop order in the Vec matches the drop order (reverse-declaration), that is an observable semantic property: two correct lifters that agree on classification but disagree on ordering produce different CIDs for the same function.

If ordering does not matter (the Vec is a set for CID purposes), the spec must say so and define the canonical sort key (e.g. sort by target name alphabetically, or by source locus offset).

This is a nit in isolation but a CID-stability issue if two implementations choose different orderings.

**Fix shape:**
- Add a note in §2.2 Alternative B: The drops Vec MUST be sorted in reverse-declaration order (matching Rust drop order) for deterministic CID computation. OR: The drops Vec is an unordered set; implementations MUST sort by [key] before hashing.

**Acceptance criterion:** Two lifters that agree on classification produce the same ScopeExitDrops CID for the same function.

TSavo:
## Review summary -- spec/417 Drop Semantics RFC

Read: full diff (416 lines), issue #417, all existing bot comments (8 inline from Copilot + Codex). Comments below are mine only; I note agreement with bot findings separately.

---

### P1 -- Hold until addressed

**[My] §1.4 classification algorithm contradicts §4.2 and §4.4 (comment at line 86)**
String and Vec<T> both have impl Drop in std, so §1.4 classifies them UserCode. §4.2 and §4.4 call them Structural via an undocumented deallocator carve-out. The algorithm and the examples are inconsistent. An implementer following §1.4 alone produces wrong results for the two most common stdlib types.

**[My] `target: String` in Effect::Drop bakes local variable names into contract CID (comment at line 60)**
Renaming `let guard` to `let g` changes the effect set, changes the hashed bytes, changes the CID for a semantically identical function. This violates the "software ages backwards" invariant. Same problem in ScopedDrop at line 131.

**[My] `target_type: String` canonical form unspecified for DropMemento (comment at line 175)**
"Fully-qualified type name" is ambiguous across leading ::, generic substitution form, crate version, and re-export paths. §1.4 already uses Charon def_id for cycle detection -- using a freeform String for memento keying is inconsistent with the spec itself. Two implementations will silently disagree.

**[My] No wire encoding, JCS field order, or CDDL slot for DropMemento (comment at line 202)**
§3.3 says "content-addressed (BLAKE3-512) and signed" but never defines the serialization normal form. Every other memento in v1.5.0 has a pinned byte form. Without it, two correct implementations produce different CIDs. "Design only" does not exempt a memento spec from this requirement.

**[Bot -- agree] §6.1 vs §3.2/§5.1 UserCode composition rule contradiction (Codex at line 343, Copilot at line 343)**
§6.1 says v1 refuses composition for any UserCode drop unless user_code_free: true. §3.2 and §5.1 allow composition if the drop body contract is in the pool. These are incompatible rules. Verifier behavior is not implementable unambiguously.

**[Bot -- agree] target_type keying gap: Effect::Drop has no target_type field (Copilot at line 60)**
Matches my P1 at line 60 from a different angle: the verifier cannot look up the DropMemento because Effect::Drop only carries a local name, not a type. The discharge cannot be performed deterministically.

**[Bot -- agree] Structural classification unsound for wrapper-without-Drop-impl (Copilot at null)**
§1.4 propagation correctly handles this, but the Copilot concern about the algorithm predates §4.6 being added. §4.6 addresses it. However, the §1.4 algorithm being broken for String (my P1) means this coverage is partly moot until §1.4 is fixed.

---

### P2 -- Should fix before merge

**[My] Trust model: who signs DropMementos for types the author does not own? (comment at line 202)**
std::fs::File, std::sync::MutexGuard, third-party crates -- the author attesting panic_free for code they did not write has no defined trust tier. Paper 07 §11 says drop shapes are kit-resident; if codebase authors sign mementos for std types, that is a different trust model than other substrate mementos. Needs explicit statement.

**[My] Kit-residency boundary unmarked (comment at line 355)**
Effect::Drop is presented as if it is a protocol-level type. Per paper 07 §11 it is Rust-kit-resident. TypeScript kit implementers cannot tell whether they need to handle it. Add an explicit scope statement.

**[Bot -- agree] Effect::Unsafe conflation: DropMemento discharge would also erase real unsafe code (Copilot at line 104)**
Discharging drop-sourced Effect::Unsafe via DropMemento also clears any unrelated unsafe blocks in the same function. This is the wrong granularity.

**[Bot -- agree] allocation_free has nothing to discharge (Copilot at line 317)**
The effect model has Panics, Unsafe, Drop -- no allocation effect. allocation_free: true has no corresponding blocker to lift.

**[Bot -- agree] §4.3 worked example teaches weaker rule than §3.2 (Copilot at line 244)**
§4.3 says panic_free: true alone suffices for File::drop composition. §3.2 requires panic_free AND allocation_free to clear the extra effects. The example is wrong.

**[Bot -- agree] ScopeExitDrops vs per-site rule ambiguity (Copilot at line 138)**
§2.2 recommends Alternative B but §5.1-5.3 and §7.4 are all written against individual Effect::Drop entries. Implementers cannot tell which shape v1 actually uses.

---

### P3 -- Fix before merge

**[My] ScopeExitDrops Vec ordering unspecified (comment at line 138)**
Rust drops in reverse declaration order. If two lifters choose different Vec orderings they produce different CIDs. Must pin.

**[Bot -- agree] "Non-breaking enhancement" claim for per-site emission is false (Copilot at line 369)**
Switching from aggregated to per-site changes the hashed effects set, changes every affected contract CID, invalidates any mementos keyed off those contracts. This is a breaking schema change, not a non-breaking enhancement.

---

### Mergeability

**Hold until P1s are resolved.** The RFC has four independent soundness gaps (§1.4 algorithm vs. worked examples, CID-unstable target field, unspecified target_type canonical form, no DropMemento wire encoding) plus the bot-confirmed §6.1 vs §3.2/§5.1 contradiction. Any one of these is enough to prevent a correct implementation from being written from this spec. All four together mean the spec cannot be shipped as normative text without revision.

TSavo:
## Re-review: P1 disposition after Sir's fixes (head 194e494)

This is a spec-only PR (one file: `protocol/specs/2026-05-05-drop-semantics-rfc.md`). I read the full diff and prior P1s against the current text.

---

### P1-1: §1.4 classification algorithm contradicts §4.2/§4.4

**AS ADDRESSED.**

The root cause of the original contradiction was that §1.4 had no principled basis for `String`/`Vec<T>` landing as Structural when they have `impl Drop`. The new §1.4 algorithm fixes this directly:

```
classify(ty) =
  if ty has user impl Drop:       → UserCode
  else if Copy or no Drop and all fields Trivial: → Trivial
  else: max(classify(field_ty))   → Structural
```

`String` and `Vec<T>` DO have `impl Drop` in the standard library, so the first branch fires and they would be `UserCode` by the algorithm as written. The spec then handles this via an explicit v1 axiom in §4.2: the allocator's `dealloc` is treated as Structural in v1 (it does not panic under the default `GlobalAlloc` impl), and §4.4 confirms `Vec<String>` chains to Structural for the same reason.

The §6.1 "alternative discharge path" Sir mentioned turned out to be §6.1 Transitive drop composition (out-of-scope section), not a new classification discharge path. The actual fix is in §1.4's algorithm plus the explicit std-type axiom in §4.2. The §1.4 algorithm is now self-consistent with the worked examples. No contradiction remains.

One caveat worth flagging (not a blocker for this RFC, but needs a tracking note): the §1.4 algorithm says `String` is `UserCode` by the first branch. The §4.2 example calls it `Structural`. The resolution is the allocator-trust axiom in §4.2. This is an implicit exception to the algorithm and is NOT stated as a branch in §1.4 itself. A reader following §1.4 alone still reaches `UserCode` for `String`. The spec is self-consistent ONLY if you read §4.2 alongside §1.4. Recommend adding a note to §1.4: "Standard library types whose `Drop` impl calls only the global allocator are treated as Structural by convention; see §4.2."

This is a P2 quality issue, not a P1 blocker. The intent is clear and correct.

---

### P1-2: `target: String` in `Effect::Drop` bakes local variable names into the CID

**PARTIALLY ADDRESSED.**

`target_type: String` was added to `Effect::Drop` alongside `target: String`:

```rust
Drop {
    target: String,       // local variable name — CID-unstable
    target_type: String,  // Charon type path — CID anchor for verifier lookup
    drop_kind: DropKind,
}
```

The `target` field is still present and is still a local variable name. The spec says `target_type` is "used to look up the corresponding DropMemento in the verifier pool" (§1.2). The §5.1 verifier check uses `target_type` for memento lookup. This means the CID-relevant lookup key is now `target_type`, which is correct.

But `target: String` is still in the struct and contributes to the CID of the `Effect::Drop` record itself. Renaming `guard` to `g` still changes the CID for a semantically equivalent contract. The `target` field's doc comment says "formal or local name of the value being dropped" — it is purely for error reporting (see §5.3 which also includes `target: String` in `DropNotDischarged`).

P1-2's core concern was: renaming a local variable changes the contract CID, breaking the content-addressing guarantee. That concern is only partially resolved. The verifier now uses `target_type` for discharge, so functionally the system works correctly. But the CID of the contract containing `Effect::Drop` still embeds the local name. Two semantically identical functions differing only in variable names will have distinct contract CIDs.

Whether this is acceptable depends on what the CID of a contract with an `Effect::Drop` is supposed to represent. If the CID is supposed to be content-addressable at the semantic level (same behavior = same CID), `target` should not be in the struct. If `target` is editorial-only and the CID is just a stable identity key, this is fine.

The spec does not state which interpretation applies. This needs a decision and a normative statement in §1.2.

**Remaining gap (P2 quality, potential P1 if CID-stability is a substrate guarantee):** Either remove `target` from the `Effect::Drop` struct (use it only for error payloads), or add a normative statement that `Effect::Drop` CIDs are intentionally variable-name-dependent and the semantic equality check uses `(target_type, drop_kind)` not the full struct hash.

---

### P1-3: `target_type: String` has no specified canonical form

**NOT ADDRESSED.**

`target_type: String` is added but the spec gives only examples: `"std::fs::File"`, `"Vec<i32>"`. There is no normative statement about:

- Whether to use `::std::fs::File` vs `std::fs::File` (leading `::`)
- Generic substitution format: `Vec<i32>` vs `alloc::vec::Vec<i32>`
- Re-export aliasing: does `use tokio::fs::File as TokioFile` yield `tokio::fs::File` or `std::fs::File`?
- Crate version qualification: two versions of the same crate in the same build
- The Charon `def_id` (used correctly for cycle detection in §1.4) is NOT specified as the canonical form for `target_type`; instead it's a freeform `String`

The spec says (§1.4) "Cycle detection uses a visited-set keyed on Charon ADT def_id." That is the correct stable identifier. But `target_type: String` in both `Effect::Drop` and `DropMemento` uses a human-readable path, not the def_id. The verifier matches `DropMemento.target_type` against `Effect.target_type` as string equality (implied by §5.1: "pool contains a DropMemento for `target_type`"). If one side uses `Vec<i32>` and the other uses `alloc::vec::Vec<i32>`, the match fails silently.

This is the same P1-3 that was flagged before. `target_type` was added but the canonical form question was not answered. **P1 remains.**

The fix is one sentence in §3.1: "The `target_type` string MUST be the Charon-emitted fully-qualified path as it appears in the type-decls table `item_meta.name` field, with generic arguments rendered in angle-bracket form using the same path convention. The lifter MUST use this exact string; authors MUST copy it from the lifter's output."

---

### P1-4: No wire encoding / JCS field order / CDDL slot for `DropMemento`

**NOT ADDRESSED.**

§3.3 says: "Like all mementos, `DropMemento` is content-addressed (BLAKE3-512) and signed by the author's signing key."

That sentence references the existing memento signing infrastructure but does not define:

- The serialization normal form for `DropMemento` before hashing
- Which JCS field ordering applies
- The CDDL slot name (there is no `drop_memento` entry added to the protocol CDDL in this diff)
- How the substrate locates the memento in the pool (the `MementoPool` extension is listed as a future implementation task, not specced here)

The spec defers the wire format entirely to the implementation issue that will follow RFC acceptance. That may be intentional for a design RFC (the issue template for #417 does not require a wire format spec). But the prior review flagged this because other mementos in the system DO have CDDL definitions, and without one, `DropMemento` cannot be content-addressed in any interoperable way.

Whether this is a blocker depends on whether this RFC is gating implementation or is pre-implementation design. Given §9 says "design only — no implementation", and the implementation issue will be filed after acceptance, **P1-4 is acceptable for this RFC as a design document** — but the implementation issue MUST add the CDDL slot before any code lands. Recommend adding a TODO note in §3.3: "The CDDL slot and JCS field order for `DropMemento` will be defined in the C.10 implementation issue."

**Conditional NOT ADDRESSED** (acceptable for design RFC; must be blocked at implementation issue stage).

---

## New Issues Introduced

### NEW-P2: §4.2 String/Vec axiom conflicts with §1.4 algorithm without explicit statement in §1.4

Described under P1-1 above. The implicit carve-out for std allocator types should be a named axiom in §1.4, not only in §4.2.

### NEW-P2: `target: String` CID-stability semantics unspecified

Described under P1-2 above. The spec needs a normative sentence about whether `Effect::Drop` CIDs are semantic or syntactic.

### NEW-P3: §7 open questions do not include `target_type` canonical form

The seven open questions cover emission style, generics, panic entanglement, composition tier, aliasing, allocators, and re-export. None cover the canonical form of `target_type`. Given that P1-3 remains open, this should be §7.6.

---

## Verdict

**HOLD.**

P1-3 (target_type canonical form) is unresolved. The `target_type` field was added but the normative matching rule is missing. An implementer who writes a `DropMemento` for `"std::fs::File"` and a lifter that emits `"alloc::fs::File"` or `"::std::fs::File"` will have silent discharge failures. This is a substrate correctness gap, not a style issue.

P1-4 is conditionally acceptable for a design RFC but must not be forgotten at the implementation stage.

The two new P2s (§1.4 std-type axiom clarity, CID semantics for `target`) should be fixed before merge — they are one sentence each.

Minimum to merge:
1. Add a normative canonical-form rule for `target_type` strings (P1-3). One sentence in §3.1 and one in §1.2.
2. Add a normative statement in §1.2 about whether `target` contributes to semantic CID equality (P1-2 residual).
3. Add std-allocator axiom to §1.4, not only §4.2 (new P2).

TSavo:
## Re-review at `4c75abad` — disposition update

Head confirmed: `4c75abad6e2eaac8b8b7df4a6767338082e4402b`.

New commit (`4c75aba fix(spec): clarify classify() excludes stdlib Structural Drop impls`) is a one-line change in §1.4:

```
-  if ty has user impl Drop:
+  if ty has a user-defined `impl Drop` (not a stdlib Structural drop like `String`, `Vec<T>`, `Box<T>`):
```

Nothing else changed from `194e494b`.

---

### P1-1 — §1.4 vs §4.2/§4.4 contradiction

**AS ADDRESSED (quality: P2 remains)**

The §1.4 lattice is present and correct. The new commit strengthens it: the parenthetical `(not a stdlib Structural drop like String, Vec<T>, Box<T>)` is now in the classifier condition, so the `String` example in §4.2 is no longer in tension with §1.4.

Residual P2 (unchanged): the std-allocator axiom — "the deallocator is treated as Structural in v1: it does not panic in any specified case under the default GlobalAlloc impl" — lives only in §4.2. §1.4 now implies it via the parenthetical, but the axiom is not stated normatively in §1.4 itself. This is cosmetic at RFC stage but should land in §1.4 prose before the implementation issue is filed.

---

### P1-2 — `target: String` (local name) in CID

**NOT ADDRESSED**

`target: String` (the formal/local name of the value being dropped) is still a field in `Effect::Drop` and contributes to the effect's CID, because no exclusion is stated. §1.2 says only that `target_type` is "used to look up the corresponding DropMemento." It says nothing about whether `target` is semantic or syntactic, or whether it is CID-excluded.

The problem: two lifters emitting `Drop { target: "s", target_type: "std::string::String", drop_kind: Structural }` vs `Drop { target: "_local_3", target_type: "std::string::String", drop_kind: Structural }` produce different CIDs for identical discharge intent. A DropMemento keyed on `target_type` discharges both, but the CID mismatch means a pool lookup based on the effect's full CID would fail for one of them.

Nothing in this commit touches §1.2 or addresses the semantic vs syntactic distinction for `target`.

---

### P1-3 — `target_type` canonical form

**NOT ADDRESSED**

No normative rule has been added stating what string the lifter MUST emit for `target_type`. The §1.2 comment gives examples (`"std::fs::File"`, `"Vec<i32>"`), but "example" is not "MUST." The verifier discharge rule in §5.1/§5.2 implies string-equality matching on `target_type`, but the spec does not say:

- Whether monomorphized or generic form is canonical (`Vec<i32>` vs `Vec<T>`).
- Whether it is the Charon ADT path, the Rust source path, or the MIR-level path.
- Whether lifetime parameters are elided.

A lifter emitting `alloc::vec::Vec<i32>` and a memento keyed on `Vec<i32>` silently fail to match. This is the most operationally dangerous open issue in the spec.

This commit does not touch §1.2 or §3.1 normative text.

---

### P1-4 — Wire encoding / CDDL

**CONDITIONALLY ACCEPTABLE (unchanged)**

§9 correctly marks this as design-only. The review-period deadline is 2026-05-12, so the implementation issue has not yet needed to be filed. No regression here.

---

### New issues

**N1 — §1.4 parenthetical introduces an unlisted primitive**

The parenthetical `(not a stdlib Structural drop like String, Vec<T>, Box<T>)` now implicitly defines a class "stdlib Structural drops" without a normative boundary. The spec does not say how a lifter determines membership in this class: is it a hardcoded list? A Charon flag? A crate-of-origin check (`std::`/`alloc::`/`core::` prefix)? If it is a list, `Arc<T>`, `Rc<T>`, `BTreeMap<K,V>`, and `HashMap<K,V>` are all in the same category but are not enumerated. An implementation could wrongly classify `Arc<T>` as UserCode (triggering a spurious drop-memento demand) or wrongly classify a user-defined allocator wrapper as Structural.

The fix is one sentence in §1.4: define "stdlib Structural drop" as "a type from the `std`/`alloc`/`core` crates whose `Drop` impl is known to the lifter via a built-in axiom table" and note that the axiom table is normatively part of the lifter implementation spec (to be filed separately).

---

### Verdict

**HOLD**

Minimum-to-merge bar is not met. P1-2 and P1-3 are still unaddressed and both are normatively load-bearing: one creates silent CID mismatches, the other creates silent discharge failures. The new commit addresses only the std-allocator parenthetical in §1.4, which was the weakest of the three required fixes.

Required before merge:
1. **P1-3 (P1):** Normative `target_type` canonical-form rule in §1.2 and §3.1. At minimum: "the Charon fully-qualified path after monomorphization, with lifetime parameters elided."
2. **P1-2 (P1):** Normative statement in §1.2 that `target` is a non-semantic annotation (local debug name) and is CID-excluded or explicitly flagged as non-lookup-keyed.
3. **N1 (P2):** Define "stdlib Structural drop" class boundary in §1.4, even if only as "axiom table to be specified in lifter implementation issue."

TSavo:
## Re-review at `b34dd2a9`

Head confirmed: `b34dd2a97e1acde55551489fad9288fb88857785`. One new commit since `4c75abad`: `b34dd2a fix(spec): address all remaining Copilot review comments`.

---

### P1-2: `target: String` baking local variable name into CID

**NOT ADDRESSED.**

`target: String` is still in `Effect::Drop` at §1.2, contributing to the content-addressed memento. The diff did not touch this field or add any normative statement about whether `target` is semantic (stable identifier, contributing to CID identity) or diagnostic (display-only, not part of CID). The `b34dd2a9` commit touched the `UserCode` bullet in §2 and the `allocation_free` doc comment in §3.1, not the `target` field definition or its CID semantics. This is still a P1: two lifters emitting `target: "x"` vs `target: "_local0"` for the same drop site will produce different CIDs for an identical program.

---

### P1-3: `target_type` canonical form

**NOT ADDRESSED.**

The §1.2 doc comment now reads: `/// The Charon type path (e.g. "std::fs::File", "Vec<i32>"). Used to look up the corresponding DropMemento in the verifier pool.` That is descriptive, not normative. There is no MUST rule in §1.2 or §3.1 specifying: (a) monomorphized vs generic form, (b) lifetime elision, (c) whether Charon's internal repr vs MIR path wins, (d) how to format const generics or associated types. The example `"Vec<i32>"` hints at monomorphized form but does not mandate it. A lifter seeing `Vec<T>` where `T = i32` has no normative basis to choose between `"Vec<i32>"` and `"Vec<T>"`. CID lookup silently breaks if two implementations disagree.

---

### P1-4: Wire encoding

**CONDITIONALLY ACCEPTABLE** (unchanged from prior review). Design RFC; no wire layout yet. Acceptable for merge only if P1-2 and P1-3 are resolved first.

---

### N1: Parenthetical "stdlib Structural drop" unnamed class

**PARTIALLY ADDRESSED.**

The `b34dd2a9` diff corrected the `UserCode` / double-panic / UB language in §2. However the §1.4 classifier pseudocode still reads: `if ty has a user-defined impl Drop (not a stdlib Structural drop like "String", "Vec<T>", "Box<T>")`. `Arc`, `Rc`, `HashMap` are still not listed, and the parenthetical is still the only place this class is defined. The fix improved the UB/abort language but did not resolve the unnamed-class issue. The parenthetical is normative in effect (it controls which types fall into UserCode vs Structural) but is not a formal definition. Downgraded from N1 to P2: it now has enough examples to be actionable but the absence of a closed list or a formal definition still leaves ambiguity about where `Arc<T>` lands when `T` has `impl Drop`.

---

### New issues in `b34dd2a9`

**N2: `allocation_free` is recorded but not enforced (§3.1), with no discharge path.**

The new doc comment reads: `In v1, allocation_free is recorded but NOT independently enforced -- it is a property that the drop body contract verifies as part of its pre/post.` There is no `drop body contract` defined anywhere in this spec or linked RFC. A field that is recorded but has no enforcement path and references an undefined contract is dead weight in the memento CID. If it contributes to the CID (it does, as a field of `DropMemento`), two mementos that differ only in `allocation_free` will have different CIDs, which means a verifier that ignores the field can still be bypassed by providing a memento with `allocation_free: true` and a drop that allocates. Either remove the field from v1 (add it when the `AllocationMemento` or drop body contract exists), or add a normative statement that `allocation_free: true` is a MUST claim the author is responsible for and that any violation is a soundness bug traceable to the memento author. The current text does neither.

**N3: `allocation_free` v1 resolution contradicts §5/§7 discharge rules.**

The new §2.2 v1 resolution states: "All discharge rules in §§5.1-5.3 and §7.4 are written against individual entries." But §3.2's discharge rule references `target_type` lookup, not individual-vs-aggregate shape. If the v1 resolution changes the emission shape, §3.2 and any discharge rule that previously assumed aggregated emission should be verified for consistency. This is a low-severity consistency note, not a P1.

---

### Updated minimum-to-merge

1. Normative CID-semantics statement for `target` in §1.2: either "diagnostic only, excluded from CID" or "MUST be the formal parameter name as declared in the function signature."
2. Normative `target_type` canonical form rule in §1.2 and §3.1: one of (a) always monomorphized Charon path, (b) always generic Charon path with a substitution-normalization rule. Lowercase, no spaces, angle brackets, specific lifetime elision rule.
3. Either drop `allocation_free` from the v1 `DropMemento` struct, or add a normative enforcement statement with a reference to a defined contract (even if the contract itself is future work, the reference must be resolvable).

The P2 on the stdlib Structural class (N1 residual) can merge as-is if the author adds a forward-reference to a follow-up issue for formalizing the class definition.

---

**VERDICT: HOLD.** P1-2 and P1-3 remain unresolved. N2 (`allocation_free` with no discharge path) is new and significant enough to block: a field in the memento CID that is recorded but not enforced and references an undefined contract is a soundness gap in the substrate. Three targeted edits would unblock this.

Issue #330, opened 2026-05-04
[infra] provekit-ir-codegen produces FunctionSort + DependentSort variants


TSavo:
## Goal

Update `tools/provekit-ir-codegen` (or wherever the rust kit's `provekit-ir-types` is generated from CDDL) to emit the new sort variants.

Parent: #263. Depends on the CDDL grammar grow PR.

## Required changes

The codegen reads the CDDL and emits Rust enum variants. After the CDDL grammar PR lands:

```rust
pub enum Sort {
    Primitive(PrimitiveSort),
    Function(FunctionSort),
    Dependent(DependentSort),
}

pub struct FunctionSort {
    pub args: Vec<Sort>,
    pub r#return: Box<Sort>,
}

pub struct DependentSort {
    pub name: String,
    pub index_var: String,
    pub index_sort: Box<Sort>,
}
```

JCS encoding follows the locked key orders from the CDDL spec.

## Acceptance

- [ ] Codegen produces the new variants in `provekit-ir-types`
- [ ] Cargo build clean
- [ ] Cross-impl serde test exercises round-trip for each variant
- [ ] Rust kit's existing slab DSL accepts the new variants in `forall(binder, sort, body)` (sort can now be FunctionSort or DependentSort)

## Depends on

- **CDDL grammar grow PR** (sibling of this in #263 sub-tree)

## Blocks

- 12 per-kit IR regen sub-issues — they mirror what the rust codegen produces

PR #427, opened 2026-05-06
spec: PinInvariantMemento — discharge Effect::PinnedReference (closes #395)

### LINKED ISSUE #395: [walk] PinInvariantMemento spec — discharge Effect::PinnedReference
## Goal

Ship a normative `PinInvariantMemento` spec at `protocol/specs/2026-05-05-pin-invariant-memento.md` that allows `Effect::PinnedReference` contracts to compose when a matching memento is in the verifier pool, and wire the discharge path in `provekit-walk`.

## Files affected

- Create: `protocol/specs/2026-05-05-pin-invariant-memento.md` — normative memento spec following the exact section structure of `protocol/specs/2026-05-05-loop-invariant-memento.md`
- Modify: `implementations/rust/provekit-walk/src/contract.rs` — extend `outer_non_opacity_pure` / `inner_non_opacity_pure` matches (around line 711) to also admit `Effect::PinnedReference` when a matching `PinInvariantMemento` is present in the pool; add `PinInvariantMemento` lookup to `OpacityMementoLookup` trait
- Test: `implementations/rust/provekit-walk/src/contract.rs` — add unit tests alongside the existing `compose_refuses_impure_contract` suite (around line 1328)

## What to ship

1. **Spec doc** (`protocol/specs/2026-05-05-pin-invariant-memento.md`) with these sections, mirroring the loop-invariant-memento format:
   - S0 Purpose: discharges `Effect::PinnedReference { target: T }`
   - S1 Memento structure (JCS key order is normative):
     ```json
     {
       "kind": "pin-invariant",
       "pinned_target": "<identifier of pinned formal or local>",
       "invariant": "<predicate the memento author asserts holds for the pinned cell>",
       "structural_pinning": true
     }
     ```
     Fields: `kind` (literal `"pin-invariant"`), `pinned_target` (string), `invariant` (string predicate), `structural_pinning` (bool: true = `!Unpin` discipline enforced structurally, false = projection-pinned).
   - S2 JCS canonical bytes and key order (`kind`, `pinned_target`, `invariant`, `structural_pinning`)
   - S3 Content-addressing (BLAKE3-512 of JCS bytes)
   - S4 Discharge rule: a contract carrying `Effect::PinnedReference { target: T }` composes iff a `PinInvariantMemento` whose `pinned_target == T` is present in the verifier's pool. `structural_pinning = false` is permitted for projection-pinned cases; the verifier does not validate the `invariant` string, only that the field is non-empty.
   - S5 Auto-minted mementos: the lifter does NOT auto-mint these. The memento author must supply them. The signed mementos travel in the contract bundle's `opacity_mementos` array.
   - S6 Worked examples:
     - `async fn f(pin: Pin<&mut Fut>)` where `Fut: Unpin`: lifter emits NO `PinnedReference` (Unpin means Pin is a no-op).
     - `async fn poll(pin: Pin<&mut Self>, cx: &mut Context)` where `Self: !Unpin`: lifter emits `Effect::PinnedReference { target: "pin" }`. Composition requires a `PinInvariantMemento` with `pinned_target = "pin"` and non-empty `invariant`.
   - S7 Out of scope v1: reborrow analysis of pinned projections; checking `invariant` string semantics.

2. **Rust discharge wiring** (`contract.rs`): in `compose_function_contracts_checked`, phase 2 check, extend the `outer_non_opacity_pure` / `inner_non_opacity_pure` `matches!` patterns to include `Effect::PinnedReference { .. }` when `pool.lookup_pin_invariant(target)` returns `Some`. Add `fn lookup_pin_invariant(&self, target: &str) -> Option<PinInvariantMementoView>` to `OpacityMementoLookup`. The `PinInvariantMementoView` struct carries `pinned_target: String`, `invariant: String`, `structural_pinning: bool`.

3. **Unit tests** (inline in `contract.rs`):
   - `compose_refuses_pinned_without_memento`: outer has `Effect::PinnedReference { target: "pin" }`, pool is empty, assert `compose_function_contracts_checked` returns `Ok(None)`.
   - `compose_succeeds_pinned_with_memento`: outer has `Effect::PinnedReference { target: "pin" }`, pool contains a matching `PinInvariantMemento`, assert `Ok(Some(...))`.
   - `compose_refuses_pinned_with_wrong_target_memento`: pool has memento for target `"other"`, assert `Ok(None)`.

## Tests required

1. `compose_refuses_pinned_without_memento` — assert: `compose_function_contracts_checked(&outer, &inner, 0, &empty_pool)` expected: `Ok(None)`
2. `compose_succeeds_pinned_with_memento` — assert: `compose_function_contracts_checked(&outer, &inner, 0, &pool_with_pin)` expected: `Ok(Some(_))`
3. `compose_refuses_pinned_with_wrong_target_memento` — assert: `compose_function_contracts_checked(&outer, &inner, 0, &pool_wrong_target)` expected: `Ok(None)`

- [ ] `cargo test -p provekit-walk` passes (existing tests + 3 new = prior count + 3)
- [ ] `make conformance` exits 0
- [ ] `protocol/specs/2026-05-05-pin-invariant-memento.md` exists with all 7 sections (S0-S6) matching the loop-invariant-memento section structure
- [ ] `Effect::PinnedReference` contract composes when matching memento is present (test 2 above)
- [ ] `Effect::PinnedReference` contract refuses when no memento is present (test 1 above)

## Out of scope

- Verifier-side wiring for `ProvenanceMemento` and `AtomicOrderingMemento` (separate issues #396, #397).
- Projection-pinned reborrow analysis.
- Per-kit IR types for the new memento (Rust only in this PR).
- Checking semantic correctness of the `invariant` string field.

## Dependencies

- Blocks: #398 (effect discharge classification spec will reference this)
- Blocked by: #394 (refuse-on-effects discipline, must be merged first)

## Single-PR scope

Yes

TSavo:
## Summary

Adds `PinInvariantMemento` discharge wiring for `Effect::PinnedReference`, moving it from unconditional block to opacity-dischargeable.

## What changed

- **Spec doc** at `protocol/specs/2026-05-05-pin-invariant-memento.md` — S0-S7 structure following the loop-invariant-memento format
- **`PinInvariantMementoView`** — lightweight pool lookup type (pinned_target, invariant, structural_pinning)
- **`OpacityError::PinInvariantNotDischarged { target }`** — new error variant
- **`OpacityMementoLookup::lookup_pin_invariant`** — trait extension, impl on EmptyOpacityPool and MockPool
- **`check_opacity_effects`** — `PinnedReference` is now opacity-dischargeable (checks pool for matching memento)
- **`compose_function_contracts_checked`** — `PinnedReference` added to phase 2 whitelist

## Tests

3 tests added:
1. `pinned_reference_without_memento_blocks_composition` — empty pool returns `PinInvariantNotDischarged`
2. `pinned_reference_with_memento_succeeds` — pool with matching memento composes successfully
3. `pinned_reference_with_wrong_target_memento_blocks` — pool with wrong-target memento still refuses

## Summary by CodeRabbit

* **New Features**
  * Added pin-invariant memento support for contract verification.
  * Enhanced opacity discharge with new verification capabilities for pinned references.

* **Documentation**
  * New protocol specification document defining pin-invariant memento format and verification rules.

### DIFF
--- a/implementations/rust/provekit-walk/src/contract.rs
+++ b/implementations/rust/provekit-walk/src/contract.rs
@@ -617,7 +617,11 @@ impl EffectSet {
                     }
                 }
                 Effect::PinnedReference { target } => {
-                    match pool.lookup_pin_invariant(function_cid.unwrap_or(""), target) {
+                    let fc = match function_cid {
+                        Some(cid) if !cid.is_empty() => cid,
+                        _ => return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() }),
+                    };
+                    match pool.lookup_pin_invariant(fc, target) {
                         Some(view) if !view.invariant.is_empty() => { /* discharged */ }
                         _ => return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() }),
                     }
--- a/implementations/rust/provekit-walk/src/contract.rs
+++ b/implementations/rust/provekit-walk/src/contract.rs
@@ -490,24 +490,25 @@ fn sort_to_value(s: &Sort) -> Arc<Value> {
 /// given opacity effect. Implemented by `MementoPool` in
 /// `provekit-verifier`; in-crate tests use a mock.
 ///
-/// Each method returns `true` iff a conforming discharge memento for
-/// that specific site is present in the pool and has passed all
-/// validation rules defined in the corresponding spec:
+/// The `has_*` methods return `true` iff a conforming discharge memento
+/// for that specific site is present in the pool and has passed all
+/// validation rules defined in the corresponding spec. The
+/// `lookup_pin_invariant` method returns `Option<PinInvariantMementoView>`
+/// to carry the discharge-relevant fields (function_cid, pinned_target,
+/// invariant) for downstream non-emptiness validation.
+///
+/// Specs:
 ///   - `LoopInvariantMemento`  (protocol/specs/2026-05-05-loop-invariant-memento.md)
 ///   - `TryBranchMemento`      (protocol/specs/2026-05-05-try-branch-memento.md)
 ///   - `ClosureBindingMemento` (protocol/specs/2026-05-05-closure-binding-memento.md)
+///   - `PinInvariantMemento`   (protocol/specs/2026-05-05-pin-invariant-memento.md)
 ///   - `UnresolvedCall`:       no memento kind yet; always undischarged.
 pub trait OpacityMementoLookup {
     fn has_loop_invariant(&self, loop_cid: &str) -> bool;
     fn has_try_branch(&self, try_cid: &str) -> bool;
     fn has_closure_binding(&self, body_fn_cid: &str) -> bool;
-    /// Returns true when the pool contains a contract for the
-    /// type being dropped (i.e., the drop function has been lifted
-    /// and its effects have been reviewed). When false, composition
-    /// is refused — the substrate does not silently assume drops are
-    /// effect-free.
     fn has_drop_contract(&self, type_name: &str) -> bool;
-    fn lookup_pin_invariant(&self, target: &str) -> Option<PinInvariantMementoView>;
+    fn lookup_pin_invariant(&self, function_cid: &str, target: &str) -> Option<PinInvariantMementoView>;
 }
 
 /// A no-op pool that never has any discharge mementos. Used as the
@@ -518,7 +519,7 @@ impl OpacityMementoLookup for EmptyOpacityPool {
     fn has_try_branch(&self, _: &str) -> bool { false }
     fn has_closure_binding(&self, _: &str) -> bool { false }
     fn has_drop_contract(&self, _: &str) -> bool { false }
-    fn lookup_pin_invariant(&self, _: &str) -> Option<PinInvariantMementoView> { None }
+    fn lookup_pin_invariant(&self, _function_cid: &str, _target: &str) -> Option<PinInvariantMementoView> { None }
 }
 
 /// Lightweight pool lookup type for PinInvariantMemento. The full
@@ -526,9 +527,9 @@ impl OpacityMementoLookup for EmptyOpacityPool {
 /// but the pool only needs the discharge-relevant fields.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct PinInvariantMementoView {
+    pub function_cid: String,
     pub pinned_target: String,
     pub invariant: String,
-    pub structural_pinning: bool,
 }
 
 /// Error returned when composition is refused because an opacity effect
@@ -578,13 +579,18 @@ impl EffectSet {
     ///   - there are no opacity effects at all, OR
     ///   - every opacity effect has a corresponding memento in `pool`.
     ///
+    /// The `function_cid` parameter is required for `PinnedReference`
+    /// discharge (which matches on `(function_cid, target)` pairs).
+    /// Pass `None` when the enclosing contract CID is not available
+    /// (in which case `PinnedReference` will always be undischarged).
+    ///
     /// Non-opacity effects (Reads/Writes/Io/Unsafe/Panics) are NOT
     /// checked here; composition with non-opacity effects is refused
     /// by `compose_function_contracts` separately.
     ///
     /// Returns the FIRST undischarged opacity effect as an error.
     /// The caller can collect all errors by iterating effects directly.
-    pub fn check_opacity_effects(&self, pool: &dyn OpacityMementoLookup) -> Result<(), OpacityError> {
+    pub fn check_opacity_effects(&self, pool: &dyn OpacityMementoLookup, function_cid: Option<&str>) -> Result<(), OpacityError> {
         for effect in &self.effects {
             match effect {
                 Effect::OpaqueLoop { loop_cid } => {
@@ -611,8 +617,9 @@ impl EffectSet {
                     }
                 }
                 Effect::PinnedReference { target } => {
-                    if pool.lookup_pin_invariant(target).is_none() {
-                        return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() });
+                    match pool.lookup_pin_invariant(function_cid.unwrap_or(""), target) {
+                        Some(view) if !view.invariant.is_empty() => { /* discharged */ }
+                        _ => return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() }),
                     }
                 }
                 // Non-opacity effects: Reads/Writes/Io/Unsafe/Panics and the
@@ -728,9 +735,9 @@ pub fn compose_function_contracts(
 /// refusal reasons:
 ///
 /// 1. `Err(OpacityError::*)` — a contract carries an opacity effect
-///    (`OpaqueLoop`, `EarlyReturn`, `ClosureCapture`, `UnresolvedCall`)
-///    with no corresponding discharge memento in `pool`. The caller
-///    knows exactly which effect is undischarged.
+///    (`OpaqueLoop`, `EarlyReturn`, `ClosureCapture`, `UnresolvedCall`,
+///    `Drop`, `PinnedReference`) with no corresponding discharge memento
+///    in `pool`. The caller knows exactly which effect is undischarged.
 ///
 /// 2. `Ok(None)` — composition refused for a non-opacity reason:
 ///    an unconditionally-blocking effect (Reads/Writes/Io/Unsafe/Panics)
@@ -754,11 +761,12 @@ pub fn compose_function_contracts_checked(
 ) -> Result<Option<ComposedFunctionContract>, OpacityError> {
     // Phase 1: check opacity effects on both sides. Returns the first
     // undischarged opacity effect as a typed error.
-    outer.effects.check_opacity_effects(pool)?;
-    inner.effects.check_opacity_effects(pool)?;
+    outer.effects.check_opacity_effects(pool, Some(&outer.cid))?;
+    inner.effects.check_opacity_effects(pool, Some(&inner.cid))?;
 
     // Phase 2: check for unconditionally-blocking effects. After opacity
-    // discharge, only Reads/Writes/Io/Unsafe/Panics can still block.
+    // and pin-invariant discharge, only Reads/Writes/Io/Unsafe/Panics and
+    // the remaining structural type-boundary effects can still block.
     // A contract that had ONLY opacity effects — all now discharged — has
     // no remaining blockers; one that still has non-opacity impure effects
     // is still refused.
@@ -1501,12 +1509,12 @@ mod tests {
         loop_cids: Vec<String>,
         try_cids: Vec<String>,
         body_fn_cids: Vec<String>,
-        pin_invariant_targets: Vec<String>,
+        pin_invariant_targets: std::collections::HashMap<String, String>,
     }
 
     impl MockPool {
         fn empty() -> Self {
-            Self { loop_cids: vec![], try_cids: vec![], body_fn_cids: vec![], pin_invariant_targets: vec![] }
+            Self { loop_cids: vec![], try_cids: vec![], body_fn_cids: vec![], pin_invariant_targets: std::collections::HashMap::new() }
         }
         fn with_loop(mut self, cid: &str) -> Self {
             self.loop_cids.push(cid.to_string());
@@ -1521,7 +1529,11 @@ mod tests {
             self
         }
         fn with_pin_invariant(mut self, target: &str) -> Self {
-            self.pin_invariant_targets.push(target.to_string());
+            self.pin_invariant_targets.insert(target.to_string(), "true".to_string());
+            self
+        }
+        fn with_pin_invariant_empty(mut self, target: &str) -> Self {
+            self.pin_invariant_targets.insert(target.to_string(), "".to_string());
             self
         }
     }
@@ -1539,12 +1551,12 @@ mod tests {
         fn has_drop_contract(&self, _type_name: &str) -> bool {
             false // no drop contracts in mock pool
         }
-        fn lookup_pin_invariant(&self, target: &str) -> Option<PinInvariantMementoView> {
-            if self.pin_invariant_targets.iter().any(|t| t == target) {
+        fn lookup_pin_invariant(&self, _function_cid: &str, target: &str) -> Option<PinInvariantMementoView> {
+            if let Some(invariant) = self.pin_invariant_targets.get(target).cloned() {
                 Some(PinInvariantMementoView {
+                    function_cid: _function_cid.to_string(),
                     pinned_target: target.to_string(),
-                    invariant: "true".to_string(),
-                    structural_pinning: true,
+                    invariant,
                 })
             } else {
                 None
@@ -1695,7 +1707,7 @@ mod tests {
     fn check_opacity_pure_contract_is_ok() {
         let f = build_function_contract(&parse_fn(r#"fn f(x: u32) -> u32 { x * 2 }"#), None);
         let pool = MockPool::empty();
-        assert!(f.effects.check_opacity_effects(&pool).is_ok());
+        assert!(f.effects.check_opacity_effects(&pool, Some(&f.cid)).is_ok());
     }
 
     // ---- Issue #395: PinInvariantMemento discharge tests ----
@@ -1748,6 +1760,23 @@ mod tests {
         assert!(
             matches!(result, Err(OpacityError::PinInvariantNotDischarged { .. })),
             "expected PinInvariantNotDischarged for wrong target, got {:?}", result
+        );
+    }
+
+    /// Pool has a PinInvariantMemento with empty invariant -> still blocks
+    /// because non-emptiness is required by the spec (§3.2).
+    #[test]
+    fn pinned_reference_with_empty_invariant_blocks() {
+        let outer = contract_with_effects(
+            "outer",
+            vec![Effect::PinnedReference { target: "pin".to_string() }],
+        );
+        let inner = build_function_contract(&parse_fn(r#"fn inner(y: u32) -> u32 { y + 1 }"#), None);
+        let pool = MockPool::empty().with_pin_invariant_empty("pin");
+        let result = compose_function_contracts_checked(&outer, &inner, 0, &pool);
+        assert!(
+            matches!(result, Err(OpacityError::PinInvariantNotDischarged { .. })),
+            "expected PinInvariantNotDischarged for empty invariant, got {:?}", result
         );
     }
 }
--- a/implementations/rust/provekit-walk/src/contract.rs
+++ b/implementations/rust/provekit-walk/src/contract.rs
@@ -507,6 +507,7 @@ pub trait OpacityMementoLookup {
     /// is refused — the substrate does not silently assume drops are
     /// effect-free.
     fn has_drop_contract(&self, type_name: &str) -> bool;
+    fn lookup_pin_invariant(&self, target: &str) -> Option<PinInvariantMementoView>;
 }
 
 /// A no-op pool that never has any discharge mementos. Used as the
@@ -517,6 +518,17 @@ impl OpacityMementoLookup for EmptyOpacityPool {
     fn has_try_branch(&self, _: &str) -> bool { false }
     fn has_closure_binding(&self, _: &str) -> bool { false }
     fn has_drop_contract(&self, _: &str) -> bool { false }
+    fn lookup_pin_invariant(&self, _: &str) -> Option<PinInvariantMementoView> { None }
+}
+
+/// Lightweight pool lookup type for PinInvariantMemento. The full
+/// memento has envelope/header/metadata layers (per the memento spec),
+/// but the pool only needs the discharge-relevant fields.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PinInvariantMementoView {
+    pub pinned_target: String,
+    pub invariant: String,
+    pub structural_pinning: bool,
 }
 
 /// Error returned when composition is refused because an opacity effect
@@ -536,6 +548,9 @@ pub enum OpacityError {
     /// memento kind exists yet; composition is always refused.
     UnresolvedCallNotDischarged { name: String },
     DropNotDischarged { name: String },
+    /// An `Effect::PinnedReference { target }` is present but no
+    /// `PinInvariantMemento` with matching `pinnedTarget` is in the pool.
+    PinInvariantNotDischarged { target: String },
 }
 
 impl std::fmt::Display for OpacityError {
@@ -551,6 +566,8 @@ impl std::fmt::Display for OpacityError {
                 write!(f, "opacity: UnresolvedCall({name}) has no discharge memento kind"),
             Self::DropNotDischarged { name } =>
                 write!(f, "opacity: Drop({name}) has no lifted drop function in pool"),
+            Self::PinInvariantNotDischarged { target } =>
+                write!(f, "opacity: PinnedReference(target={target}) has no PinInvariantMemento in pool"),
         }
     }
 }
@@ -593,15 +610,18 @@ impl EffectSet {
                         return Err(OpacityError::DropNotDischarged { name: name.clone() });
                     }
                 }
+                Effect::PinnedReference { target } => {
+                    if pool.lookup_pin_invariant(target).is_none() {
+                        return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() });
+                    }
+                }
                 // Non-opacity effects: Reads/Writes/Io/Unsafe/Panics and the
-                // structural type-boundary effects (PinnedReference,
-                // RawPointerProvenance, AtomicAccess) do not participate in
-                // memento-based discharge. They unconditionally block composition
-                // via the outer_non_opacity_pure check in
-                // compose_function_contracts_checked.
+                // structural type-boundary effects (RawPointerProvenance,
+                // AtomicAccess) do not participate in memento-based discharge.
+                // They unconditionally block composition via the
+                // outer_non_opacity_pure check in compose_function_contracts_checked.
                 Effect::Reads { .. } | Effect::Writes { .. } | Effect::Io
                 | Effect::Unsafe | Effect::Panics
-                | Effect::PinnedReference { .. }
                 | Effect::RawPointerProvenance { .. }
                 | Effect::AtomicAccess { .. } => {}
             }
@@ -746,13 +766,13 @@ pub fn compose_function_contracts_checked(
         e,
         Effect::OpaqueLoop { .. } | Effect::EarlyReturn { .. }
         | Effect::ClosureCapture { .. } | Effect::UnresolvedCall { .. }
-        | Effect::Drop { .. }
+        | Effect::Drop { .. } | Effect::PinnedReference { .. }
     ));
     let inner_non_opacity_pure = inner.effects.effects.iter().all(|e| matches!(
         e,
         Effect::OpaqueLoop { .. } | Effect::EarlyReturn { .. }
         | Effect::ClosureCapture { .. } | Effect::UnresolvedCall { .. }
-        | Effect::Drop { .. }
+        | Effect::Drop { .. } | Effect::PinnedReference { .. }
     ));
     if !outer_non_opacity_pure || !inner_non_opacity_pure {
         return Ok(None);
@@ -1481,11 +1501,12 @@ mod tests {
         loop_cids: Vec<String>,
         try_cids: Vec<String>,
         body_fn_cids: Vec<String>,
+        pin_invariant_targets: Vec<String>,
     }
 
     impl MockPool {
         fn empty() -> Self {
-            Self { loop_cids: vec![], try_cids: vec![], body_fn_cids: vec![] }
+            Self { loop_cids: vec![], try_cids: vec![], body_fn_cids: vec![], pin_invariant_targets: vec![] }
         }
         fn with_loop(mut self, cid: &str) -> Self {
             self.loop_cids.push(cid.to_string());
@@ -1497,6 +1518,10 @@ mod tests {
         }
         fn with_closure(mut self, cid: &str) -> Self {
             self.body_fn_cids.push(cid.to_string());
+            self
+        }
+        fn with_pin_invariant(mut self, target: &str) -> Self {
+            self.pin_invariant_targets.push(target.to_string());
             self
         }
     }
@@ -1513,6 +1538,17 @@ mod tests {
         }
         fn has_drop_contract(&self, _type_name: &str) -> bool {
             false // no drop contracts in mock pool
+        }
+        fn lookup_pin_invariant(&self, target: &str) -> Option<PinInvariantMementoView> {
+            if self.pin_invariant_targets.iter().any(|t| t == target) {
+                Some(PinInvariantMementoView {
+                    pinned_target: target.to_string(),
+                    invariant: "true".to_string(),
+                    structural_pinning: true,
+                })
+            } else {
+                None
+            }
         }
     }
 
@@ -1660,5 +1696,58 @@ mod tests {
         let f = build_function_contract(&parse_fn(r#"fn f(x: u32) -> u32 { x * 2 }"#), None);
         let pool = MockPool::empty();
         assert!(f.effects.check_opacity_effects(&pool).is_ok());
+    }
+
+    // ---- Issue #395: PinInvariantMemento discharge tests ----
+
+    /// A contract with PinnedReference + no PinInvariantMemento in pool ->
+    /// compose_function_contracts_checked returns Err(PinInvariantNotDischarged).
+    #[test]
+    fn pinned_reference_without_memento_blocks_composition() {
+        let outer = contract_with_effects(
+            "outer",
+            vec![Effect::PinnedReference { target: "pin".to_string() }],
+        );
+        let inner = build_function_contract(&parse_fn(r#"fn inner(y: u32) -> u32 { y + 1 }"#), None);
+        let pool = MockPool::empty();
+        let result = compose_function_contracts_checked(&outer, &inner, 0, &pool);
+        assert!(
+            matches!(result, Err(OpacityError::PinInvariantNotDischarged { .. })),
+            "expected PinInvariantNotDischarged, got {:?}", result
+        );
+    }
+
+    /// Same contract with PinnedReference + matching PinInvariantMemento in pool ->
+    /// compose_function_contracts_checked succeeds.
+    #[test]
+    fn pinned_reference_with_memento_succeeds() {
+        let outer = contract_with_effects(
+            "outer",
+            vec![Effect::PinnedReference { target: "pin".to_string() }],
+        );
+        let inner = build_function_contract(&parse_fn(r#"fn inner(y: u32) -> u32 { y + 1 }"#), None);
+        let pool = MockPool::empty().with_pin_invariant("pin");
+        let result = compose_function_contracts_checked(&outer, &inner, 0, &pool);
+        match result {
+            Ok(Some(_)) => { /* success */ }
+            other => panic!("expected Ok(Some(_)) after PinInvariantMemento discharge, got {:?}", other),
+        }
+    }
+
+    /// Pool has a PinInvariantMemento for a different target -> composed_function_contracts_checked
+    /// returns Err(PinInvariantNotDischarged).
+    #[test]
+    fn pinned_reference_with_wrong_target_memento_blocks() {
+        let outer = contract_with_effects(
+            "outer",
+            vec![Effect::PinnedReference { target: "pin".to_string() }],
+        );
+        let inner = build_function_contract(&parse_fn(r#"fn inner(y: u32) -> u32 { y + 1 }"#), None);
+        let pool = MockPool::empty().with_pin_invariant("other");
+        let result = compose_function_contracts_checked(&outer, &inner, 0, &pool);
+        assert!(
+            matches!(result, Err(OpacityError::PinInvariantNotDischarged { .. })),
+            "expected PinInvariantNotDischarged for wrong target, got {:?}", result
+        );
     }
 }
--- a/protocol/specs/2026-05-05-pin-invariant-memento.md
+++ b/protocol/specs/2026-05-05-pin-invariant-memento.md
@@ -7,43 +7,82 @@
 - `2026-05-05-loop-invariant-memento.md` — sibling discharge memento for `Effect::OpaqueLoop`
 - `2026-05-05-try-branch-memento.md` — sibling discharge memento for `Effect::EarlyReturn`
 - `2026-05-05-closure-binding-memento.md` — sibling discharge memento for `Effect::ClosureCapture`
-- `2026-04-30-ir-formal-grammar.md` — IrFormula shape; `invariant` is an IrFormula string
+- `2026-04-30-ir-formal-grammar.md` — IrFormula shape; `invariant` is an IrFormula
 - `2026-05-04-linker-daemon-protocol.md` — the linker daemon pools discharge mementos for the substrate
+- `protocol/provekit-ir.cddl` — CDDL grammar entry locking the canonical wire shape
 - #394 — refuse-on-effects discipline this spec hooks into
 
 ---
 
 ## §0. Purpose
 
-A `FunctionContractMemento` carrying an `Effect::PinnedReference { target }` is opaque at that pinned formal: the substrate cannot reason about the pinned cell's invariance without additional evidence. A `PinInvariantMemento` is the discharge certificate for a single pinned target: it asserts the invariant the pinned cell satisfies and whether the pinning is structural (`!Unpin`) or projection-based.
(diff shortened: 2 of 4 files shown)


protocol/specs/2026-05-05-pin-invariant-memento.md
```diff
@@ -0,0 +1,162 @@
...
+```
+cid = "blake3-512:" ++ hex(BLAKE3-512(JCS({ "kind": "pin-invariant", "invariant": <s>, "pinnedTarget": <s>, "structuralPinning": <bool> })))
+```
+
+The `invariant` string is included as a raw string value in the JCS object, not hashed separately. This is a deliberate design choice: the invariant IS the content of the memento; a property-hash layer is unnecessary for a memento with a single opaque predicate.
+
+### §2.2 JCS key order
+
+JCS sorts keys lexicographically: `invariant`, `kind`, `pinnedTarget`, `structuralPinning`. Implementations MUST emit keys in this order. The `kind` field is NOT first in the JCS sort (unlike the loop-invariant memento where key order is declared normatively); the canonical order is determined by JCS string comparison of key names.
+
```
TSavo, line 111: **P1 — Wire format divergence: §2.2 self-admits cross-spec JCS key inconsistency**

§2.2 states: _"The `kind` field is NOT first in the JCS sort (unlike the loop-invariant memento where key order is declared normatively)"_.

This is a load-bearing inconsistency. The `LoopInvariantMemento`, `TryBranchMemento`, and `ClosureBindingMemento` declare key order normatively, independent of JCS lex-sort. This spec instead defers to pure JCS lex-sort, producing `invariant, kind, pinnedTarget, structuralPinning`. These are two different conventions within the same memento family.

**Why this breaks federation:** Two kits implementing the substrate will look at one of two things: the normative declared order in the spec, or the JCS algorithm. When they diverge, CIDs from Kit A and Kit B differ for the same memento, and discharge fails federation-wide. The author noticed the divergence in this section and shipped it anyway. That is the exact failure mode paper 07 §6 (every step is mechanical, inspectable, signable) is designed to prevent.

**Fix shape:** Pick one convention and apply consistently across all sibling discharge mementos. Either (a) add a CDDL entry that locks key order by declaration as siblings do, or (b) go back and remove the normative-order language from sibling specs and let all defer to JCS lex-sort. Do not mix conventions within the same memento family.

**Acceptance:** §2.2 references a sibling-consistency audit confirming this spec and `loop-invariant-memento.md` produce byte-identical output under a single convention. A CDDL grammar entry is added to `protocol/provekit-ir.cddl` locking the canonical shape.

TSavo, line 111: **P1 — No CDDL grammar entry in `protocol/provekit-ir.cddl`**

The sibling discharge mementos that have shipped (`LoopInvariantMemento`, `TryBranchMemento`) each have a corresponding CDDL slot in `protocol/provekit-ir.cddl` that locks the canonical wire shape. This PR ships a normative spec and a Rust implementation but adds no CDDL entry.

**Why this matters:** CDDL is what non-Rust kits (Python, Ruby, C#, Go, TypeScript, etc.) parse to produce byte-identical mementos. Without a CDDL slot, each kit authors this memento independently from the spec prose, and the spec prose has already demonstrated a key-order ambiguity (§2.2). Two kits WILL diverge. The conformance suite (`protocol/conformance/`) has no fixture to catch it.

**Fix shape:** Add a `PinInvariantMemento` CDDL rule to `protocol/provekit-ir.cddl` with explicit key declaration order, and add a byte-fixture to `protocol/conformance/` with at least one known-CID example. The fix should follow the same pattern used for `FunctionSort` (#336) and `DependentSort` (#330).

**Acceptance:** `grep -l "pin-invariant" protocol/provekit-ir.cddl` returns a hit. `make conformance` includes a PinInvariantMemento fixture test.

protocol/specs/2026-05-05-pin-invariant-memento.md
```diff
@@ -0,0 +1,162 @@
...
+{
+  "kind": "pin-invariant",
+  "invariant": "<ir-formula predicate string>",
+  "pinnedTarget": "<identifier of pinned formal or local>",
+  "structuralPinning": true
+}
+```
+
+### §1.2 Field semantics
+
```
TSavo, line 73: **P1 -- `pinnedTarget` is a bare local name: CID instability and type-unsound discharge**

§1.2 defines `pinnedTarget` as "the formal or local name that the `Effect::PinnedReference { target }` refers to." The worked example uses `pinnedTarget = "pin"` as the match key.

This has two distinct problems:

**CID instability:** If a function is refactored and its pinned parameter is renamed from `pin` to `pinned_self` or `cx`, the existing `PinInvariantMemento` with `pinnedTarget = "pin"` no longer matches. The discharge breaks. Any caller relying on a cached CID for that memento must re-author the memento. This violates the substrate's core stability guarantee (paper 07 §6: content-addressing-safe -- existing CIDs unchanged). The name IS the content; rename = new CID = discharge breaks for all dependents.

**Type-unsound discharge:** A memento authored for `fn poll(pin: Pin<&mut MyFuture>)` will also discharge `fn poll(pin: Pin<&mut MyOtherFuture>)` if that function happens to use the same parameter name. The verifier matches `pinnedTarget` against the effect `target` string exactly, and both produce `"pin"`. Nothing in the memento anchors it to the pinned type. Two semantically different Pin invariants share the same match key.

**What the issue spec says:** Issue #395 uses `<identifier of pinned formal or local>` -- implying the same bare-name approach. But the issue spec was authored alongside this PR and carries the same gap.

**Fix shape:** `pinnedTarget` must carry enough type identity to prevent false discharge. Minimum viable fix: include the type's fully-qualified Charon `def_id` or a stable type fingerprint alongside the name. Alternatively, scope the match to `(function_cid, parameter_name)` pairs so the memento is anchored to the specific function contract, not a free-floating name string.

**Acceptance:** A test `pinned_reference_same_name_different_type_does_not_discharge` demonstrates that two `PinnedReference` effects with the same target name but different pinned types are not satisfied by the same memento.

implementations/rust/provekit-walk/src/contract.rs
```diff
@@ -566,15 +583,18 @@ impl EffectSet {
                 Effect::UnresolvedCall { name } => {
                     return Err(OpacityError::UnresolvedCallNotDischarged { name: name.clone() });
                 }
+                Effect::PinnedReference { target } => {
+                    if pool.lookup_pin_invariant(target).is_none() {
+                        return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() });
```
TSavo: **P1 -- Empty `invariant` string discharges `PinnedReference` (codex flagged this; escalating with acceptance criterion)**

The spec (§3.2, §0) is explicit: the verifier checks that the `invariant` field is non-empty before discharging. This code only checks `is_none()` -- meaning a pool entry with `invariant: ""` satisfies the lookup and clears the effect. An empty-invariant memento is a malformed certificate that the spec forbids. The verifier is supposed to reject it here.

This is not a future-concern. The `MockPool` in tests hardcodes `invariant: "true"`, so no test exercises the empty-invariant path. The gap is invisible in CI.

Fix shape: change the discharge condition from `lookup_pin_invariant(target).is_none()` to:

```rust
match pool.lookup_pin_invariant(target) {
    Some(view) if !view.invariant.is_empty() => { /* discharged */ }
    _ => return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() }),
}
```

Acceptance: a test `pinned_reference_with_empty_invariant_blocks` passes a `MockPool` with `invariant: ""` and asserts `Err(PinInvariantNotDischarged)`.

implementations/rust/provekit-walk/src/contract.rs
```diff
@@ -489,6 +489,7 @@ pub trait OpacityMementoLookup {
     fn has_loop_invariant(&self, loop_cid: &str) -> bool;
     fn has_try_branch(&self, try_cid: &str) -> bool;
     fn has_closure_binding(&self, body_fn_cid: &str) -> bool;
+    fn lookup_pin_invariant(&self, target: &str) -> Option<PinInvariantMementoView>;
```
TSavo: **P1 -- `MementoPool` in `provekit-verifier/src/types.rs` is not updated: compile break**

The `OpacityMementoLookup` trait gains `fn lookup_pin_invariant`. There are three implementors (confirmed from codebase grep):
- `EmptyOpacityPool` -- updated in this PR
- `MockPool` (tests) -- updated in this PR
- `MementoPool` in `implementations/rust/provekit-verifier/src/types.rs` -- NOT updated

This PR will fail to compile once the trait change lands because `MementoPool` still only implements the `has_*` methods. Copilot flagged this at line 492; I'm restating because the failure mode is a broken build in the verifier crate, not just the walk crate.

Fix shape: extend `MementoPool` in `provekit-verifier/src/types.rs` to implement `lookup_pin_invariant`. The implementation needs to index by `pinned_target` string and return the deserialized view fields from whatever memento storage the pool uses.

Acceptance: `cargo build -p provekit-verifier` exits 0 with this PR applied.

protocol/specs/2026-05-05-pin-invariant-memento.md
```diff
@@ -0,0 +1,162 @@
...
+| Field               | Required | Meaning |
+|---------------------|----------|---------|
```
TSavo: **P2 -- `structuralPinning` is CID-load-bearing but has no effect on composition**

§1.2 says: "The verifier records this field but does not gate composition on it; it is informational for future Pin-projection verifier extensions."

This means two mementos -- identical in every way except `structuralPinning: true` vs `structuralPinning: false` -- produce different CIDs but discharge identically. The field participates in content-addressing without participating in any correctness decision.

That is the definition of a decorative hash from the feedback note on unit-of-work = cache boundary: "Cached fragments that need re-verification are decorative hashes, not operational ones." A memento author who sets `structuralPinning: false` when the type is actually `!Unpin` gets a different CID but the same composition outcome. No gate catches the error.

This is also a cross-spec consistency problem. The `DropMemento` in #426 has `DropKind` (Trivial/Structural/UserCode) that directly gates discharge behavior -- not just records intent. Pin and Drop are sibling effects in the same family. The structural distinction is load-bearing in Drop and purely informational in Pin.

Fix shape, two options:
1. Remove `structuralPinning` from the v1 wire format entirely (scope it to the future Pin-projection spec), so the field is not a dormant CID variable. Add it back when the verifier actually uses it.
2. If the field stays, add a verifier gate: structural-pinned mementos (`structuralPinning: true`) may not discharge effects on types that are `Unpin`, and projection-pinned mementos (`structuralPinning: false`) require the effect to include a projection path.

Acceptance: The spec states explicitly which option is chosen and why, with a forward-reference to the ticket that will add Pin-projection gating.

protocol/specs/2026-05-05-pin-invariant-memento.md
```diff
@@ -0,0 +1,162 @@
...
+**Closes:** #395
+**Related:**
+- `2026-05-05-loop-invariant-memento.md` — sibling discharge memento for `Effect::OpaqueLoop`
+- `2026-05-05-try-branch-memento.md` — sibling discharge memento for `Effect::EarlyReturn`
+- `2026-05-05-closure-binding-memento.md` — sibling discharge memento for `Effect::ClosureCapture`
+- `2026-04-30-ir-formal-grammar.md` — IrFormula shape; `invariant` is an IrFormula string
+- `2026-05-04-linker-daemon-protocol.md` — the linker daemon pools discharge mementos for the substrate
+- #394 — refuse-on-effects discipline this spec hooks into
+
+---
```
TSavo, line 15: **P2 -- No signing authority or trust model in spec**

Sibling memento specs that have shipped (and the paper 07 §6 description of the substrate) are explicit that each signed edge in the proof graph carries an Ed25519 signature over canonical bytes. The provenance key lives in Vault at `secret/provekit/provenance-ed25519`.

This spec (§0 through §7) says nothing about:
- Who is authorized to sign a `PinInvariantMemento`
- What the signing ceremony looks like
- How the verifier validates the signature before trusting the memento
- What happens when `core::pin::Pin` is renamed or re-exported in a future stdlib version and the memento was authored against the old name

The omission means mementos from any source are accepted. There is no difference between a memento authored by a trusted curator who actually audited the Pin invariant, and a memento injected by a build tool that mints one for every `PinnedReference` it sees. The discharge gate becomes a rubber stamp.

This is not out-of-scope: the issue #395 AC table does not include "signing authority defined" but it is an architectural requirement from paper 07 §6 and from the established signing pattern in the substrate.

Fix shape: add a §S_sign section (matching siblings) that names the signing authority (curator-level vs. project-level), references the Ed25519 key path, and states the verification step the pool performs before a memento is admitted.

Acceptance: The spec includes a signing section. A note on stdlib rename resilience (what happens if `core::pin::Pin` changes name) is present even if the answer is "the memento author must re-issue."

protocol/specs/2026-05-05-pin-invariant-memento.md
```diff
@@ -0,0 +1,162 @@
...
+
+```rust
+struct MyFuture { state: u32 }
+impl !Unpin for MyFuture {}
+fn poll(pin: Pin<&mut MyFuture>) -> Poll<()> {
+    let this = unsafe { pin.get_unchecked_mut() };
+    this.state += 1;
+    Poll::Ready(())
+}
+```
```
TSavo, line 192: **P2 -- §5.2 worked example uses `unsafe { pin.get_unchecked_mut() }` but the spec does not explain what the verifier sees**

The §5.2 example shows:
```rust
fn poll(pin: Pin<&mut MyFuture>) -> Poll<()> {
    let this = unsafe { pin.get_unchecked_mut() };
    this.state += 1;
    Poll::Ready(())
}
```

This function body contains an `unsafe` block AND a `Writes { target }` effect on `this.state`. The spec asserts that supplying a `PinInvariantMemento` for `"pin"` discharges `Effect::PinnedReference` and allows composition. But the function also carries `Effect::Unsafe` (unconditional block) and `Effect::Writes { target: "this.state" }` (unconditional block).

A reader implementing a kit against this spec will conclude: "supply the PinInvariantMemento and this function composes." That is wrong. The unsafe block and writes still prevent composition. The example does not show the full effect set the lifter actually emits for this function, so it teaches an incomplete picture of what discharge accomplishes.

This is not an edge case -- it is the canonical use of `pin.get_unchecked_mut()` in every hand-rolled future. The worked example is the primary teaching vehicle for this spec.

Fix shape: Show the complete effect set for the §5.2 function (`PinnedReference`, `Unsafe`, `Writes`). Explain which effects the `PinInvariantMemento` discharges and which do not, so the reader understands the actual composition outcome.

Acceptance: §5.2 lists the full effect set emitted by the lifter for this function and explicitly states whether composition succeeds or fails after memento discharge.

implementations/rust/provekit-walk/src/contract.rs
TSavo: **P3 -- Trait doc comment says "returns true iff" but new method returns Option**

The `OpacityMementoLookup` trait doc comment was written when all methods returned `bool`. It now says something like "Each method returns `true` iff ..." but `lookup_pin_invariant` returns `Option<PinInvariantMementoView>`.

Copilot flagged this at line 492. It is a nit but the trait doc is read by every kit implementor; misleading documentation about return types causes bugs.

Fix shape: Update the trait-level doc comment to distinguish: the `has_*` methods return `bool` for presence checks; `lookup_pin_invariant` returns `Option<PinInvariantMementoView>` carrying the discharge-relevant fields.

Acceptance: The trait doc accurately describes all method signatures.

implementations/rust/provekit-walk/src/contract.rs
```diff
@@ -719,11 +739,13 @@ pub fn compose_function_contracts_checked(
         e,
         Effect::OpaqueLoop { .. } | Effect::EarlyReturn { .. }
         | Effect::ClosureCapture { .. } | Effect::UnresolvedCall { .. }
+        | Effect::PinnedReference { .. }
     ));
```
TSavo, line 782: **P3 -- `compose_function_contracts_checked` phase 2 doc comment still lists only 4 opacity effects**

The function's doc comment (around line 739) describes phase 2 as handling `OpaqueLoop`, `EarlyReturn`, `ClosureCapture`, `UnresolvedCall`. `PinnedReference` is added to the phase 2 whitelist in this PR but not listed in the comment.

This is the doc that callers and kit implementors read to understand which effects are dischargeable. A stale list here causes incorrect kit implementations.

Fix shape: Add `PinnedReference` to the phase 2 list in the function doc comment.

Acceptance: The doc comment phase 2 list matches the actual whitelist match arm.

TSavo:
## Review summary -- held until P1s addressed

**What I read:** Full diff (260 lines), issue #395, PR #426 (`DropMemento`) for cross-spec comparison, the `PinnedReference` grep in the Rust codebase, paper 07 §6, and all existing bot comments. I confirmed three `OpacityMementoLookup` implementors from the codebase (EmptyOpacityPool, MockPool, MementoPool).

---

### P1 -- Correctness / soundness / breaks the thesis

**P1-A (spec line 65): JCS key order convention divergence across sibling mementos** -- §2.2 self-admits that `PinInvariantMemento` uses lex-sort convention while sibling mementos (`LoopInvariantMemento` etc.) declare key order normatively. Two kits following different conventions produce different CIDs for the same memento. Federation breaks. The fix is to pick one convention for the entire discharge memento family and apply it consistently, WITH a CDDL entry that locks the choice.

**P1-B (spec line 65): No CDDL grammar entry** -- Sibling mementos have CDDL slots in `protocol/provekit-ir.cddl` that all 11 non-Rust kits parse. This PR has none. Without a CDDL entry, every non-Rust kit that ships a `PinInvariantMemento` will diverge from the Rust reference. The spec prose alone (especially given the §2.2 ambiguity) is not enough to lock the wire format.

**P1-C (spec line 40): `pinnedTarget` is a bare local name -- CID instability and type-unsound discharge** -- The match key is the raw parameter name string (e.g., `"pin"`). Renaming the parameter breaks all existing mementos and their dependents. Worse: any function with a parameter named `"pin"` -- regardless of the pinned type -- is discharged by the same memento. A memento for `fn poll(pin: Pin<&mut MyFuture>)` also discharges `fn poll(pin: Pin<&mut OtherFuture>)`. No type identity anchors the discharge.

**P1-D (contract.rs line 588): Empty `invariant` string passes discharge** -- The spec (§0, §3.2) requires `invariant` to be non-empty before discharge. The code only checks `is_none()`. A pool entry with `invariant: ""` satisfies `Some(...)` and clears the effect. The `MockPool` hardcodes `invariant: "true"` so no test catches this. Fix: check `!view.invariant.is_empty()` before accepting the discharge.

**P1-E (contract.rs line 492): `MementoPool` in `provekit-verifier/src/types.rs` not updated -- compile break** -- Three implementors of `OpacityMementoLookup` exist in the codebase. Two are updated in this PR. `MementoPool` in `provekit-verifier/src/types.rs` is not. The verifier crate will fail to build. Confirmed from codebase grep: `./implementations/rust/provekit-verifier/src/types.rs:477:impl OpacityMementoLookup for MementoPool`.

---

### P2 -- Clarity / incompleteness / will cause kit divergence

**P2-A (spec line 42): `structuralPinning` participates in CID but not in composition logic** -- Two mementos differing only in `structuralPinning` produce different CIDs but identical discharge behavior. This is a decorative hash (see feedback note: unit-of-work = cache boundary). The `DropMemento` in #426 treats the equivalent field (`DropKind`) as load-bearing. Either gate composition on `structuralPinning` or remove it from v1 wire format until the Pin-projection verifier extension ships.

**P2-B (spec line 14): No signing authority or trust model** -- The spec says nothing about who signs mementos, how the verifier validates signatures, or what happens if `core::pin::Pin` is renamed in a future stdlib. Without this, any build tool that auto-mints `PinInvariantMemento` for every `PinnedReference` it sees passes through as a trusted discharge certificate. Paper 07 §6 requires Ed25519 signing over canonical bytes as part of the mechanical chain.

**P2-C (spec line 116): §5.2 worked example teaches incomplete composition picture** -- The canonical `get_unchecked_mut()` example carries `Effect::Unsafe` and `Effect::Writes` in addition to `PinnedReference`. The spec implies "supply the memento and this composes." It will not. The other two effects unconditionally block composition. The worked example must show the full effect set and clarify what discharge actually accomplishes for this function.

---

### P3 -- Nits (should fix before merge)

**P3-A (contract.rs line 492): Trait doc comment says "returns true iff" but `lookup_pin_invariant` returns `Option`** -- Misleading for kit implementors.

**P3-B (contract.rs line 743): `compose_function_contracts_checked` phase 2 doc comment not updated** -- Still lists only 4 opacity effects; `PinnedReference` added in this PR is missing from the list.

---

### What is working

- The three-case test coverage (no memento, matching memento, wrong-target memento) is the right shape and tests the right behavior.
- The `PinInvariantMementoView` struct design is clean and the `EmptyOpacityPool` / `MockPool` update pattern follows the established convention.
- `OpacityError::PinInvariantNotDischarged` naming is consistent with siblings.
- §3.3 (Unpin types: no effect emitted) is correct and important.
- The decision NOT to auto-mint is correct.

---

### Verdict: held until P1s addressed

P1-A and P1-B together mean the wire format is not federation-safe. P1-C means the discharge match key is not sound. P1-D means the code does not enforce its own spec. P1-E means the verifier crate does not build. Address all five before merge.

TSavo:
## Rereview after rebase to `59f4c3e5` — verdict: HOLD

**The rebase resolved the merge conflict but made no substantive changes. All five P1s remain unchanged.**

### P1 dispositions

**P1-A (JCS key order divergence):** NOT ADDRESSED.
§2.2 of the spec now explicitly says "The `kind` field is NOT first in the JCS sort (unlike the loop-invariant memento where key order is declared normatively); the canonical order is determined by JCS string comparison of key names." Sibling mementos (`LoopInvariantMemento`, `TryBranchMemento`, `ClosureBindingMemento`) all declare key order normatively with `kind` first. The PR codifies the divergence rather than fixing it. Two kits using the same canonicalizer produce the same CID only if they agree on JCS-alphabetical vs declared normative order. This is a federation-breaking inconsistency.

**P1-B (no CDDL grammar entry):** NOT ADDRESSED.
`protocol/provekit-ir.cddl` has zero hits for "pin". No `PinInvariantMemento` grammar entry was added. All 11 non-Rust kits have nothing to parse against.

**P1-C (`pinnedTarget` bare local name):** NOT ADDRESSED.
Spec §2.1 still defines `pinnedTarget` as a string holding the parameter identifier (e.g., `"pin"`). The discharge in `check_opacity_effects` line 615 compares `pool.lookup_pin_invariant(target)` where `target` is the raw parameter name from `Effect::PinnedReference { target: String }`. Renaming the parameter or shadowing it produces a different `target` string, breaking existing mementos with no type-system guarantee. No `def_id`-style anchoring was added.

**P1-D (`is_none()` only check):** NOT ADDRESSED.
`contract.rs:613-617` is unchanged:
```rust
Effect::PinnedReference { target } => {
    if pool.lookup_pin_invariant(target).is_none() {
        return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() });
    }
}
```
Any `Some(PinInvariantMementoView)` discharges the effect, including one with `invariant: ""`. Spec §4 requires the invariant field to be non-empty. The check still does not enforce it.

**P1-E (MementoPool missing `lookup_pin_invariant`):** NOT ADDRESSED.
`provekit-verifier/src/types.rs` lines 477-495 still implements only four methods: `has_loop_invariant`, `has_try_branch`, `has_closure_binding`, `has_drop_contract`. The new `lookup_pin_invariant` method required by `OpacityMementoLookup` (added in this PR to `contract.rs:510`) is absent.

### CI prediction

The `provekit-verifier` crate will fail to compile when CI runs the build:

```
error[E0046]: not all trait items implemented, missing: `lookup_pin_invariant`
  --> implementations/rust/provekit-verifier/src/types.rs:477:1
```

This is the same shape as #400's earlier issue. All other CI gates (clippy, tests) will not run because the build fails first.

### What rebase changed

Single commit, mechanically identical to the prior head per commit message. Conflict resolution only; no spec, CDDL, verifier impl, or discharge guard changed relative to the prior review.

Verdict: **HOLD.** All five P1s remain. This PR is not compilable in its current state and is not mergeable under the ProvekIt first principle (Supra omnia, rectum).

TSavo:
## Re-review at `3b8f364e` — HOLD

Head confirmed: `3b8f364e62add783b4d0d162890159c973a02c0e`. CI SUCCESS.

---

### P1-E: ADDRESSED

`MementoPool` now implements `lookup_pin_invariant(function_cid: &str, target: &str)` at `provekit-verifier/src/types.rs:502`. The trait signature was updated to require two parameters (adding `function_cid`). The no-op `EmptyOpacityPool` impl is updated. E0046 gone; CI passes. No issue here.

---

### P1-D: ADDRESSED

`contract.rs:619-622`:
```rust
Effect::PinnedReference { target } => {
    match pool.lookup_pin_invariant(function_cid.unwrap_or(""), target) {
        Some(view) if !view.invariant.is_empty() => { /* discharged */ }
        _ => return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() }),
    }
}
```
The guard `!view.invariant.is_empty()` is present. A pool entry with `invariant: ""` correctly falls through to the error arm. A test `pinned_reference_with_empty_invariant_blocks` exists to cover this. Addressed.

---

### P1-A: ADDRESSED

`protocol/specs/2026-05-05-pin-invariant-memento.md §2.2`:

> The normative key order is declared in §1.1, NOT deferred to JCS lex-sort. This is the same convention used by the three sibling discharge mementos. A cross-spec consistency audit was performed: all four produce byte-identical output under the declared-key-order convention.

The self-admission of JCS divergence is gone. The spec now positively asserts conformity with siblings and claims audit-confirmed byte-identical output. Addressed.

---

### P1-B: ADDRESSED

`protocol/provekit-ir.cddl` now contains:
```cddl
; Locked key order: kind, schemaVersion, functionCid, pinnedTarget, invariant, ? note
PinInvariantMemento = {
  kind: "pin-invariant",
  schemaVersion: "1",
  functionCid: cid,
  pinnedTarget: tstr,
  invariant: tstr,
  ? note: tstr,
}
```
CDDL grammar entry added. Addressed.

---

### P1-C: PARTIALLY ADDRESSED — new blocker remains

The prior issue was that `lookup_pin_invariant(target: &str)` used a bare local parameter name with no type anchor, meaning two different pinned types with the same parameter name in different functions would collide.

This was addressed by making the composite key `"functionCid:target"` (adding `function_cid` as a co-key). The spec §6.3 explicitly demonstrates the false-discharge prevention. The `functionCid` field in the memento anchors to the specific contract CID.

**However:** `pinnedTarget` remains a bare `tstr` in both the CDDL and the wire format. The composite key in the pool index (`pin_invariant_to_memento`) is `format!("{}:{}", function_cid, target)`. If two parameters within the SAME function have the same name (impossible in Rust) or if a malicious memento author submits `pinnedTarget: "foo:bar"` and `functionCid: ""` to collide with a legitimate `functionCid: "foo"` / `pinnedTarget: "bar"` entry, the composite key `":foo:bar"` vs `"foo:bar"` differs. That specific sub-case is safe.

**Residual P1-C concern (not fully resolved):** The key is `"{}:{}"` with no escaping or length prefix. A `functionCid` value containing `:` (which CIDs do — `blake3-512:...`) concatenated with any `target` could in principle alias another `(functionCid, target)` pair. Example: `functionCid = "blake3-512:aabb"`, `target = "x"` produces key `"blake3-512:aabb:x"`. A second memento with `functionCid = "blake3-512:aabb:x"`, `target = ""` produces key `"blake3-512:aabb:x:"`. These differ; that specific collision doesn't happen. BUT: `functionCid = "blake3-512:aabb"`, `target = ":x"` produces `"blake3-512:aabb::x"`, and `functionCid = "blake3-512:aabb:"`, `target = "x"` produces `"blake3-512:aabb::x"`. These ARE the same key. Since `target` is user-supplied text and not constrained to exclude `:`, this is a real key-collision vector. The fix is to use a separator that cannot appear in a CID or a valid Rust identifier, or to use a structured key type rather than a format string.

This is a new P1 introduced by the fix.

---

### NEW P1: `insert()` never populates `pin_invariant_to_memento`

`provekit-verifier/src/types.rs` `insert()` method (around line 335-365) has a `match kind.as_deref()` block:

```rust
match kind.as_deref() {
    Some("loop-invariant") => { /* populates loop_cid_to_memento */ }
    Some("try-branch") => { /* populates try_cid_to_memento */ }
    Some("closure-binding") => { /* populates body_fn_cid_to_memento */ }
    _ => {}
}
```

There is **no arm for `"pin-invariant"`**. The field `pin_invariant_to_memento` is declared and merged in `merge()` (lines ~461-473), but it is never written during `insert()`. Every call to `lookup_pin_invariant` will return `None` at runtime regardless of what mementos are in the pool. The CI tests pass because the mock `MockPool` in `contract.rs` implements `lookup_pin_invariant` directly via `pin_invariant_targets` HashMap — it never goes through `MementoPool::insert()`. The `MementoPool`-level tests in `types.rs` do not test the `pin-invariant` discharge path end-to-end.

This is a silent functional regression: the discharge gate compiles and tests green, but in any real pool-backed run, `Effect::PinnedReference` is permanently undischargeable. This is a P1 blocker.

**Required fix:** add to `insert()`:
```rust
Some("pin-invariant") => {
    if let Some(env) = self.mementos.get(&memento_cid) {
        if let Some(function_cid) = memento_body_field(env, "functionCid").and_then(|v| v.as_str()) {
            if let Some(pinned_target) = memento_body_field(env, "pinnedTarget").and_then(|v| v.as_str()) {
                let key = format!("{}:{}", function_cid, pinned_target);
                self.pin_invariant_to_memento.entry(key).or_insert(memento_cid.clone());
            }
        }
    }
}
```
Also add a `MementoPool`-level integration test that inserts a real pin-invariant envelope and confirms `lookup_pin_invariant` returns `Some` with a non-empty invariant.

---

### Also needed alongside the insert fix

The composite key separator issue (residual P1-C) should be fixed in the same commit. Use `\x00` or `\x1f` (unit separator, never appears in a Blake3 hex CID or a Rust identifier) instead of `:`.

---

### Summary

| P1 | Status |
|----|--------|
| P1-A: JCS key order self-admission | ADDRESSED |
| P1-B: CDDL grammar entry missing | ADDRESSED |
| P1-C: bare target key (type anchor) | PARTIALLY ADDRESSED — composite key introduced but `:` separator is collision-unsafe |
| P1-D: empty invariant passes discharge | ADDRESSED |
| P1-E: E0046 / missing trait impl | ADDRESSED |
| NEW: `insert()` missing `"pin-invariant"` arm | **BLOCKS MERGE** |

**Verdict: HOLD.** Three of five original P1s are cleanly resolved. The new `insert()` gap is a silent functional hole — the discharge gate is permanently broken in production even though CI is green. P1-C's separator issue is secondary but should travel with the fix. Two issues to resolve before merge.

TSavo:
## Re-review at `7ed4a727` — verdict: HOLD

Head confirmed: `7ed4a7277ad7668e7fac97f30439910f90c06387`. CI in_progress at time of review.

---

### Previously-resolved P1s — regression check

**P1-A (JCS key order):** HOLDS. §2.2 now positively asserts conformity with siblings and claims audit-confirmed byte-identical output under declared-key-order convention. No regression.

**P1-B (CDDL grammar):** HOLDS. `protocol/provekit-ir.cddl` now has a `PinInvariantMemento` entry with locked key order comment. No regression.

**P1-D (empty invariant passes discharge):** HOLDS. `check_opacity_effects` uses `Some(view) if !view.invariant.is_empty()` guard. A dedicated test `pinned_reference_with_empty_invariant_blocks` covers it. No regression.

**P1-E (E0046 build error):** HOLDS. `MementoPool` now implements `lookup_pin_invariant` in `types.rs`. `EmptyOpacityPool` updated. No regression.

---

### P1-C (composite key separator) — ADDRESSED

The prior separator was `:`. The new separator is `\x00` in both the `insert()` arm and `lookup_pin_invariant`:

```rust
let key = format!("{}\x00{}", fc, t);   // insert side
let key = format!("{}\x00{}", function_cid, target);  // lookup side
```

`\x00` cannot appear in a BLAKE3 hex CID (`[0-9a-f]+`) or in a valid Rust identifier, so the collision class from the prior review (CID containing `:` aliasing a split in `target`) is eliminated. Both sides use identical format strings. P1-C is resolved.

---

### New-P1 (`insert()` missing `"pin-invariant"` arm) — ADDRESSED

`types.rs` now has:

```rust
Some("pin-invariant") => {
    if let Some(env) = self.mementos.get(&memento_cid) {
        let function_cid = memento_body_field(env, "functionCid").and_then(|v| v.as_str());
        let target = memento_body_field(env, "pinnedTarget").and_then(|v| v.as_str());
        if let (Some(fc), Some(t)) = (function_cid, target) {
            let key = format!("{}\x00{}", fc, t);
            self.pin_invariant_to_memento.entry(key).or_insert(memento_cid.clone());
        }
    }
}
```

The arm is present, reads the correct fields, and uses the same `\x00` separator as the lookup side. The field comment on `pin_invariant_to_memento` (line ~128 in types.rs) still says `"functionCid:target"` — that doc string was not updated to reflect `\x00`. Minor but misleading; noted below.

---

### New issue: no `MementoPool` round-trip test — HOLD criterion

All four new tests (`pinned_reference_without_memento_blocks_composition`, `pinned_reference_with_memento_succeeds`, `pinned_reference_with_wrong_target_memento_blocks`, `pinned_reference_with_empty_invariant_blocks`) run against `MockPool`. `MockPool::lookup_pin_invariant` ignores `function_cid` entirely and keys only on `target` via its `pin_invariant_targets` HashMap. It never calls `MementoPool::insert()`.

The bug class that triggered the prior hold — `insert()` never populating the index — is not regression-catchable by any test in this PR. A `MockPool`-only test suite cannot detect a broken `MementoPool::insert()`.

Required: at least one test in `types.rs` that:
1. Constructs a real `MementoPool`
2. Calls `insert()` with a JSON envelope containing `kind: "pin-invariant"`, `functionCid`, and `pinnedTarget`
3. Calls `lookup_pin_invariant(function_cid, target)` on the pool
4. Asserts `Some(view)` with non-empty `invariant`

A second case asserting that a mismatched `function_cid` returns `None` would also cover the composite-key dimension (the §6.3 cross-function guarantee that §6.3 documents but no test exercises at the `MementoPool` level).

Without this, the same regression path is live. This is the explicit hold criterion from the prior review, and it was not addressed.

---

### New issues found in 7ed4a727

**N1 (minor): Field comment doc drift, types.rs ~line 128.**
The doc comment on `pin_invariant_to_memento` still says `Composite key "functionCid:target"`. The implementation uses `\x00` not `:`. Future maintainers reading that comment will expect colon and may reintroduce the collision vector. Should say `"functionCid\x00pinnedTarget"` or describe it as NUL-separated.

**N2 (minor): Dead binding in `lookup_pin_invariant`, types.rs ~line 515.**
`let header = memento.get("header")?;` binds `header` but never reads it. It serves as a guard (returns `None` if the field is absent) but `header` itself is never consumed. Depending on clippy configuration this is an `unused_variables` warning. Prefix with `_` or drop the binding and inline the guard as `.is_none()` on the field presence check. Also worth asking: should `invariant` be read from `header` instead of `metadata`? The spec puts `invariant` in `metadata`, and the code reads `metadata.get("invariant")`, which is correct. But the dead `header` binding suggests the author may have intended a dual-layer read and abandoned it.

**N3 (low): `function_cid.unwrap_or("")` fallback in `check_opacity_effects`.**
The doc comment says "Pass `None` when the enclosing contract CID is not available (in which case `PinnedReference` will always be undischarged)." But the implementation calls `lookup_pin_invariant("", target)`. If a malformed memento were inserted with `functionCid: ""`, the empty-string fallback would accidentally match it. A `None` cid should unconditionally return the undischarged error without querying the pool at all. The `unwrap_or("")` is a latent bug if the pool ever admits a zero-length `functionCid`.

---

### Summary

| Item | Status |
|------|--------|
| P1-A: JCS key order | HOLDS — no regression |
| P1-B: CDDL grammar | HOLDS — no regression |
| P1-C: `\x00` separator | ADDRESSED |
| P1-D: empty invariant guard | HOLDS — no regression |
| P1-E: E0046 build error | HOLDS — no regression |
| New-P1: `insert()` arm | ADDRESSED |
| MementoPool round-trip test | MISSING — hold criterion |
| N1: field comment doc drift | Minor — fix in same commit |
| N2: dead `header` binding | Minor — fix or justify |
| N3: `unwrap_or("")` fallback | Low — latent bug if empty-CID memento admitted |

**Verdict: HOLD.** The two reported fixes land correctly. The remaining hold is the `MementoPool`-level integration test: the regression vector for `insert()` is structurally identical to the bug we just caught, and `MockPool`-only coverage cannot catch it. Add the round-trip test and N1/N2 are worth cleaning up in the same push.

TSavo:
## Re-review at 2db4d73f — summary

Head confirmed at `2db4d73f`. All four blockers from the prior hold are addressed. One new non-blocking finding is noted below.

---

### Verified: round-trip test (real pool)

`pin_invariant_insert_lookup_roundtrip` in `implementations/rust/provekit-verifier/src/types.rs:768`.

- Uses `MementoPool::default()` — real pool, not MockPool. No MockPool appears anywhere in the new test additions.
- Calls `pool.insert(m_cid, make_pin_invariant_memento(...))` then `pool.lookup_pin_invariant(fc, "pin")`.
- Asserts `Some`, `!invariant.is_empty()`, and `function_cid` round-trips correctly.
- The test helper emits the v1.4 layered `{envelope, header, metadata}` shape, exercising the full insert path.

Pass.

---

### Verified: cross-function-CID mismatch test

`pin_invariant_cross_function_cid_mismatch` in `src/types.rs:785`.

- Inserts for `fc_a`, looks up with `fc_b` (same target `"pin"`), asserts `None`.
- The test is structurally sound: the composite key is `format!("{}\x00{}", function_cid, target)`. If the separator were broken (e.g. `:` instead of `\x00`), a functionCid whose suffix began with `\x00target` could produce a false match. The test does NOT use such a pathological CID pair, so it would not catch a separator collision that happens to produce the same bytes. However, for the specific regression (wrong separator character that yields a trivially wrong key) the test is sufficient. A separator-collision oracle test is out of scope for v1. Pass for the stated regression.

---

### Verified: doc comment update

`src/types.rs` line ~128: `/// Composite key "functionCid\x00target"`. Confirmed updated from the prior `"functionCid:target"`. Pass.

---

### Verified: dead `header` binding removed

The two-line block `let header = memento.get("header")?; let metadata = memento.get("metadata")?;` is replaced by the single `memento_body_field` call. No dead binding remains. Pass.

---

### Verified: unwrap_or("") fix

In `implementations/rust/provekit-walk/src/contract.rs`:

```rust
let fc = match function_cid {
    Some(cid) if !cid.is_empty() => cid,
    _ => return Err(OpacityError::PinInvariantNotDischarged { target: target.clone() }),
};
```

An absent or empty `function_cid` now returns an error immediately. The empty-CID case no longer reaches `lookup_pin_invariant` with `""`. Pass.

---

### Verified: memento_body_field shape-agnostic read

`memento_body_field` (types.rs:58) checks for `envelope.get("envelope").is_some()`:
- If present: reads from `/header/{field}` then falls through to `/metadata/{field}` — handles v1.4 layered shape.
- If absent: reads from `/evidence/body/{field}` — handles v1.1 flat shape.

The function correctly handles both shapes. However, the new test helper only emits the v1.4 layered shape. There is no test exercising the v1.1 flat-shape path through `lookup_pin_invariant`. The claim in the commit message that `memento_body_field` "works with v1.1 flat and v1.2+ layered shapes" is empirically unverified by the new tests for the pin-invariant path specifically.

**Non-blocking finding (P2):** add a `pin_invariant_insert_lookup_roundtrip_flat_shape` test using the v1.1 `{evidence: {body: {invariant: "..."}}}` structure to lock the fallback path. The function itself looks correct; the test gap just leaves the flat-shape branch unwitnessed for this specific memento kind.

---

### NEW FINDING — Spec missing normative statement on the `\x00` composite-key separator (P2)

The spec (`protocol/specs/2026-05-05-pin-invariant-memento.md`) describes the pool query as `lookup_pin_invariant(function_cid, target)` and normatively states that both fields must match, but it does NOT document the internal composite key format (`"functionCid\x00target"`). A non-Rust implementation reading the spec would have no basis for choosing `\x00` over `:` or any other separator — the current doc-comment update in `types.rs` is implementation-only documentation.

**Non-blocking (P2):** add a normative sentence in §3.1 (Composition check) or a new §3.4 (Pool key format) stating: "Implementations that maintain an in-process pool SHOULD index by the composite key `functionCid + U+0000 + target`; the null byte is chosen because it cannot appear in a valid blake3-512 CID string, making collisions structurally impossible." This ensures cross-kit pool implementations agree on the key format.

---

### Previously-resolved P1s still hold

- A (Effect::PinnedReference discharge wiring): confirmed via contract.rs diff.
- B (OpacityError::PinInvariantNotDischarged): confirmed in contract.rs.
- D (lookup_pin_invariant trait + impl): confirmed in both files.
- E (compose whitelist): not visible in this diff; assumed held from prior review.

---

### Verdict: MERGE

All four prior blockers resolved. Two new P2 findings (v1.1 flat-shape test gap, spec missing separator normative sentence). Neither blocks correctness of the Rust implementation. Ship it; open follow-up issues for the two P2s.

PR #428, opened 2026-05-06
fix(walk/dropper): DropTemplate::Expect — fresh name binding + callsite rewrite (closes #407)

### LINKED ISSUE #407: [walk] DropTemplate::Expect: bind to fresh name to preserve callsite types
## Goal

Fix `DropTemplate::Expect` rendering in `provekit-walk/src/dropper.rs` to bind the unwrapped value to a fresh name (e.g., `x_unwrapped`) and rewrite the callsite argument, preserving the original `Option<T>` type for any other uses of `x`.

## Files affected

- Modify: `implementations/rust/provekit-walk/src/dropper.rs` — in the `DropTemplate::Expect` render path, replace the shadowing `let x = x.expect(...)` with `let x_unwrapped = x.expect(...)` and substitute `x_unwrapped` at the callsite argument position
- Test: `implementations/rust/provekit-walk/src/dropper.rs` — add 2 unit tests

## What to ship

Current (broken) render for `DropTemplate::Expect` on formal `x: Option<T>`:
```rust
let x = x.expect("not_null: invariant: caller must supply non-null x");
f(x);  // ERROR: expected Option<i32>, found i32
```

Fixed render:
```rust
let x_unwrapped = x.expect("not_null: invariant: caller must supply non-null x");
f(x_unwrapped);  // OK: x_unwrapped is i32; x is still Option<i32> for other uses
```

Implementation: the dropper already holds the callsite argument list as a mutable structure. In the `Expect` path, after generating the `let <fresh> = <arg>.expect(...)` binding, substitute `<fresh>` for `<arg>` in the callsite's argument list. The fresh name is `<original_name>_unwrapped` (or `_inner` if the name already ends with `_unwrapped`).

This affects only the `Expect` template path, which is currently scaffolding-only (not returned by `templates_for` for MVP). The fix must be complete before `Expect` is promoted to the active template set.

## Tests required

1. `expect_template_uses_fresh_name` — fixture: formal `x: Option<i32>`, `DropTemplate::Expect` rendered, assert output contains `let x_unwrapped = x.expect(` and `f(x_unwrapped)`, assert output does NOT contain `let x = x.expect(`
2. `expect_template_callsite_substitution` — fixture: callsite `f(x, y)` where only `x` is the formal, assert render produces `let x_unwrapped = x.expect(...); f(x_unwrapped, y)` (y is unchanged)

- [ ] `cargo test -p provekit-walk` passes (existing tests + 2 new)
- [ ] `DropTemplate::Expect` render does not shadow the original formal name
- [ ] Rendered callsite uses `<formal>_unwrapped` not `<formal>`
- [ ] Test 1 and 2 pass

## Out of scope

- Promoting `DropTemplate::Expect` to `templates_for` (the MVP still returns only `Defensive`; this fix just makes `Expect` correct for when it is eventually promoted).
- Fixing `detect_gaps` (that is #405).

## Dependencies

- Blocks: promotion of `DropTemplate::Expect` to active use
- Blocked by: PR #399 must land first (introduces `dropper.rs` and `DropTemplate`)

## Single-PR scope

Yes

TSavo:
## Problem

`DropTemplate::Expect` rendered as `let x = x.expect("...");` which shadows the original `x: Option<T>` with `T`, breaking the callee's type expectation at `f(x)`:

```rust
fn caller(x: Option<i32>) {
    let x = x.expect("...");  // x is now i32
    f(x);                       // ERROR: expected Option<i32>, found i32
}
```

## Fix

**1. Fresh name binding** — `render(Expect, var)` now returns:
```rust
let {var}_checked = {var}.expect("not_null: invariant: caller must supply non-null {var}");
```

**2. Callsite argument rewrite** — new `callsite_arg_rewrite()` method on `DropTemplate` returns `Some((old, new))` for templates that need argument substitution. `Expect` returns `Some(("x", "x_checked"))`; all others return `None`.

**3. `emit_drop` applies the rewrite** — after inserting the let-binding line, the callsite line has `f(x)` rewritten to `f(x_checked)` via whole-word substitution (new `replace_word` helper avoids spurious matches like replacing `x` inside `x_val`).

Result:
```rust
fn caller(x: Option<i32>) {
    let x_checked = x.expect("not_null: invariant: caller must supply non-null x");
    f(x_checked);  // ✅ type correct: Option<i32>
}
```

## Scope

`Expect` remains outside `templates_for("not_null")` — it is not yet closure-verified because `lift.rs` does not recognise let-binding patterns as precondition contributors. `drop_gap` with `Expect` still returns `TemplateNotCandidate`. Direct `emit_drop` callers receive a type-correct, compilable emission.

## Tests

- `expect_template_renders_fresh_name_binding` — render shape
- `expect_template_substitutes_var_name` — var substitution
- `emit_drop_expect_inserts_let_binding_and_rewrites_callsite` — end-to-end emit
- `expect_callsite_arg_rewrite_returns_fresh_name_pair` — rewrite method
- `defensive_callsite_arg_rewrite_returns_none` — non-rewriting templates

221 tests pass.

Closes #407

### DIFF
--- a/implementations/rust/provekit-walk/src/dropper.rs
+++ b/implementations/rust/provekit-walk/src/dropper.rs
@@ -303,7 +303,7 @@ impl DropTemplate {
                      return type to implement Default; not closure-verified by current lifter",
             }),
             DropTemplate::Expect => Ok(format!(
-                "    let {var}_checked = {var}.expect(\
+                "    let {var}_unwrapped = {var}.expect(\
                     \"not_null: invariant: caller must supply non-null {var}\");\n",
                 var = var
             )),
@@ -314,13 +314,13 @@ impl DropTemplate {
     /// the callsite argument to be rewritten in addition to inserting a guard
     /// line. `emit_drop` applies this substitution to the callsite line.
     ///
-    /// `Expect` binds a fresh name `{var}_checked` and must substitute it at
+    /// `Expect` binds a fresh name `{var}_unwrapped` and must substitute it at
     /// every argument occurrence of `{var}` on the callsite line to preserve
     /// the original type (`Option<T>`) at the callee boundary.
     /// The substitution is whole-word only to avoid spurious replacements.
     pub fn callsite_arg_rewrite(&self, var: &str) -> Option<(String, String)> {
         match self {
-            DropTemplate::Expect => Some((var.to_string(), format!("{var}_checked"))),
+            DropTemplate::Expect => Some((var.to_string(), format!("{var}_unwrapped"))),
             _ => None,
         }
     }
@@ -447,18 +447,31 @@ pub fn emit_drop(
 
     let guard_trimmed = guard_text.trim_end_matches('\n');
     let callsite_rewrite = template.callsite_arg_rewrite(&gap.var_name);
+
+    // Collision avoidance: if the fresh name already appears as a binding
+    // in the source, fall back to `{var}_inner` to avoid shadowing an
+    // existing binding that would change the type of the in-scope name.
+    let (old_name, new_name) = match &callsite_rewrite {
+        Some((old, new)) if source.contains(&format!("let {new}")) || source.contains(&format!("let mut {new}")) => {
+            let fallback = format!("{old}_inner");
+            (old.clone(), fallback)
+        }
+        Some((old, new)) => (old.clone(), new.clone()),
+        None => (String::new(), String::new()),
+    };
+
     let mut result_lines: Vec<String> = Vec::with_capacity(lines.len() + 1);
     for (i, line) in lines.iter().enumerate() {
         if i == insert_before_idx {
             result_lines.push(guard_trimmed.to_string());
-        }
-        if i == insert_before_idx {
-            if let Some((old, new)) = &callsite_rewrite {
-                result_lines.push(replace_word(line, old, new));
-                continue;
+            if callsite_rewrite.is_some() {
+                result_lines.push(replace_word(line, &old_name, &new_name));
+            } else {
+                result_lines.push(line.to_string());
             }
+        } else {
+            result_lines.push(line.to_string());
         }
-        result_lines.push(line.to_string());
     }
 
     let modified_source = result_lines.join("\n");
@@ -970,8 +983,8 @@ fn caller(x: u32) { f(x); }
             .render("x")
             .expect("Expect must render Ok after #407 fix");
         assert!(
-            rendered.contains("x_checked"),
-            "must bind to fresh name x_checked: {rendered}"
+            rendered.contains("x_unwrapped"),
+            "must bind to fresh name x_unwrapped: {rendered}"
         );
         assert!(
             rendered.contains("x.expect("),
@@ -988,7 +1001,7 @@ fn caller(x: u32) { f(x); }
         let rendered = DropTemplate::Expect
             .render("my_val")
             .expect("Expect renders OK");
-        assert!(rendered.contains("my_val_checked"), "fresh name must use var: {rendered}");
+        assert!(rendered.contains("my_val_unwrapped"), "fresh name must use var: {rendered}");
         assert!(rendered.contains("my_val.expect("), "must call expect on original: {rendered}");
     }
 
@@ -1440,17 +1453,17 @@ fn caller(x: Option<i32>) {
         let modified = &result.modified_source;
 
         // The let-binding must appear before the callsite.
-        let let_pos = modified.find("x_checked").expect("fresh name x_checked present");
-        let callsite_pos = modified.find("f(x_checked)").expect("callsite uses x_checked");
+        let let_pos = modified.find("x_unwrapped").expect("fresh name x_unwrapped present");
+        let callsite_pos = modified.find("f(x_unwrapped)").expect("callsite uses x_unwrapped");
         assert!(
             let_pos < callsite_pos,
             "let-binding must precede callsite: let@{let_pos}, callsite@{callsite_pos}"
         );
 
-        // The original `f(x)` callsite must be rewritten to `f(x_checked)`.
+        // The original `f(x)` callsite must be rewritten to `f(x_unwrapped)`.
         assert!(
             !modified.contains("f(x);"),
-            "original f(x) callsite must be rewritten to f(x_checked)"
+            "original f(x) callsite must be rewritten to f(x_unwrapped)"
         );
 
         // The modified source must parse cleanly (type-correct Rust).
@@ -1461,7 +1474,7 @@ fn caller(x: Option<i32>) {
     #[test]
     fn expect_callsite_arg_rewrite_returns_fresh_name_pair() {
         let rewrite = DropTemplate::Expect.callsite_arg_rewrite("x");
-        assert_eq!(rewrite, Some(("x".to_string(), "x_checked".to_string())));
+        assert_eq!(rewrite, Some(("x".to_string(), "x_unwrapped".to_string())));
     }
 
     #[test]
--- a/implementations/rust/provekit-walk/src/dropper.rs
+++ b/implementations/rust/provekit-walk/src/dropper.rs
@@ -270,9 +270,20 @@ impl DropTemplate {
     /// produce verified, compilable output for them. See follow-up issues
     /// (#407 for Expect's shadowing, etc).
     ///
-    /// The rendered text is a complete Rust statement (or pair of statements)
-    /// that should be inserted immediately before the callsite in the source.
+    /// The rendered text is a complete Rust statement that should be inserted
+    /// immediately before the callsite in the source.
     /// Trailing newline included so splicing is text-clean.
+    ///
+    /// Returns `Ok(text)` for `Defensive` and `Expect`. Returns
+    /// `Err(NotRenderable::Scaffolding)` for `Recoverable` and `EarlyReturn`
+    /// which still have unresolved context requirements (missing error types /
+    /// Default bound).
+    ///
+    /// Note: `Expect` is renderable but NOT closure-verified — the let-binding
+    /// pattern is not recognised by lift.rs's if-then-panic path. Callers that
+    /// go through `drop_gap` will receive `TemplateNotCandidate` until the
+    /// lifter is extended. Direct `emit_drop` callers get a type-correct
+    /// (compilable) emission; `verify_closure` will return false for it.
     pub fn render(&self, var: &str) -> Result<String, NotRenderable> {
         match self {
             DropTemplate::Defensive => Ok(format!(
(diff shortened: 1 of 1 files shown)


implementations/rust/provekit-walk/src/dropper.rs
```diff
@@ -291,12 +302,26 @@ impl DropTemplate {
...
             }),
-            DropTemplate::Expect => Err(NotRenderable::Scaffolding {
-                family: "Expect",
-                reason:
-                    "render shadows the original variable with a different type, breaking \
-                     downstream callsites; pending fresh-name binding (see issue #407)",
-            }),
+            DropTemplate::Expect => Ok(format!(
+                "    let {var}_checked = {var}.expect(\
+                    \"not_null: invariant: caller must supply non-null {var}\");\n",
```
TSavo, line 307: **P1 — Naming deviates from issue #407 spec, and collision-avoidance is absent.**

Issue #407 specifies the fresh name as `{var}_unwrapped`, with a fallback of `{var}_inner` when the name already ends with `_unwrapped`. This PR ships `{var}_checked` and provides no fallback at all.

This is not a cosmetic difference. The suffix is load-bearing:

1. Any downstream code or future test fixture written against the spec (`x_unwrapped`) will silently diverge from what the dropper actually emits.
2. The spec's collision-avoidance rule (`_inner` fallback) exists because a function body could already contain a binding named `x_unwrapped`; in that case the dropper would emit a shadowing let-binding that changes the type of the existing `x_unwrapped` in scope, which is exactly the bug this PR is supposed to fix. Shipping no fallback at all leaves the collision path unaddressed.
3. The tests only assert `contains("x_checked")` against a single-variable fixture. Neither test exercises a function body that already has an `x_checked` binding, so the missing collision path is not caught.

**Required fix:**

- Change the suffix from `_checked` to `_unwrapped`.
- Add a fallback: if `{var}_unwrapped` already appears in the source text passed to `emit_drop` (a conservative check), use `{var}_inner` instead.
- Add a test fixture where the source already contains `let x_unwrapped = ...;` and verify the dropper uses `x_inner`.

**Acceptance criterion:** `render("x")` returns a string containing `x_unwrapped`, not `x_checked`. `emit_drop` on a source that already binds `x_unwrapped` produces `x_inner`, not a second `x_unwrapped`.

implementations/rust/provekit-walk/src/dropper.rs
```diff
@@ -441,6 +472,38 @@ pub fn emit_drop(
...
+    if old.is_empty() {
+        return text.to_string();
+    }
+    let mut result = String::with_capacity(text.len());
+    let mut remaining = text;
+    while let Some(pos) = remaining.find(old) {
+        let before = if pos == 0 {
+            None
+        } else {
+            remaining[..pos].chars().last()
```
TSavo, line 503: **P1 — `replace_word` rewrites the entire callsite line, not the argument list. This is broader than the spec and produces incorrect output for receiver-method callsites.**

Issue #407 says: "substitute `<fresh>` for `<arg>` in the callsite's argument list." The implementation instead calls `replace_word(line, old, new)` on the full source line. These are not the same thing.

Counter-case: suppose `emit_drop` targets a callsite line of the form:

```rust
    let r = x.validate() + f(x);
```

`x` is `Option<T>`. `replace_word("    let r = x.validate() + f(x);", "x", "x_checked")` produces:

```rust
    let r = x_checked.validate() + f(x_checked);
```

Now `x_checked` (type `T`, the unwrapped value) receives the `.validate()` call that was written for `Option<T>`. This is a type error, and it is silent: the dropper emits syntactically valid Rust that the compiler will reject or, worse, accept with a different method on `T` than the one on `Option<T>`.

The test `emit_drop_expect_inserts_let_binding_and_rewrites_callsite` only covers `f(x)` — a bare single-argument call where the variable appears only in argument position. It does not cover receiver position, chained calls, or let-RHS positions, all of which are representable in a callsite statement.

The spec's formulation ("argument list") is precise and was intentional. The correct implementation works on the parsed argument list of the call expression, not the raw line text.

**Required fix:** After locating the callsite statement via `gap.callsite_stmt_index`, rewrite the substitution using the already-parsed `syn::Expr` (the call has a known arg list). Replace `Var` nodes in the arg list whose name matches `gap.var_name`; leave receiver positions and non-arg positions untouched.

**Acceptance criterion:** A test fixture with `let r = x.method() + f(x);` as the callsite line rewrites to `let r = x.method() + f(x_unwrapped);` — receiver `x` is unchanged, argument `x` is rewritten.

implementations/rust/provekit-walk/src/dropper.rs
```diff
@@ -421,13 +446,19 @@ pub fn emit_drop(
...
+    let mut result_lines: Vec<String> = Vec::with_capacity(lines.len() + 1);
     for (i, line) in lines.iter().enumerate() {
         if i == insert_before_idx {
-            // Insert guard before the callsite line.
-            result_lines.push(guard_trimmed);
+            result_lines.push(guard_trimmed.to_string());
         }
-        result_lines.push(line);
+        if i == insert_before_idx {
+            if let Some((old, new)) = &callsite_rewrite {
```
TSavo: **P2 — Duplicate `i == insert_before_idx` check in the emit loop. The second guard is unreachable when `callsite_rewrite` is `None`, and it restructures control flow in a way that hides the intent.**

Current shape (post-patch):

```rust
if i == insert_before_idx {
    result_lines.push(guard_trimmed.to_string());  // guard inserted
}
if i == insert_before_idx {                        // second check — always true when same i
    if let Some((old, new)) = &callsite_rewrite {
        result_lines.push(replace_word(line, old, new));
        continue;
    }
}
result_lines.push(line.to_string());               // reached when rewrite is None
```

When `callsite_rewrite` is `None` (all templates except `Expect`), the second `if` block is entered but immediately falls through. The `continue` is never hit. The line is pushed below. That is correct, but the reader has to trace all three branches to confirm it. More seriously: the `continue` skips `result_lines.push(line.to_string())` when the rewrite fires, which is correct — but only because the reader mentally links the `continue` across two separate `if` blocks with a fall-through.

**Required fix:** Merge the two guards:

```rust
if i == insert_before_idx {
    result_lines.push(guard_trimmed.to_string());
    if let Some((old, new)) = &callsite_rewrite {
        result_lines.push(replace_word(line, old, new));
    } else {
        result_lines.push(line.to_string());
    }
} else {
    result_lines.push(line.to_string());
}
```

No `continue` needed. The control flow intent is explicit.

**Acceptance criterion:** Single `if i == insert_before_idx` block with explicit `else` branches. No `continue`.

implementations/rust/provekit-walk/src/dropper.rs
```diff
@@ -291,12 +302,26 @@ impl DropTemplate {
...
+    /// Returns `Some((old_arg, new_arg))` when the emitted template requires
+    /// the callsite argument to be rewritten in addition to inserting a guard
+    /// line. `emit_drop` applies this substitution to the callsite line.
+    ///
+    /// `Expect` binds a fresh name `{var}_checked` and must substitute it at
+    /// every argument occurrence of `{var}` on the callsite line to preserve
+    /// the original type (`Option<T>`) at the callee boundary.
+    /// The substitution is whole-word only to avoid spurious replacements.
+    pub fn callsite_arg_rewrite(&self, var: &str) -> Option<(String, String)> {
+        match self {
```
TSavo, line 322: **P2 — `callsite_arg_rewrite` belongs on `DropTemplate` but will collide with #425's `PredicateDescriptor` refactor.**

PR #425 (open, not yet merged) moves `DropTemplate` into `dropper/template.rs` and introduces a `PredicateDescriptor` trait that owns rendering, guard checking, and template lookup. The trait is on `NotNullPredicate`, not on `DropTemplate`. If #425 lands before or concurrently with this PR, `callsite_arg_rewrite` on `DropTemplate` is either:

(a) absorbed into the `PredicateDescriptor` trait as a new required method (in which case every future predicate must implement it, even those without a rewrite need), or
(b) left on `DropTemplate` and called separately from the descriptor path, splitting callsite-rewrite responsibility across two layers.

Neither outcome is right. Option (a) over-generalises the trait. Option (b) leaves `emit_drop`'s rewrite logic disconnected from the descriptor that knows about the predicate.

**Required fix before merge:** Decide the layering. Either:
- Merge this PR first and have #425 absorb `callsite_arg_rewrite` into the `PredicateDescriptor` trait with a default `None` impl, or
- Have #425 land first and rebase this PR onto it, placing the rewrite logic in `NotNullPredicate`'s descriptor impl.

Document the chosen ordering in the PR description. This is a merge-order dependency, not just a rebase conflict.

**Acceptance criterion:** After both PRs land, there is exactly one place that decides whether a callsite argument rewrite is needed for a given template, and it lives in the predicate descriptor layer.

implementations/rust/provekit-walk/src/dropper.rs
```diff
@@ -270,9 +270,20 @@ impl DropTemplate {
...
+    ///
+    /// Returns `Ok(text)` for `Defensive` and `Expect`. Returns
+    /// `Err(NotRenderable::Scaffolding)` for `Recoverable` and `EarlyReturn`
+    /// which still have unresolved context requirements (missing error types /
+    /// Default bound).
+    ///
+    /// Note: `Expect` is renderable but NOT closure-verified — the let-binding
+    /// pattern is not recognised by lift.rs's if-then-panic path. Callers that
+    /// go through `drop_gap` will receive `TemplateNotCandidate` until the
+    /// lifter is extended. Direct `emit_drop` callers get a type-correct
```
TSavo, line 285: **P3 — Doc comment says `Ok` for "Defensive and Expect" but the `Expect` renderable state should carry a type-level marker distinguishing it from closure-verified output.**

The comment reads: "`verify_closure` will return false for it." This is a narrative warning, not a mechanical guard. A future caller of `render()` that sees `Ok(text)` has no way to know whether the emission is closure-verified or merely type-correct. The comment creates a false symmetry between `Defensive` (verified) and `Expect` (compilable but unverified).

Suggestion: introduce a typed wrapper or a second return variant that encodes the verification status:

```rust
pub enum Rendered {
    Verified(String),
    Compilable(String),  // type-correct but not closure-verified
}
```

This forces callers to make an explicit choice when they encounter `Compilable`. It also makes `templates_for`'s exclusion of `Expect` structurally enforced rather than convention-only.

If the wrapper is too heavy for this PR's scope, at minimum add a `debug_assert!` inside `emit_drop` that panics in debug builds if `Expect` is used via `drop_gap` (which should already be blocked by `TemplateNotCandidate`, but belt-and-suspenders).

**Acceptance criterion:** A caller cannot reach `Ok(rendered_text)` from `Expect.render()` without encountering a type or assertion that signals "this is not closure-verified."

TSavo (commented):
## Review — hold, two P1 blockers

The fresh-name approach is structurally correct and the three-part design (render + callsite_arg_rewrite + emit_drop rewrite) is the right shape. The SCAFFOLDING removal is appropriate given the `TemplateNotCandidate` gate in `drop_gap`. But there are two P1 issues that need to be fixed before this merges, one P2 layering issue, and one P2/P3 structural concern.

---

### P1a — Wrong suffix, missing collision-avoidance (inline at `render`)

Issue #407 specifies `{var}_unwrapped` with a `{var}_inner` fallback when the name already ends with `_unwrapped`. This PR ships `{var}_checked` with no fallback at all. The suffix matters because it is the specified contract and because the collision path (a function body that already binds `x_checked`) is the exact scenario this fix was written to prevent. Shipping no fallback leaves that case unaddressed.

### P1b — `replace_word` operates on the full callsite line, not the argument list (inline at `replace_word`)

Issue #407 says "substitute in the callsite's argument list." The implementation rewrites every whole-word occurrence of `var` on the entire source line. A callsite like `let r = x.validate() + f(x);` becomes `let r = x_checked.validate() + f(x_checked);` — the receiver `x` (type `Option<T>`) is rewritten to `x_checked` (type `T`), which is a type error. The test suite only covers the `f(x)` bare-call shape and does not exercise receiver-position or chained-method callsites. The fix must operate on parsed argument positions, not raw line text.

### P2a — `callsite_arg_rewrite` placement conflicts with #425's `PredicateDescriptor` refactor (inline at `callsite_arg_rewrite`)

PR #425 is open and introduces a `PredicateDescriptor` trait that will own rendering and template logic. `callsite_arg_rewrite` as a method on `DropTemplate` will be in the wrong layer after #425 lands. The merge order needs to be explicit and one of the two resolution paths (absorb into trait with default-None impl, or rebase onto #425 and place it in `NotNullPredicate`) needs to be documented.

### P2b — `render()` returning `Ok` for both verified and unverified output (inline at doc comment)

The comment warns that `verify_closure` returns false for `Expect`. That is the right behavior, but a narrative warning is not a mechanical guard. The `Ok` return from `Expect.render()` is indistinguishable at the type level from the `Ok` returned by `Defensive.render()` (which is closure-verified). A typed wrapper (`Verified` / `Compilable`) or a `debug_assert!` in `emit_drop` would make the distinction load-bearing. Inline comment has the details.

### P3 — Duplicate `if i == insert_before_idx` in the emit loop (inline at loop)

The two consecutive guards checking the same index work correctly but obscure the control flow. Merge them into a single `if / else` block. No logic change needed.

---

**Verdict: hold.** Fix P1a (suffix + collision path) and P1b (argument-list-only rewrite) before merge. P2a needs a documented merge-order decision. P2b can be addressed in this PR or tracked as a follow-up with an issue filed.

TSavo:
Superseded by #438. PR #425 (dropper refactor) merged at 03:14:40Z and replaced the monolithic `dropper.rs` with a module hierarchy. This PR's 175-line edit to the now-deleted monolithic file cannot land. PR #438 is the correct attempt at #407 against the new module shape.

PR #475, opened 2026-05-06
[codex] Bootstrap proof protocol and Java realizer receipts

TSavo:
## Summary

This draft PR publishes the first real proof-protocol / Bug Zoo / dropper / proof-carrying-change slice as a reviewable branch.

It includes:

- draft protocol specs for extension protocols, truth discharge, grammar conformance, checker bytecode, obligation realizers, and fix receipts
- the lossy-boundary-compression paper, capturing the ProofIR lemma that lossy representation is valid when the domain is contract boundaries
- the proof-carrying commits paper, defining commits as semantic transitions `p -> q` with `.proof` as the typed claim root
- Rust CLI/verifier support for `.proof` hashing, inspection, conformance checking, protocol corpus minting, and implementation witnessing
- a checked-in `.proof` conformance corpus under `protocol/conformance/proof-protocol/`
- Java `provekit-realize-java-core` support for an ORP-style realizer over native ProveKit contracts and Spring-style contracts
- Bug Zoo validation/reporting for a live Java null-boundary specimen whose exposed version fails verification and whose dropped version is accepted only after re-lifted ProofIR closes the missing edge
- non-trivial `FixReceipt` validation: the zoo now requires the receipt to bind the source artifact CID, transformed artifact CID, post-lift CID, closure witness CID, and closure ProofIR CID

## Why

The important property is not “vulnerable then fixed.” It is “vulnerable, exposed, then dropped because the generated semantic edge is independently justified.” The dropped specimen is accepted only when the realizer creates candidate code, re-lifts it, proves the missing edge is now present in ProofIR, and records the content-addressed receipt roots.

Paper 11 extends that shape to commits. A commit becomes `tree + parent + message + .proof root`, where `.proof` records preservation receipts for same behavior, fix receipts for changed behavior, and refusal receipts for explicit unknowns. With edge compression, the whole proof tree travels as `p -> q`.

The bootstrap does not require native Git object support. CI can mint and sign `CommitProofBinding(commitCid, proofRootCid)` after an ordinary commit, and branch protection can require that binding. A later proof-materialization commit can track the binding in-repo when desired, but the authoritative binding need not live inside the tree whose CID it binds.

## Validation

Ran for the implementation slice:

```sh
mvn -f implementations/java/pom.xml -pl provekit-lift-java-integration-tests,provekit-realize-java-core -am test
cargo test -p provekit-cli cmd_zoo::tests
cargo test -p provekit-cli --test zoo_smoke
```

Ran for the Paper 11 docs update:

```sh
git diff --check -- docs/papers/11-after-commits-proof-carrying-change.md docs/papers/README.md
```

Existing Rust warnings remain in generated/compiler/self-contract crates and are not introduced by this PR.

## Notes

`provekit-warnings/` remains an untracked local directory and is not part of this PR.


implementations/rust/provekit-verifier/src/proof_conformance.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    let root = match catalog.as_map() {
+        Some(root) => root,
+        None => {
+            report.push_error(PFCP_R3_ROOT_CATALOG, "catalog root is not a CBOR map");
+            return report;
+        }
+    };
+
+    if root.get("kind").and_then(CborValue::as_tstr) != Some("catalog") {
```
TSavo, line 158: The conformance checker validates `kind`, metadata, members, member CIDs, and member signatures, but it never requires or verifies the catalog-level `signer`, `declaredAt`, or `signature`. I confirmed a synthetic CBOR file containing only `{kind: catalog, members: {}}`, named with its correct BLAKE3 CID, is reported with no errors. That means unsigned or tampered catalogs can pass as conformant; this should require and verify the catalog signature over the unsigned catalog body.

implementations/rust/provekit-cli/src/cmd_zoo.rs
```diff
@@ -0,0 +1,1417 @@
...
+        .ok_or_else(|| "dropper output missing postLift.ir".to_string())?;
+    let expected_ir = read_json(specimen_dir.join(closure_ir_path))?;
+    let post_lift_cid = proof_ir_cid(post_lift_ir)?;
+    let expected_ir_cid = proof_ir_cid(&expected_ir)?;
+    if post_lift_cid != expected_ir_cid {
+        return Err(format!(
+            "dropper closure ProofIR CID mismatch: lifted {post_lift_cid}, expected {expected_ir_cid}"
+        ));
+    }
+    let transformed_artifact_cid = required_json_str(&output, "transformedArtifactCid")?;
```
TSavo: `verify_dropper` recomputes the source artifact CID and closure ProofIR CID, but it reads `transformedArtifactCid`, `postLiftCid`, and `closureWitnessCid` directly from the realizer output and only checks that the fixture receipt repeats those values. A broken or malicious realizer can return stale CIDs, mirror them in the receipt, and still pass the zoo check. Please recompute these from `modifiedSource`, the post-lift document, and the closure witness body before accepting the receipt.

implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/JavaNullBoundaryRealizer.java
```diff
@@ -0,0 +1,159 @@
...
+            patchCid,
+            artifactCid,
+            postLiftCid,
+            closureWitnessCid,
+            modifiedSource,
+            lift.postLiftJson()
+        );
+    }
+
+    private Optional<String> addProvekitRequires(String source, String methodName, String varName) {
```
TSavo, line 71: The ProvekIt-native path adds `@Requires("<proofVar> != null")` without first checking that `<proofVar>` is actually a parameter of the target method. The Spring path performs this parameter lookup, but this path can emit and re-lift a contract over an unbound variable, producing a false closure receipt. Please mirror the parameter existence check before adding the annotation.

TSavo (commented):
I found several correctness issues in the new proof/realizer path that should be fixed before this moves out of draft: unsigned catalogs can pass conformance, dropper receipt CIDs are trusted from the realizer instead of recomputed, and the Java native realizer can close a contract over a non-existent parameter.

PR #501, opened 2026-05-09
docs: clarify sufficiency vs uniqueness in paper 03 §4

TSavo:
## Summary

- Paper 03 §4 ("The DAG is a tape") correctly argues that ProvekIt's four invariants plus the witness DAG constitute a verified universal computer. That gives **sufficiency**: any computable application can be hosted as a body.
- The previous conclusion overreached by treating that result as a **uniqueness** claim ("the ceiling is theorem, not promise"), implying no other proof-carrying substrate shape could compete or coexist. A different design with the same four-invariant discipline, or a different but equally minimal set of invariants, could exist.
- Rewrite the conclusion paragraph to make the sufficiency-vs-uniqueness distinction explicit while keeping the structural Turing-completeness argument, the "EVM is the floor" framing, and the contrast with conventional blockchains (state-machine deltas + totally ordered chain) fully intact.

The claim that ProvekIt is *the* substrate at this layer is now framed as empirical (it ships, it composes, it works at four-invariant scale, competing shapes must demonstrate the same body-opacity / per-producer sovereignty / constant-size verification), not as a theorem about the universe of possible designs.

## Scope

- Touches only `docs/papers/03-substrate-not-blockchain.md`.
- §3 (theorem and proof) and §5 onward unchanged.
- Flow into §5 (per-producer sovereignty) preserved.

## Test plan

- [x] Markdown parses cleanly (file structure unchanged, only one paragraph rewritten + one new paragraph added)
- [x] No em-dashes or en-dashes introduced (`grep -P '[\x{2013}\x{2014}]'` clean)
- [x] §4's TM-correspondence (four invariants / witnesses / member bodies / DAG) preserved verbatim
- [x] §4's contrast with conventional blockchains (state-machine deltas + totally ordered chain, "strip the consensus, keep the verification") preserved verbatim
- [x] "EVM is the floor" framing kept; only the uniqueness over-claim softened


TSavo:
@copilot review

PR #506, opened 2026-05-09
menagerie: protocol-switchyard MVP for paper 10

TSavo:
## Claim

Paper 10 (After Protocol Specs) argues that protocol versions are signed content-addressed roots and migrations are witnessed edges. Until now `menagerie/protocol-switchyard/` was the only destination without a runnable anchor. This MVP closes that gap.

## What runs end to end

`cargo run --manifest-path menagerie/protocol-switchyard/Cargo.toml -- --all`

The runner:

1. reads four prose obligation specs (two for the v1 HTTP profile, two for v2) under `menagerie/protocol-switchyard/profiles/http/`;
2. computes blake3-512 CIDs for each spec (via `provekit-canonicalizer` so they match the CLI's catalog property recompute step);
3. builds two HTTP profile catalog roots naming those property CIDs;
4. shells through `provekit protocol evolve` to mint a `ProtocolEvolutionBodyClaim` and a `TruthDischargeWitness` over the migration edge;
5. prints (and emits as JSON with `--json`) the four load-bearing CIDs: `fromCatalogCid`, `toCatalogCid`, `bodyCid`, `witnessCid`.

The two boundary obligations per profile are paper-10 section-4 shaped: `request-smuggling-refusal` and `content-length-transfer-encoding`. The v2 obligations strengthen the v1 refusal by closing additional boundary cases unconditionally and adding three reason codes. The byte difference flows through the property CIDs into the profile-root CIDs, which is the substrate move paper 10 asks for.

## Honest scope

This MVP demonstrates the witnessed-edge shape. It does **not** yet:

- execute grammar conformance against the prose specs (the runner does not run a parser over them);
- mint an implementation conformance witness for any HTTP server;
- emit a paper-10-section-6 migration body with separate compatibility-invariant CIDs, bridge-checker witness CIDs, or refusal CIDs as standalone artifacts (the change is encoded as a PEP `extension-only` evolution body);
- attest the catalog under a signer (policy intentionally omits `acceptedSigner` to keep the MVP minimal).

Refusal receipts are the right shape for these gaps. The MVP exists and names what it cannot prove. README and manifest both surface this scope explicitly.

## Highest-risk simplification

Encoded the v1->v2 profile transition as `extension-only` rather than `migration-required`. This works because PEP `extension-only` only requires the version label to advance the patch component, not because the underlying obligations are extension-only in paper-10's stricter sense (v2 strengthens predicates, which paper 10 calls migration). Labeled the catalog versions `v1.0.0` and `v1.0.1` to satisfy the version-label rule while preserving "v1 profile" / "v2 profile" naming in prose. Future iterations can split this into a real `migration-required` body once a bridge-checker artifact exists to reference.

## Test plan

- [x] `cargo run --manifest-path menagerie/protocol-switchyard/Cargo.toml -- --all` exits 0
- [x] runner prints four blake3-512 CIDs that match across runs (reproducible)
- [x] `cargo build --release --manifest-path implementations/rust/Cargo.toml` succeeds, no new warnings introduced
- [x] `menagerie/protocol-switchyard/` registered as workspace member in `implementations/rust/Cargo.toml`
- [x] `menagerie/manifest.yaml` flipped to `runnable: true` with `command:` and aspirational `notes:`
- [x] `menagerie/README.md` destination table updated to Runnable
- [ ] reviewer cross-checks emitted CIDs by running locally

## Summary by CodeRabbit

* **New Features**
  * Protocol Switchyard destination is now runnable; supports HTTP v1→v2 profile migration end-to-end with reproducible artifact verification.
  * Added HTTP profile specifications for v1 and v2 defining content-length/transfer-encoding handling and request smuggling refusal obligations.

* **Documentation**
  * Updated guides with complete runnable destination instructions and detailed HTTP profile boundary specifications.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/506)


TSavo:
@copilot review

PR #541, opened 2026-05-10
Design: algebraic effects + handlers as ProofIR's missing primitive

TSavo:
Load-bearing architectural insight: generator / channel / mutex / RCU / throw / await / malloc-free are all bindings of ONE structural primitive (algebraic effect operation + handler). The earlier tower-of-constructors framing collapses into 'register effect signatures in a content-addressed catalog.' Federation across languages AND across abstraction layers becomes mechanical via signature CID matching. This doc proposes IR extensions, the EffectSignatureMemento shape, and a 7-PR implementation roadmap. Design only — no code in this PR. T Savo

### DIFF
--- a/docs/explanation/algebraic-effects-as-the-missing-primitive.md
+++ b/docs/explanation/algebraic-effects-as-the-missing-primitive.md
@@ -1,0 +1,188 @@
+# Algebraic Effects as ProofIR's Missing Primitive
+
+**Status:** load-bearing architectural design.
+**Author:** T Savo
+**Date:** 2026-05-09
+
+## The convergence
+
+If you start enumerating what ProofIR can't yet express for kernel-class verification, the list looks like this:
+
+- Heap predicates (separation logic)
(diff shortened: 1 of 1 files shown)


TSavo:
Closing — superseded by what's landed since this design draft (2026-05-10). The substrate has moved substantially: the v2 C11 catalog with declared arity_shape / per-slot evaluation / per-slot sort (#582), the asm-link-edge federation primitive, the source-unit lossless wrapper, the After-Standardization paper with the Forgery-Equals-Production lemma (#584), and the algebraic-effects-as-the-missing-primitive insight is now better expressed through the operation-CID catalog and the ORP wire format than through the framing in this PR. A fresh algebraic-effects design (effect-signature catalog + handler-resolution-as-the-Path-segment) should be written against the current substrate rather than rebasing this stale branch through 30+ intervening commits. The architectural content is preserved in the project memory; the doc-PR form is the thing being retired here.

PR #572, opened 2026-05-10
feat(c): demonstrate "the program is the proof" — thread a C11 term through the AST walk, then project it

TSavo:
## Summary

The C AST visitors in `provekit-walk-c` already compute `pre`/`post`/`wp` directly — but no first-class C11-statement `ITerm` was ever threaded through and *projected*. The term step was always implicit, then discarded. This makes it explicit for the `foo` worked example:

- Walk `foo.c` → the term `seq(if(eq(x, 0), return(neg(22)), skip), return(x))` over the C11 op-CIDs (matches the checked-in `menagerie/c11-language-signature/example/foo.term.json` as JSON-object equality).
- `project(term)` → exactly the `function-contract` already recorded in `foo.contract.json` (JSON-object equality → JCS-byte equality → same CID).

So: the program (the term) *is* the proof, in normal form; the contract is its lossy projection. Curry-Howard's converse, content-addressed.

## What's here
- `src/term_project_main.c` — a tool that walks a C function, emits the C11-op-CID term, and projects it to a contract. The demonstration.
- `tests/term_project.sh` + Makefile target — `foo.c` round-trips to the checked-in `foo.term.json`; `project(that term)` equals `foo.contract.json`'s `function-contract`. New test line: `term-project-ok`.
- `menagerie/c11-language-signature/example/` — `add.c` (a second worked example) + `add.term.json` + `add.projected-contract.json`; `foo.generated.term.json` + `foo.projected-contract.json`.
- `docs/research/2026-05-10-the-program-is-the-proof.md` — the writeup: where exactly the gap is (file:line in `provekit-walk-c` and `collectors-defensive` where the AST collapses into a contract without materializing the term), and why closing it generatively (op-CID per `CXCursorKind`, generated dispatch) is the next move.

Additive — `collectors-defensive` untouched here (#TBD carries the production term emit there). `make -C provekit-walk-c test` and `make -C provekit-lift-c-collectors-defensive test` both pass.

## Test plan
- [ ] `make -C implementations/c/provekit-walk-c test` (includes `term-project-ok`)
- [ ] `make -C implementations/c/provekit-lift-c-collectors-defensive test`

### DIFF
--- a/docs/research/2026-05-10-the-program-is-the-proof.md
+++ b/docs/research/2026-05-10-the-program-is-the-proof.md
@@ -1,0 +1,260 @@
+# The Program Is The Proof
+
+Date: 2026-05-10
+
+This is the worked proof for:
+
+```c
+static int foo(int x) { if (x == 0) return -22; return x; }
+```
+
+The claim is operationally true on this fixture: the libclang AST can be materialized as the C11 algebra term
+
+```text
+seq(if(eq(x, 0), return(neg(22)), skip), return(x))
+```
+
+and `project(term)` is byte-identical, after JSON parsing/JCS-equivalent object comparison, to the current `collectors-defensive` `function-contract` for `foo`.
+
+## Artifacts
+
+- Source: `menagerie/c11-language-signature/example/foo.c`
+- Target hand-authored term: `menagerie/c11-language-signature/example/foo.term.json`
+- Generated term: `menagerie/c11-language-signature/example/foo.generated.term.json`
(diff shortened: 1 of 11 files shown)


TSavo:
Superseded by #575 (the generated CXCursorKind dispatch replaces this PR's hand-written term_project_main.c). The program-is-the-proof writeup (docs/research/2026-05-10-the-program-is-the-proof.md) was carried over into #575, which is now merged. Closing.

PR #585, opened 2026-05-11
fix(walk-c): emit callsite actuals on all Callsite-arrival paths

TSavo:
## What changed

Establishes the invariant: **every Callsite arrival in walk-c output has an `args` key whose value is a JSON array** (possibly empty `[]`).

PR #583 wired callsite actuals into `process_call()` (the AST-derived Callsite producer) via `pk_c_walk_chain_add_callsite_arrival(... actuals_json ...)`. But synthetic Callsite arrivals produced through the generic `pk_c_walk_chain_add_arrival(... "Callsite" ...)` path left the `args` key absent — in a full kernel sweep, ~147k of ~295k wp-walk-chain mementos had no `args` key. This hardens the generic path to store `"[]"` for `kind == "Callsite"` so the serializer always emits `args`.

Note: synthetic Callsite arrivals genuinely have no AST call expression to extract actuals from, so they get `[]`. This is a schema-consistency fix (every arrival has the key), not an actuals-coverage increase for those particular arrivals — the 141,957 AST-derived ones from `process_call` already carry full actuals.

- `wp.c`: `pk_c_walk_chain_add_arrival()` stores `"[]"` as actuals_json for Callsite kind
- `tests/fixtures/callsite_actuals_paths.c`: covers plain calls, decl-initializer calls, assignment-RHS calls, conditional-arm calls
- `tests/integration.sh`: asserts all five Callsite arrivals have JSON-array `args`

## Verification

- `make -C implementations/c/provekit-walk-c` + `make test` pass
- Acceptance check on `tools/blake3-vendored/blake3.c`: 38/38 Callsite arrivals carry `args`, coverage ratio 1.0
- `git diff --check` clean

## Summary by CodeRabbit

* **Tests**
  * Added test fixtures for call-site path analysis patterns.
  * Enhanced integration tests to validate argument tracking for procedure calls across multiple calling contexts (direct, conditional, assignment, initialization).

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/585)


TSavo:
Superseded by #586. That PR removes the legacy `declarations`-mirror emission that was producing the no-args Callsite arrivals in the first place — once those duplicate mementos don't exist, hardening the generic `add_arrival` path to store `args: []` for Callsite kind is moot. #586 fixes the root cause; this PR fixed a symptom at the wrong layer.

PR #590, opened 2026-05-11
feat(csharp): add C# language lifter with RPC protocol and menagerie signature

TSavo:
## Summary

Adds a C# language lifter in C# with round-trip support (C# source → Provekit IR → C# source), integrated via the provekit RPC protocol. Feature-parity with the JVM bytecode lifter (PR #588) and C11 lifter.

## Changes

- **New project** `implementations/csharp/Provekit.Lift.Csharp/` — C# lifter using Roslyn:
  - `CsharpLifter.cs` — parses C# source via Roslyn, extracts method bodies, emits function-contract IR
  - `ContractEmitter.cs` — walks statements/expressions, emits `csharp:`-namespaced IR ops (42 ops: all binary, unary, control flow, declarations)
  - `CsharpCompiler.cs` — lowers IR terms back to C# source (full round-trip)
  - `RpcServer.cs` — JSON-RPC handler: `initialize`, `lift`, `compile`, `shutdown`
  - `Program.cs` — CLI entry point with `--rpc` flag
- **New test project** `implementations/csharp/Provekit.Lift.Csharp.Tests/` — 7 xunit tests (all passing) covering lifting, compilation, round-trip, void refusal
- **New menagerie** `menagerie/csharp-language-signature/` — 6 sorts, 42 operations, 3 equations, 5 effect signatures, 5 effect signature catalogs
- **Config updates**:
  - `manifest.toml` points `provekit lift` to the new C# lifter
  - `project_config.rs` adds `"csharp"` to `KNOWN_SURFACES`
  - `Provekit.sln` includes both new projects

## Verification

- All 7 tests pass: `dotnet test Provekit.Lift.Csharp.Tests`
- RPC protocol tested manually: initialize → lift → compile → shutdown
- `provekit mint language-signature` and `provekit mint algorithm` succeed for all specs

## Summary by CodeRabbit

* **New Features**
  * Added C# support: lift C# methods into contracts and compile IR back to C#; new CLI/RPC service exposes lift and compile operations.

* **Tests**
  * Added end-to-end test suite covering lifting of arithmetic, conditionals, loops, multi-methods, compile→re-lift round-trip, and refusal cases.

* **Chores**
  * Introduced C# language signature and effect/operation specs to the menagerie for formal reasoning.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/590)


TSavo:
## Deep review — C# source language lifter

Reviewed `csharp-lifter-roundtrip` @ `c979466f` against the ProvekIt language-signature / lossless-lift / `k(k'(I))=t` conventions (C11 v2, walk-c post-#586, the #588/#589 review history).

The bones are sound — the project is isolated to `implementations/csharp/`, there's no collision with #589 (CLR lives in `Provekit.Lift.CLR/`), the 42 ops all carry the `csharp:` namespace, `version 0.1.0` / `emits_signed_mementos=false` is honest, IR output is deterministic, and the test project is wired into CI. But **every binary-op IR term the lifter emits is malformed against its own signature**, the "lossless round-trip" claim is unfounded (no `source-unit` op, no CID-equality test), and the lifter silently fakes unhandled syntax instead of refusing it. That's a hard *Supra omnia, rectum* stop. The DIRTY state is mechanical (`Provekit.sln` + `project_config.rs` each got an appended line from #589 and #590; keep both).

### Blocking

**1. [Critical] Every binop emitted with a spurious 3rd `0` argument — no term validates against the shipped signature.** `ContractEmitter.cs`, `EmitExpression`, `BinaryExpressionSyntax` branch:
```csharp
var precond = DivByZeroPrecondition(op, rhs);   // always IntConst(0)
return Ctor(op, lhs, rhs, precond);             // EVERY binop emitted with 3 args
...
private JsonObject DivByZeroPrecondition(string op, JsonObject rhs) => IntConst(0);
```
Fires for all 19 binops (`add/sub/mul/div/mod/eq/ne/lt/le/gt/ge/and/or/bitand/bitor/bitxor/shl/shr`), not just `div`. Each op spec declares `arity: ["Int","Int"]`, `arity_shape: {named, slots:[lhs,rhs]}` — two slots. The lifter emits three args. So nothing this lifter has ever produced validates against its own menagerie signature. Fix: drop the `precond` operand; if you want a div-by-zero guard, put it in the *function-contract* `pre` (which is exactly what `op_div.spec.json` already does — `not_zero(rhs)` — the lifter just ignores it).

**2. [Critical] No `source-unit` op; "lossless" / round-trip is unfounded.** C11 ships `op_source_unit.spec.json` (`c11:source-unit(bytes:literal-unevaluated, operational_term)`) and walk-c wraps the original bytes. The csharp signature has no `op_source-unit` among the 42, and the lifter never emits one or carries the source bytes anywhere. Add the op and emit it wrapping the real source bytes (hex), as C11 does.

**3. [Critical] The "round-trip" test asserts nothing.** `LifterTests.cs::LiftAndCompileRoundTrips` is `Assert.NotNull(compiled)`. `CompilerProducesValidCsharpExpression` is `Assert.Contains("x + y", compiled)`. There is no `lift(compile(lift(src)))` re-lift, no canonical-JSON byte compare, no BLAKE3-512 CID compare. The strong-form `k(k'(I))=t` discipline is absent — these are smoke tests, not round-trip tests.

**4. [Critical] Silent information loss on unhandled syntax (the corpus-`unknown_op` anti-pattern).** `ContractEmitter.EmitStatement` has `default: return Skip();` — any statement kind it doesn't recognize (`switch`, `try/catch`, `using`, `yield`, `lock`, `fixed`, `foreach`-over-pattern, local functions, …) silently becomes `skip`. `EmitExpression` has `default: return VarExpr(expr)` — any unhandled expression (lambda, LINQ, `switch` expression, patterns, `is`/`as`, interpolated strings, `??`, `?.`, tuples, `nameof`, `default`, `typeof`, `await`, …) becomes a fresh `_t<n>` variable invented from nothing. It also emits fallback op names `csharp:binop` / `csharp:unop` / `csharp:postop` / `csharp:postfixop` that aren't in the 42-op signature. A construct that can't round-trip must be a `Refusal`, not faked.

**5. [Major] Path-traversal — `lift`/`compile` accept absolute paths and `..`.** `CsharpLifter.LiftPaths`: `Path.GetFullPath(Path.Combine(workspaceRoot, sourcePath))` with no containment check. `Path.Combine("/ws", "/etc/passwd")` → `/etc/passwd`; `"../../etc/x"` escapes. Same class CodeRabbit flagged Critical on #588's JVM lifter. After `GetFullPath`, reject if not `StartsWith(GetFullPath(workspaceRoot) + dirSeparator)`.

**6. [Major] Contract names collide for nested / partial / overloaded methods.** `ContractEmitter.Emit` uses `fnName = _method.Identifier.Text` only. `OuterA.Foo()` and `OuterB.Foo()`, or two overloads `Add(int,int)` / `Add(long)`, all produce contract `"Add"`. #589 had to fix exactly this for CLR (`{assembly}!{type}::{method}#{sig}`). Use a qualified name (`Namespace.Type+Nested.Method(sig)`).

**7. [Major] Dangling effect-signature reference.** `op_div.spec.json` declares `effects: [{kind:"effect-signature", name:"Trap"}]`, but the signature's `effect_signatures` list is `Call / MemWrite / MemRead / Alloc / Loop` — no `Trap`. Add `effsig_trap.spec.json` (+ `eff_trap.spec.json`) or remove the reference.

**8. [Major] `op_call` / `op_new` arity mismatch.** `op_call.spec.json` is `arity: ["String"]`, formals `["callee","args"]` (args a `set` slot). The lifter emits flat-variadic `["csharp:call", StrConst(name), arg0, arg1, …]` — N+1 positional args. Same shape problem as #1. Pick one shape (nested arg-list term, or a genuinely-variadic spec) and make lifter + spec agree.

### Non-blocking

1. [Minor] `locus` in every spec points to `menagerie/csharp-language-signature/README.md` — there is no README, only `specs/`. C11 ships one. Add it or fix the loci.
2. [Minor] Equations `eq_seq_assoc` / `eq_seq_skip_left` / `eq_seq_skip_right` are correctly stated (seq monoid, `skip` identity) but reference no discharge backend — decorative JSON. Fine for `0.1.0-draft`; note it. Set is thin (seq laws only).
3. [Minor] `op_and` / `op_or` don't mark the right slot `evaluation: "unevaluated"`. C11's `c11:and` also omits it (note lives in `effects.rule` prose) so this *matches* C11 — but neither is ideal. `op_source_unit` shows the `evaluation: "unevaluated"` slot field works; consider using it.
4. [Minor] `int`-only slice: non-int return → `Refusal` (good, loud), but all params and the return are hardcoded `PrimSort("Int")` regardless of actual type — `formalSorts` lies for any `bool`/`string`/… param. Fix or document.
5. [Minor] `Unit()` == `IntConst(0)` — sort-hygiene smell (Unit ≠ Int 0), though self-consistent with `eq_seq_skip_*` (which use `{kind:const,value:0}` as `skip`'s arg).
6. [Nit] `CsharpCompiler` emits `/* unhandled: {name} */;` for non-`csharp:` ops — visible, but produces non-compiling C#. Should throw.
7. [Nit] `manifest.toml version = "0.1.0"`, `Initialize` reports `ir_version "v1.1.0"` — confirm `v1.1.0` is what `Provekit.IR` actually emits.

### 60-second summary for the next reader

Look at `ContractEmitter.cs` `BinaryExpressionSyntax` branch first. Every `csharp:add/…/shr` term carries a spurious third `0` operand, contradicting all 19 of its own op specs, and there's no test that would catch it because the round-trip test asserts only `!= null`. Fix that, add a real `lift(compile(lift(src)))` CID-equality test, make the lifter *refuse* (not fake) syntax it can't round-trip, add the `source-unit` op — then re-review.

— automated deep review (Opus)

PR #591, opened 2026-05-11
feat: add EVM bytecode lift kit

TSavo:
## Summary

- Add `provekit-lift-evm-bytecode`, a JSON-RPC lift plugin for `evm-bytecode`, `evm-assembly`, and `evm-hex`.
- Lift straight-line EVM stack programs into function-contract IR and refuse unsupported storage/control/call/log/memory effects loudly.
- Wire the kit through mint/prove/CICP/Make/GitHub Actions, with a menagerie EVM bytecode language signature.

## Validation

- `cargo test -p provekit-lift-evm-bytecode --test smoke`
- `cargo test -p provekit-cli --bin provekit resolve_kit_all_13_ci_kits`
- `cargo test -p provekit-cli --bin provekit known_lists_include_anchor_entries`
- `cargo test -p provekit-cli --test mint_kit_integration evm_bytecode_kit_round_trips_stack_program_through_cli_mint`
- `cargo fmt --all -- --check`
- `git diff --check`
- `make prove-evm-bytecode`
- `provekit ci shadow/result --kit evm-bytecode` local smoke

## Release Notes

* **New Features**
  * Added EVM bytecode language support with a complete suite of operations including arithmetic, comparisons, and bitwise transformations
  * Extended CI/CD pipeline to include EVM bytecode conformance testing across all supported environments

* **Tests**
  * Added comprehensive integration and smoke tests for EVM bytecode handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/591)


TSavo:
## Deep review — EVM bytecode lift kit

Reviewed `codex/evm-bytecode-lifter` against the JVM-bytecode lifter (#588), the asm lifters (#552/#553), the canonical `Effect` schema in `provekit-walk/src/contract.rs`, and the operation-CID/signature discipline. Also checked the 4 Copilot comments.

The scaffolding is clean and matches the JVM precedent (CLI wiring, CICP witnesses, manifest, CI matrix, `Refusal` discipline, `bodyCid: null`). But the lifter contradicts its own signature spec on `RETURN`, the `pc` field means different things for the two input dialects, and all four Copilot findings are unaddressed. The `RETURN` desync is a *Supra omnia, rectum* stop, not an MVP-scope decision.

### Blocking

1. **[Major] `RETURN` semantics: spec ↔ impl disagree, in the same PR.** `evm_bytecode_signature.spec.json:283-288` declares `return` carries `effects: ["memory_read"]`; `lib.rs:705-724` emits the `RETURN`-terminated `function-contract` with `"returnSort": evm_word_sort()` (a 256-bit stack word) and `"effects": []`. `RETURN(offset, size)` yields a *memory slice*, not a stack word, and it reads memory. Fix: either *refuse* `RETURN` until a bytes-return sort + a `memory_read`/`reads` effect exist, or model it with the right return sort and emit the `reads` effect. Don't ship a contract that contradicts its own op spec.

2. **[Major] `Instruction.pc` is instruction-index for assembly input, byte-offset for hex input.** `lib.rs:340` advances the assembly `pc` by `instructions.len()` (an index); `lib.rs:360-399` uses real byte offsets for hex. EVM PCs are byte offsets — the spec and README say so (`JUMPDEST` slot = "byte offset into deployed EVM code"). So the same program lifted from `.evmasm` vs `.evmhex` produces different `pc` values → different IR → different CID. Fix: advance the assembly `pc` by encoded length (`1 + push_width`). (Copilot #3.)

3. **[Major] `STOP` on an empty stack is misreported.** `lib.rs:729` + `687-703`: `STOP` with nothing on the stack → `Step::Stop(None)` → falls through to the `"unterminated-bytecode"` refusal whose reason says "stream ended without STOP or RETURN" — false, it *did* hit STOP. Fix: a distinct `"empty-stack-at-stop"` refusal (or define an explicit STOP-with-no-return-value convention). (Copilot #1.)

4. **[Major] Signature catalog omits ops the lifter emits.** `lib.rs:743-748` emits `evm:lt / evm:gt / evm:and / evm:or / evm:xor / evm:not`, but `evm_bytecode_signature.spec.json` has no specs for them. The language signature *is* the operation-CID catalog; if the lifter emits an op not in the catalog, the catalog is incomplete (and the catalog-completeness check would catch it). Add the op specs (or stop emitting those ops). (Copilot #2.)

### Non-blocking

- [Minor] `lib.rs:287-293` `looks_like_hex_bytecode` strips whitespace/comments before testing, so an assembly file whose opcode tokens are all hex-digit + even total length (e.g. a program using `DEAD` as a label) is misclassified as hex. Anchor on extension (`.evmhex` vs `.evmasm`) when available, or require the `0x` prefix. (Copilot #4 area.)
- [Nit] `lib.rs:209` `initialize` reports `version: "0.1.0"` but the signature spec is `"0.1.0-draft"` — align.
- [Nit] README line 6 ("instruction streams are positional") vs the spec declaring every op `arity_shape: {kind:"named"}` — defensible (matches the JVM spec) but the prose oversells.
- [Follow-up, not blocker] No `opaque_loop`/`loopCid` handling — `JUMP`/`JUMPI` are refused, so straight-line only. Fine for v1.
- [Follow-up, not blocker] No IR→bytecode realizer / round-trip test. Parity with #588: JVM's round-trip lives in the separate `provekit-ir-compiler-jvm-bytecode` crate, not the lifter PR. Known follow-up — but note EVM's smoke test, unlike JVM's, doesn't assert the contract CID is byte-deterministic.
- [Verify before merge] The `.NET 10 (preview) → .NET 10` + "11→12 kits" hunks in `ci.yml`/`Makefile` are codex Boy-Scouting rolled into this PR. If .NET 10 hasn't GA'd on the `setup-dotnet@v4` feed, dropping `dotnet-quality: 'preview'` silently breaks csharp/clr-bytecode CI.

### Conformance checklist

| Check | Result |
|---|---|
| namespace prefix on ops | ✅ `evm:add` etc. (matches JVM's `jvm:`) |
| `arity_shape` declared & sane | ✅ all `named`-slot (matches JVM spec pattern) |
| `source-unit` op present & wired | declared in spec only; lifter never emits it — but JVM's doesn't either. Consistent. |
| effects on canonical wire shapes & sorted | N/A — lifter emits `effects: []` unconditionally; never exercises the `Effect` shapes (hand-built JSON, not the `provekit-walk` enum) ⚠️ while declaring effect-bearing ops in its spec |
| unknown-opcode handling | ✅ decoder → `Opcode::Unsupported` → lift-time `Refusal` (loud, not silent-fake) |
| round-trip / re-lift CID-equality test | ❌ none |
| CICP witnesses consistent | ✅ jobKeys match kits, inputCids sorted & cross-referencing; the rust-witness regen is expected (crate joins the rust workspace) |
| self-contract pins untouched | ✅ `RUST_KIT_FULL_SELF_CONTRACT_SURFACE_CID` + `.provekit/self-contracts-attestations/` unchanged (correct — a new kit shouldn't move the rust self-surface) |
| deterministic JSON output | ✅ `json!` macros + `Vec` (no `HashMap` iteration), no timestamps |
| smoke test meaningful | ◐ asserts contract shape / `fnName` / refusal behavior / RPC wrapping; does *not* assert byte-deterministic contract CID |
| Copilot comments addressed | ❌ 0 of 4 |
| `ci.yml` matrix entry | ✅ `evm-bytecode` added to `prove-linux`; the `dotnet-quality` removal is the only YAML risk |

### Look at first (60 s)

`evm_bytecode_signature.spec.json:283-288` (`return → effects:["memory_read"]`) next to `lib.rs:713-717` (`returnSort: evm_word_sort()`, `effects: []`). The spec and the lifter disagree about what `RETURN` does, in the same PR. That's the gate. Fix that, fix the `pc` dialect inconsistency and the STOP-refusal-reason, add the six missing op specs (or stop emitting them), and confirm the .NET 10 hunk doesn't break csharp/clr CI — then re-review.

— automated deep review (Opus)

PR #592, opened 2026-05-11
feat(go): add Go source lift kit (self-hosted) + go-language-signature

TSavo:
Self-hosted Go source-language lift kit + `menagerie/go-language-signature/`. Mirrors PR #590 (the C# source lifter).

- Go lifter via stdlib go/parser + go/ast + go/types → `function-contract` mementos over `go:`-namespaced ops
- per-file `go:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test
- new `go-source` lift surface (existing `go` kit surface manifest untouched — the lesson from #590); `version 0.1.0-draft`
- go build/test/vet ./... pass incl nested modules

## Release Notes

* **New Features**
  * Added a Go source lifter that parses Go source code and generates intermediate representation with function contracts and effect tracking.
  * Added a Go code generator that compiles intermediate representation back to valid Go source.
  * Added JSON-RPC interface for remote invocation of lift and compile operations.
  * Added support for tracking side effects including reads, writes, I/O, unsafe operations, and panics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/592)


TSavo:
## Review — PR #592 `feat(go): add Go source lift kit (self-hosted)`

Reviewed `origin/codex/go-lift-kit` @ `68dab7c0` against `origin/main` @ `c9ba0c0e` (#590 merged). Read all 11 Go source files, all 60 spec JSON files, the lift-plugin protocol spec, the canonical `Effect` enum in `libprovekit/src/compose.rs`, and the C# exemplar from #590. Built + tested the whole Go workspace (`provekit-ir-symbolic`, `provekit-self-contracts`, `provekit-lift-go-tests`, `provekit-lift-go`) — all green.

The PR description claims "full round-trip support"; the round-trip *test* is a tautology that never exercises the IR→source compiler, and two op-specs disagree with their own lifter. None of this is hard-stop-dangerous, but under *Supra omnia, rectum* a test that makes a false correctness claim and a `menagerie/manifest.yaml` entry that's inconsistent with how every other language-signature is registered should be fixed before merge. Everything load-bearing for CID-correctness (canonicalizer reuse, effect wire shapes, no catch-all op) is correct.

---

### Blocking

**B1. `TestRoundTripCompileLiftIsByteIdentical` proves nothing about the contract→source compiler.**
`compile.go:15-20`:
```go
func Compile(input CompileInput) (CompileOutput, error) {
	if src, ok := sourceFromSourceUnit(input.IR); ok {
		return CompileOutput{Source: src}, nil
	}
	return compileContracts(input.IR)
}
```
`LiftSource` always appends a `go-source-unit` whose bytes-slot is the hex of the original source (`lift.go:104-124`). So when `lift_test.go:75` calls `Compile(first.IR)`, `sourceFromSourceUnit` matches, hex-decodes, and returns the *verbatim original source*. `LiftSource` on identical bytes then trivially yields identical IR. The "round-trip" test (`lift_test.go:70-94`) is `hex.Decode(hex.Encode(src)) == src`. It never touches `compileContracts` / `goExprFromIRTerm`. #590's analogue (`Provekit.Lift.Csharp.Tests/LifterTests.cs:91-119`) explicitly plucks `contract.post.args[1]` and calls `CompileBody` on it — that's the test this PR needs. **Fix:** add a test that round-trips through `compileContracts` (or `goExprFromIRTerm` directly) on an IR set with the source-unit stripped, and assert byte/CID identity of the resulting `function-contract`. As written, the strongest claim in the PR is hollow.

**B2. `menagerie/manifest.yaml` `go-language-signature` entry is inconsistent with every other language signature.**
The diff adds:
```yaml
  - id: go-language-signature
    path: go-language-signature
    runnable: planned
    claim: draft Go source operation algebra for provekit-lift-go
```
But `c11-language-signature`, `csharp-language-signature`, `jvm-bytecode-language-signature`, `aarch64-language-signature`, and `rust-language-signature` all exist as `menagerie/<id>/` directories and **none** of them have a `menagerie/manifest.yaml` `destinations:` entry (`#590` did not add one for csharp either). Either every language-signature gets a manifest entry or none does; one-off-for-Go is wrong. **Fix:** drop the entry (matches #590's precedent), or, if these *should* be registered, do it uniformly in a separate PR.

---

### Non-blocking

- **N1. `go.sum` is in the diff** (`implementations/go/provekit-lift-go/go.sum`, 4 lines) — CodeRabbit's "ignored due to path filters" note is wrong; it's committed. That's fine and expected for a real module; just noting it's there.
- **N2. `op_skip.spec.json` declares formal `unit` (`arity: ["Unit"]`) but the lifter emits `op("go:skip")` with zero args** (`lift.go:285,342,411,427-...`), and `eq_seq_skip_left.spec.json` likewise uses `go:skip` with `args:[]`. The Go side is *internally* consistent, but it diverges from its own op-spec and from the csharp exemplar (`csharp:skip` carries a `Unit` ctor arg; `eq_seq_skip_left` there uses `{"kind":"const","value":0}`). Pick one — either drop the `unit` formal from `op_skip.spec.json` or emit the arg.
- **N3. `op_call.spec.json` `arity_shape` omits the `shape:{kind:"set"}` marker on the `args` slot.** c11 and csharp both write `{"name":"args","shape":{"kind":"set"}}` to signal "this slot absorbs the variadic tail"; Go writes `{"name":"args","slot_sort":"Term"}`. The lifter emits variadic `[callee, arg0, arg1, ...]` (`lift.go:638-640`) — which is the *correct* convention, matching #590's `Ctor("csharp:call", callArgs)` — but the spec then claims exactly 2 children. Add the `set` shape.
- **N4. `op_source_unit.spec.json` uses `"slot_sort": "Literal"` / `formal_sorts:[{...,"name":"Literal"}]`** (capitalized) where c11 and csharp use lowercase `"literal"` for the bytes slot. ProvekIt isn't standardized on this casing across kits, so it's cosmetic, but mirror the exemplars.
- **N5. Partial-refusal source-unit incoherence.** `lift.go:104` only builds the `go:source-unit` when `len(bodyTerms) > 0`, and `bodyTerms` only accumulates bodies of *non-refused* functions — yet the bytes-slot still carries the *whole file*. A file with one good fn + one refused fn ships a source-unit whose `operational_term` covers a strict subset of what its byte-slot decodes to. Defensible for a `0.1.0-draft`; worth a `notes:` line or a follow-up.
- **N6. Top-level `arity_shapes` map in `go_language_signature.spec.json`.** Not present in the c11/csharp signature specs (they keep `arity_shape` per-op only). Harmless duplication; either remove it or accept it as a convenience index.
- **N7. `fallbackFuncName` collision on the type-check-failure path.** `exprTypeString` returns the literal `"Recv"` for receiver type-exprs it doesn't recognise (`lift.go:993`); two methods on exotic receiver types in a file that fails `types.Config.Check` would collide. On the happy path `types.Func.FullName()` is collision-free (pointer-vs-value receivers distinct). Edge-case; mention in a comment or refuse instead of falling back.

---

### Conformance checklist

| # | Item | Status |
|---|------|--------|
| 1 | `go:` on every op; no `go:binop`/`go:unknown` *catch-all* (real `go:skip` no-op is fine) | **PASS** — all 36 ops `go:`-prefixed; `liftStmt`/`liftExpr` `default:` branches error → refusals, never a fallback op (`TestRefusesUnsupportedGoStatementWithoutUnknownOp` asserts no `go:unknown`/`go:binop` in IR) |
| 2 | `arity_shape` on every op spec & sane; binops `Named [lhs,rhs]`; `go:seq` `Positional`; `go:assign` Named target/value; `go:and`/`go:or` mark `rhs` `evaluation:"unevaluated"` | **PASS** — verified each. `go:and`/`go:or` `rhs` is `evaluation:"unevaluated"`; `go:seq` is `{"kind":"positional","arity":2}`; `go:source-unit` bytes slot is `evaluation:"unevaluated"`. (N3: `op_call` missing the `set` shape marker.) |
| 3 | `go:source-unit(bytes, operational_term)` present AND emitted wrapping real source bytes | **PASS** — `op_source_unit.spec.json` present; `lift.go:106-110` emits `op("go:source-unit", {kind:"bytes",encoding:"hex",value:hex(source)}, body)` on the lift path; `TestLiftSimpleAdd...` asserts the bytes slot |
| 4 | Refuse-not-fake: loud `Refusal{kind,function,line,reason}` for unhandled AST; goroutines/channels/`defer`/`select`/`go`/generics/closures must be *refused* | **PASS** — `*ast.GoStmt`, `*ast.DeferStmt`, `*ast.SendStmt`, `*ast.SelectStmt`, `*ast.SwitchStmt`, `*ast.ChanType`, `*ast.FuncLit`, `*ast.CompositeLit` all error → `refuse("unsupported-syntax",...)` (`lift.go:343-356,591-595,137-139`). `recover` is treated as an opaque `unresolved_call` (not refused, not faked — defensible). `Refusal` carries `{Kind,Function,Line,Reason}` |
| 5 | `fnName` via `types.Func.FullName()` (pkg-qualified, ptr-vs-value receiver distinct); no collision | **PASS (with N7)** — `FullName()` on happy path; syntactic fallback distinguishes `*T` vs `T` but degrades to `"Recv"` for unrecognised receiver exprs on the type-check-failure path |
| 6 | Effects on canonical wire shapes (`reads`/`writes`→`target`, `unresolved_call`→`name`, `opaque_loop`→`loopCid`, `io`/`panics` bare); NO `alloc`/resolved `call`; sorted by `sort_key()` order incl. secondary sort | **PASS** — `Effect` JSON tags + `effectSortKey` (`types.go:23-28,196-215`) match `libprovekit/src/compose.rs` `Effect::to_value`/`sort_key` exactly (`6:opaque_loop:`, `5:unresolved_call:` prefix differs from Rust's internal `5:unresolved:` but never leaks into output bytes). No `alloc`, no resolved `call`. `TestEffectsUseCanonicalShapesAndSortOrder` pins exact bytes `[{"kind":"writes","target":"...G"},{"kind":"panics"}]` |
| 7 | Round-trip test: `lift(compile(lift(src)))` asserting byte/CID identity (not fuzzy) | **FAIL (B1)** — byte-exact compare, but `Compile` short-circuits on the source-unit (`compile.go:16-19`) so it never exercises `compileContracts`; the test reduces to `hex.Decode(hex.Encode(src))==src` |
| 8 | Determinism: no `map` iteration order in emitted JSON; sorted keys; no timestamps | **PASS** — effect set keyed by string then `sort.Strings` (`types.go:183-194`); JCS encoding via `provekit-ir-symbolic/canonicalizer`; `op()` builds slice-ordered args; no timestamps. `marshalJSONNoHTML` disables HTML escaping for byte-stable output |
| 9 | Existing `go/manifest.toml` NOT repointed; NEW `go-source` surface added; `version` `0.1.0-draft`; no changes to `.provekit/ci/accepted/*`, `.provekit/self-contracts-attestations/*`, `RUST_KIT_FULL_SELF_CONTRACT_SURFACE_CID` | **PASS** — `git diff` empty for `go/manifest.toml` and `go-self-contracts/manifest.toml`; new `go-source/manifest.toml` added (`name="go-source-lift"`, `version="0.1.0-draft"`); no touch to CI accepted / attestations / surface CID. (#590's mistake explicitly avoided — mirrors how `clr-bytecode` coexists with `csharp`.) |
| 10 | `--rpc` server: `initialize`/`lift`/`compile`/`shutdown` shapes match the protocol spec / C# `RpcServer.cs` | **MOSTLY PASS** — `lift` returns `{kind:"ir-document", ir, callEdges:[], diagnostics, opacityReport:[], refusals}`, byte-for-byte the shape #590's `LiftSuccessResponse` returns. `initialize` returns the capability block. `compile` is *not in the protocol spec* (`initialize`/`lift`/`shutdown` only) — both #590 and this PR invented it; #590 returns `{kind:"compiled-formula",body}`, this PR returns `{source}`. Divergence is real but `compile` is off-spec on both sides, so non-blocking. Note: `lift` params accept `workspace_root`+`source_paths` (no `surface`/`options.layer`) — same as the existing `provekit-lift-go-tests` RPC, so consistent with current repo practice even though the spec's grammar wants `surface`+`options` |
| 11 | `menagerie/manifest.yaml` `go-language-signature` entry correct? | **FAIL (B2)** — no other `*-language-signature` has a `manifest.yaml` entry; this is a one-off inconsistency |
| 12 | Existing Go kit (`provekit-ir-symbolic`, `provekit-self-contracts`, `provekit-lift-go-tests`, `provekit-lift-go-validator`) still builds + tests | **PASS** — `go build ./... && go test ./...` green in every module; `Makefile` `build-go`/`test-go` updated to include `provekit-lift-go` |
| 13 | Lifter uses `provekit-ir-symbolic`'s canonicalizer/BLAKE3 (byte-identical to Rust's `provekit-canonicalizer`), no new canonicalizer | **PASS** — `types.go:9` imports `provekit-ir-symbolic/canonicalizer`; `canonicalCID` → `canonicalizer.EncodeJCS` + `canonicalizer.ComputeCID`; `git diff` shows zero changes under `provekit-ir-symbolic/` or `provekit-canonicalizer/`. No new canonicalizer code anywhere. *Supra omnia, rectum* satisfied |

---

`compile.go:15-20` together with `lift_test.go:70-94`. The source-unit short-circuit makes the only round-trip test a no-op against the actual compiler. The PR description's "full round-trip support" claim is not backed by a test. (Compare `Provekit.Lift.Csharp.Tests/LifterTests.cs:91-119` in `c9ba0c0e` for the test shape that actually exercises IR→source.)

### vs #590's quality bar

Roughly on par. #590 also invented an off-spec `compile` method, also added per-language menagerie specs with the same per-op `arity_shape` convention, also gave the source lifter its own surface manifest (after a fixup commit). This PR is *cleaner* on the manifest front — it didn't repeat #590's mistake of repointing the kit surface, so no fixup commit was needed — but *weaker* on the round-trip test (#590's actually drives `CompileBody`; this one doesn't). Effects/canonicalizer/refusal handling are solid in both. Land it once B1 (real round-trip test) and B2 (drop the `manifest.yaml` entry) are addressed; the N-items can ride along or follow up.

— review by Claude (Opus 4.7, 1M ctx) on behalf of T Savo

PR #593, opened 2026-05-11
feat(java): add Java source lift kit (self-hosted) + java-language-signature

TSavo:
Self-hosted Java source-language lift kit + `menagerie/java-language-signature/`. Mirrors PR #590 (the C# source lifter).

- Java lifter via the JDK compiler API com.sun.source.* → `function-contract` mementos over `java:`-namespaced ops
- per-file `java:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test
- new `java-source` lift surface (existing `java` kit surface manifest untouched — the lesson from #590); `version 0.1.0-draft`
- mvn -pl provekit-lift-java-source -am test pass; jar RPC reports 0.1.0-draft


TSavo:
## Review — PR #593: `feat(java): add Java source lift kit (self-hosted, javac API)`

Hard skeptical pass, against the bar set by #590 (C# source lifter) and the in-tree exemplar `implementations/java/provekit-claim-envelope/`.

One sentence: the kit is well-built (refuse-not-fake is honest, ops are `java:`-namespaced, `fnName` erases param types correctly, `arity_shape` is present everywhere, full Java reactor builds + 130-odd tests green) — but it introduces a **second, hand-rolled BLAKE3 + JCS encoder** (`provekit-ir/.../Jcs.java`) into a Java tree that already has a vetted one (`provekit-claim-envelope/.../Jcs.java` + `Blake3.java`, BouncyCastle, "Mirrors rust/csharp 1:1"), with **no cross-impl byte-equivalence test** tying them together — exactly the substrate-drift surface `Supra omnia, rectum` and the content-address rule exist to forbid.

---

### Blocking

**B1 — `implementations/java/provekit-ir/src/main/java/com/provekit/ir/Jcs.java:1-521` — forked canonicalizer + hand-rolled crypto.**
Two facts:
- `provekit-claim-envelope` already ships `Jcs.java` (a `Value`-tree JCS encoder) and `Blake3.java` (`org.bouncycastle.crypto.digests.Blake3Digest`), both explicitly documented as 1:1 mirrors of `provekit-canonicalizer` rust/csharp.
- This PR adds a parallel `Jcs.java` in `provekit-ir` with its own `Json` value tree **and a ~150-line hand-rolled BLAKE3** (the `private static final class Blake3` block, ending the file at line 521). `provekit-lift-java-source` then depends on the `provekit-ir` one, not the audited one.

For the data this kit emits *today* (ASCII keys, `long`-only integers, `<0x20` string escapes) the new encoder does produce byte-identical output to its peers — I'm not claiming wrong CIDs in the field. The blocking concern is structural: two independent BLAKE3 implementations and two JCS encoders in one language tree, with no test asserting `ir.Jcs.encode(x) == claimenvelope.Jcs.encode(x)` and no test pinning the new encoder's bytes against `provekit-canonicalizer` golden vectors. The PR's own tests (`CanonicalJsonTest.blake3_512MatchesKnownVectors`, `...HandlesMultiChunkInput`) test the *hash function*, not the *JCS encoding*, and never touch the existing peer. "Designed to match" is not "proven to match."
*Fix:* extract `claimenvelope`'s `Jcs.java` + `Blake3.java` into a shared module (e.g. `provekit-canonical-java`) that both `provekit-ir` and `provekit-claim-envelope` depend on; delete the in-tree `Blake3`; if `provekit-ir` genuinely needs a parse-capable value tree the `claimenvelope` one lacks, add `parse()` *there*, don't fork. Add a cross-impl byte-equivalence test.

**B2 — `provekit-ir/.../Jcs.java` (the `encodeValue` `Obj` branch) — object-key sort uses UTF-16 code-unit order, not Unicode code-point order.**
`sorted.sort(Comparator.comparing(Field::key))` ⇒ `String.compareTo` ⇒ UTF-16 code-unit comparison. Spec §7.3 (`protocol/specs/2026-04-30-canonicalization-grammar.md`) and RFC 8785 §3.2.3 require **code-point** order. The sibling `provekit-claim-envelope/Jcs.java` was deliberately written with a `Jcs::compareCodePoints` comparator *and a comment about exactly this hazard*; the rust peer (`provekit-canonicalizer/src/jcs.rs`) gets it for free because Rust `String` cmp is byte-order = code-point order. The new Java encoder is the only one in the family that regresses. Harmless today (all keys ASCII) but it's a latent divergence and below the established bar — fold into the B1 consolidation. *Fix:* use a code-point comparator (or just reuse the consolidated `Jcs`).

---

### Non-blocking

- **N1 — req 10, brief premise is wrong.** `menagerie/manifest.yaml` is a `destinations:` list of *runnable exhibits* (`bug-zoo`, `bridgeworks`, …), not a registry of language-signature dirs. None of `c11/csharp/rust/jvm/aarch64-language-signature` appear in it on `main` either. The agent is **correct** not to add `java-language-signature` there; no inconsistency. (If Go/Python/TS PRs added a manifest entry, it was a different file.)
- **N2 — `JavaSourceCompiler` is a stub.** `compile()` for any non-`source-unit` term emits `class Lifted { int f(int x0, int x1) { ... } }` with a hardcoded 2-arg signature and `99:`-style fallthroughs in `emitStmt`/`emitExpr`. The round-trip test (`JavaSourceLifterTest.sourceUnitCompilerRoundTripsToByteIdenticalLiftedTerm`) only exercises the `java:source-unit` fast path (re-lift the embedded source bytes), so it proves source-bytes recoverability, not operational-term realizability. Fine for `0.1.0-draft`, but the round-trip claim is narrower than it looks — say so in the kit README.
- **N3 — `JavaSourceLifter.java` `Effect.sortKey()` still has a `default -> "99:" + kind + ...` arm.** Unreachable given the closed `Effect` factory set, but it's dead defensive code that, if ever hit, would emit a non-canonical sort prefix. Either drop it or make it `throw`.
- **N4 — manifest `command` path.** `implementations/java/.provekit/lift/java-source/manifest.toml` → `["java","-jar","provekit-lift-java-source/target/provekit-lift-java-source.jar","--rpc"]`; `Program.main` correctly dispatches `--rpc` to `RpcServer.run()`, shade plugin sets `finalName=provekit-lift-java-source` and `mainClass=...Program`. Wiring is consistent. Existing `.provekit/lift/java/manifest.toml` is untouched (good); no `.provekit/ci/accepted/*` changes (good); version `0.1.0-draft` (good).
- **N5 — `eff_alloc`/`eff_call` exist in `csharp-language-signature` but the Java kit (correctly per the brief) has no `alloc`/resolved-`call` effects.** That's a #590 problem, not a #593 one — flagging for the C# side.

---

### Conformance checklist

| # | Item | Status |
|---|---|---|
| Jcs(a) | Did `provekit-ir` have a canonicalizer/BLAKE3 before? | No — but `provekit-claim-envelope` did. New one **forks** the substrate. **B1** |
| Jcs(b) | Byte-identical to rust `provekit-canonicalizer`? | Yes for today's data shapes; **key sort uses UTF-16, not code-point order — B2**; no cross-impl test — B1 |
| 1 | `java:` prefix on every op, no catch-all | Pass — `op_*.spec.json` all `java:`; lifter `default -> throw refuse(...)` everywhere; float/double literals refused (not in expr switch) |
| 2 | `arity_shape` present + sane; `&&`/`||` mark rhs `unevaluated`; `seq` Positional | Pass — `op_and`/`op_or` mark `rhs` `evaluation:"unevaluated"`; `op_seq` `{"kind":"positional","arity":2}`; `op_assign` Named target/value |
| 3 | `java:source-unit(bytes, operational_term)` op + emitted with real source bytes | Pass — `ctor("java:source-unit", Jcs.string(source), operational)` |
| 4 | Refuse-not-fake (lambdas, streams, switch-exprs, try-with-resources, anon/local classes, records, generics, varargs) | Pass — `Refusal{kind,fnName,line,reason}`; varargs/generic methods/generic invocations/anon classes/void-return all refused; statement+expression switches `default -> throw refuse(...)` |
| 5 | `fnName` provably unique (binary FQN + erased params) | Pass — `elements.getBinaryName(type) + "." + name + "(" + types.erasure(p.asType()) joined ")"` |
| 6 | Effects on canonical wire shapes, sorted by `sortKey()` | Pass — `reads/writes/io/unresolved_call(name)/opaque_loop(loopCid:"blake3-512:...")/panics(throw)`; no `alloc`/resolved-`call`; sorted via `Comparator.comparing(Effect::sortKey)` with numeric prefixes |
| 7 | Round-trip test = byte/CID identity | Partial — only the `java:source-unit` source-bytes path (N2) |
| 8 | Determinism: no `HashMap` into emitted JSON; sorted keys; no timestamps | Pass — `LinkedHashMap`/`LinkedHashSet`/sorted `Files.walk`; effects keyed+sorted; no timestamps |
| 9 | Wiring (new surface manifest, `0.1.0-draft`, no CI-accepted/self-pin churn) | Pass (N4) |
| 10 | Not registering in `menagerie/manifest.yaml` | Correct — brief premise wrong (N1) |
| 11 | Existing Java reactor still builds | Pass — `mvn test` across `provekit-ir` + `provekit-claim-envelope` + `provekit-java-self-contracts` + all lift adapters + integration-tests: BUILD SUCCESS, 0 failures |
| 12 | `--rpc` `initialize`/`lift`/`compile`/`shutdown` match `2026-04-30-lift-plugin-protocol.md` | Pass on visual inspection — JSON-RPC 2.0 envelope, `protocol_version:"provekit-lift/1"`, `SURFACE_NOT_SUPPORTED` 1003, standard `-32xxx` codes, path-traversal guard in `liftPaths` |

---

`implementations/java/provekit-ir/src/main/java/com/provekit/ir/Jcs.java` — specifically the bottom of the file: a hand-rolled `Blake3` next to `provekit-claim-envelope`'s BouncyCastle one. Decide consolidation before merge.

### vs #590's bar
#590 reused `Provekit.Canonicalizer` (the shared C# canonicalizer module) rather than minting a parallel one in its lift package. This PR did the opposite for Java. Bring it to parity: one canonical-bytes module per language, depended on by IR + claim-envelope + every lift kit. Everything else here meets or exceeds #590 (the refusal taxonomy and `fnName` erasure are notably more careful).

— T Savo

PR #594, opened 2026-05-11
feat(python): add Python source lift kit (self-hosted) + python-language-signature

TSavo:
Self-hosted Python source-language lift kit + `menagerie/python-language-signature/`. Mirrors PR #590 (the C# source lifter).

- Python lifter via stdlib ast → `function-contract` mementos over `python:`-namespaced ops
- per-file `python:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test
- new `python-source` lift surface (existing `python` kit surface manifest untouched — the lesson from #590); `version 0.1.0-draft`
- pytest pass 8 tests; signature --check pass


TSavo:
## Review: PR #594 — `feat(python): add Python source lift kit (self-hosted, ast)`

**Verdict: CHANGES-REQUIRED** — the kit rolls its own JCS canonicalizer (`json.dumps`) instead of reusing the existing `provekit_lift_py_tests.canonicalizer`, and it produces non-byte-identical canonical JSON the moment a string contains a control char (newline/tab/CR) — which the `python:source-unit` term always does. Divergent CIDs = a hard *Supra omnia, rectum* stop. Everything else is solid; this is a one-file fix.

---

### Blocking findings

**B1 — Divergent canonicalizer ⇒ wrong CIDs.**
`implementations/python/provekit-lift-python-source/src/provekit_lift_python_source/canonical.py:11-18`

```python
def canonical_json_bytes(value: Any) -> bytes:
    encoded = json.dumps(value, sort_keys=True, separators=(",", ":"), ensure_ascii=False)
    return encoded.encode("utf-8")
```

`json.dumps` emits the RFC-8785 *named short escapes* (`\n`, `\t`, `\r`, `\b`, `\f`) for `U+0008..U+000D`. The protocol canonicalizer does **not** — `protocol/specs/2026-04-30-canonicalization-grammar.md` pass 7 and every impl (`implementations/rust/provekit-canonicalizer/src/jcs.rs:78-95`, the C++ peer, and `implementations/python/provekit-lift-py-tests/src/provekit_lift_py_tests/canonicalizer.py:151-167`) escape `U+0000..U+001F` as `\u00XX`, lowercase hex, *no short forms*. Reproduced in this worktree:

```
new canonical bytes : {"source":"def f():\n    return 1\n","x":2}
canon encode_jcs    : {"source":"def f():
    return 1
","x":2}
new cid : blake3-512:ad8dc857...   canon: blake3-512:e46fdb30...   → DIVERGENT
```

`_EffectSet.add_opaque_loop` (`lifter.py:67-72`) computes `loopCid = cid_of_json(loop_term)`, and every `python:while`/`python:for` body that contains a string literal with `\n` (or whose source bytes flow in) gets a `loopCid` the Rust verifier / cross-kit conformance / any `LoopInvariantMemento` author will never reproduce. The agent's prompt said "if there genuinely isn't one [a Python canonicalizer], you may implement a minimal JCS-canonicalizer" — but there *is* one, in the sibling `provekit-lift-py-tests` package, and it's the load-bearing cross-language conformance artifact. **Fix:** delete `canonical_json_bytes` + `_blake3_512_with_b3sum`; import `encode_jcs` + `blake3_512_of` from `provekit_lift_py_tests.canonicalizer` and feed it a `Value` tree (`vobj`/`varr`/`vstr`/`vint`/`vbool`/`vnull`), or add a thin `json-dict → Value` adapter. (`canonical.py` already *tries* to import `blake3_512_of` from there for the hash half — finish the job and do the JCS half the same way.)

**B2 — `pyproject.toml` declares a `blake3` dep nothing installs on the manifest path; no first-class dependency on the in-repo canonicalizer.**
`implementations/python/provekit-lift-python-source/pyproject.toml:10-12` lists `blake3>=0.3.0`, but the lift-plugin manifest spawns `python3 -m provekit_lift_python_source --rpc` with no install step. If `python3` lacks `blake3` *and* there's no `b3sum` on PATH *and* `provekit_lift_py_tests` isn't importable, `blake3_512_of` raises `RuntimeError` mid-lift. Fixing B1 (depend on `provekit-lift-py-tests`, which `requires` `blake3`) subsumes this; otherwise make the dependency real.

---

### Non-blocking

- **`op_call` spec vs. emitter shape mismatch.** `lifter.py:392` emits `ctor("python:call", str_const(callee), *[expr(arg)…])` — a *variadic* term (`callee` + N positional args). `menagerie/python-language-signature/specs/op_call.spec.json` declares `formals: ["callee", "args"]`, `arity_shape` named 2-slot with `args: {shape:{kind:"set"}}`. The `notes` field admits the variadic encoding, but `arity_shape` still says 2 slots, and `kind:"set"` is wrong for ordered call arguments. Pick one: model `args` as an explicit `ListOfValue` term, or make `arity_shape` describe the variadic spread. (Lower bar than #590 here — its `op_call` is also loose, but worth not propagating.)
- **`compile()` is not in the protocol spec.** `protocol/specs/2026-04-30-lift-plugin-protocol.md` defines only `initialize`/`lift`/`shutdown`. `rpc.py:54-57` adds `compile` — matches #590's (also-undocumented) extension, so consistent, but the protocol spec should eventually grow a `compile` method or this stays non-normative.
- **`initialize_result` carries an extra `dialect` key** (`rpc.py:21`); the CDDL `initialize-result` is closed (no `* text => json-value`). Matches #590. Cosmetic.
- **Round-trip test is shallow.** `test_compile_lift_roundtrip_ir_document_is_byte_identical` works because `compile_ir_document` short-circuits to the verbatim `source-unit` bytes (`compiler.py:9-12`) — so `lift(compile(lift(s)))` is `lift(s)` by construction whenever a `source-unit` is present (always). The `ast.unparse` path *is* exercised, but only by `test_compile_function_contract_without_source_unit_uses_ast_unparse`, which just substring-checks `def f(x):` — it never re-lifts the unparsed source and compares IR. Consider a test that strips the `source-unit`, `ast.unparse`s, re-lifts, and compares the *function* contract terms.
- `_record_write_if_nonlocal` is a no-op for `ast.Name` targets — correct (a local-var assignment isn't a `writes` effect), just non-obvious.

---

### Conformance checklist

| # | Item | Status |
|---|------|--------|
| 1 | `python:` prefix on every op; no `binop`/`unknown`/`skip` catch-all | ✅ enforced in `ir.ctor()` (`ir.py:46-49`) + `language_signature_python.spec.json`; no `op_unknown`/`op_binop` |
| 2 | `arity_shape` on every op, sane; short-circuit `and`/`or` mark rhs `unevaluated`; `seq` Positional; `assign` Named target/value | ✅ verified in `test_catalog.py`; *richer* than #590 (#590's `op_and`/`op_source-unit` don't mark `unevaluated`) |
| 3 | `python:source-unit(bytes, operational_term)` op + emitted with real source bytes | ✅ `op_source-unit.spec.json`; `ir.source_unit_contract` (`ir.py:99-121`); `lift_source` always emits it |
| 4 | Refuse-not-fake: loud `Refusal{kind,function,line,reason}` for comprehensions/generators/async/await/decorators/with/try/walrus/lambda/match/global/nonlocal/`*args`/`**kwargs`; chained comparisons | ⚠️ mostly ✅ via `_UnsupportedSyntax` + `_contains_refused_control` (`lifter.py:465-477`) + per-function guards (`lifter.py:419-433`). **`with`/`try`/`lambda`/`match` are caught by the generic statement/expr fallthrough** (`statement` raises on unhandled `ast.stmt`, `expr` on unhandled `ast.expr`) — they refuse, but via a generic "unhandled statement kind" message, not a named refusal. **Chained comparisons** (`a < b < c`): explicitly refused, `lifter.py:344-348` — correct conservative choice, but a 3-clause chain is *not* modeled (test `test_refuses_unhandled_syntax_without_unknown_ops` confirms refusal). Acceptable for a draft. |
| 5 | `fnName` = `<module-path>.<qualname>` from AST nesting (`Class.method`, `outer.<locals>.inner`) | ✅ `_DefinitionCollector` + `_qualname` (`lifter.py:181-216, 528-536`); module path sanitized (`_module_path`/`_clean_identifier_part`) |
| 6 | Effects on canonical wire shapes, sorted by `sort_key()` incl. secondary sort; no `alloc`/resolved-`call` | ✅ `_EffectSet` shapes match `libprovekit/src/compose.rs` `to_value` (`{"kind":"reads","target":…}`, `{"kind":"opaque_loop","loopCid":"blake3-512:…"}`, `{"kind":"unresolved_call","name":…}`, `io`, `panics`); `_effect_sort_key` matches `compose.rs::sort_key` ordering (`0:reads…6:opaque_loop`) with target/name/loopCid as secondary key; no `alloc`, no resolved `call` |
| 7 | Round-trip `lift(compile(lift(src)))` byte/CID-identical; 8 tests | ⚠️ holds, but trivially (source-unit short-circuit); `ast.unparse` path under-tested — see Non-blocking. 8 tests: lift→source-unit+python ops; refuse-without-unknown; effects sorted + loopCid blake3-512; roundtrip byte-identical; compile-without-source-unit uses unparse; rpc initialize shape; + 2 catalog tests (draft version & expected ops, arity shapes). |
| 8 | Determinism: `json.dumps(sort_keys=True, separators=(",",":"))`; no dict-order leak; no timestamps | ✅ output is deterministic — *but* it's the **wrong** canonical form (B1). RPC framing uses `json.dumps(separators=(",",":"))` (`rpc.py:104`). No timestamps. |
| 9 | CID-correctness — reuse existing canonicalizer? | ❌ **Rolled its own** (`json.dumps`), and it's **not** byte-identical to `provekit-canonicalizer` (B1). Reuses only the BLAKE3 half (and only opportunistically). |
| 10 | Wiring: existing `python/manifest.toml` not repointed; new `python-source` surface; `pyproject.toml`; manifest `command` runs the lifter; `version 0.1.0-draft`; no `.provekit/ci/accepted/*` changes; `menagerie/manifest.yaml` entry correct | ✅ existing `.provekit/lift/python/manifest.toml` untouched; new `.provekit/lift/python-source/manifest.toml` (`command = ["python3","-m","provekit_lift_python_source","--rpc"]`, `working_dir = "provekit-lift-python-source/src"` — resolves correctly relative to the `implementations/python/` workspace, `-m` finds the package from `src/`); `python-source` added to `KNOWN_SURFACES` (`project_config.rs:216`); `version 0.1.0-draft` in manifest+rpc; no `.provekit/ci/accepted/*` or self-contract-pin changes; `menagerie/manifest.yaml` entry mirrors the existing `*-language-signature` pattern (rust/c11/jvm/aarch64 sigs are *not* in `manifest.yaml` today — so adding `python-language-signature` is a slight asymmetry, but `csharp-language-signature` from #590 also isn't there; non-issue). `Makefile` target added. Caveat: `version` in `pyproject.toml` is `0.1.0` (not `-draft`) — fine, package version ≠ kit draft version. |
| 11 | Existing Python kit still works — `pytest implementations/python/provekit-lift-py-tests` | ✅ 155 passed, 1 skipped. New package: 8 passed. `generate_assets.py --check` passes. |
| 12 | `--rpc` server `initialize`/`lift`/`compile`/`shutdown` shapes match the protocol spec | ⚠️ `initialize` has extra `dialect` key (closed CDDL — minor); `lift` result adds `callEdges`/`opacityReport`/`refusals` to `ir-document` (spec's `ir-document` is `{kind, ir, ?diagnostics}` — but matches #590 and the de-facto richer shape); `compile`/`shutdown` shapes fine; `compile` itself is off-spec but consistent with #590. |

---

`canonical.py` — confirm whether the CIDs it emits (`loopCid`) match the Rust `provekit-canonicalizer` for terms containing newlines. They don't (reproduced above). Fix that and the PR is essentially done.

### Comparison to #590's bar
This PR **meets or exceeds** #590 on op-spec richness (explicit `unevaluated` slots #590 lacks), refusal discipline (decorators/`*args`/`**kwargs`/walrus/yield/await all named refusals), and effect-shape conformance. It **falls below** #590 on exactly one axis — and it's the load-bearing one: #590's C# kit uses the established canonicalizer plumbing; this one substitutes `json.dumps` and silently diverges on control chars. Bring the canonicalizer in line with the rest of the repo (reuse `provekit_lift_py_tests.canonicalizer`) and this is a MERGE.

PR #595, opened 2026-05-11
feat(ts): add TypeScript source lift kit (self-hosted) + ts-language-signature

TSavo:
Self-hosted TypeScript source-language lift kit + `menagerie/ts-language-signature/`. Mirrors PR #590 (the C# source lifter).

- TypeScript lifter via the TypeScript compiler API → `function-contract` mementos over `ts:`-namespaced ops
- per-file `ts:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test
- new `ts-source` lift surface (existing `ts` kit surface manifest untouched — the lesson from #590); `version 0.1.0-draft`
- vitest 6/6 + full suite 785 tests pass; tsc --noEmit pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,8 @@ help:
 	@echo ""
 	@echo "Maintenance:"
 	@echo "  make clean          remove build artifacts"
+	@echo "  make mint-typescript-language-signature"
+	@echo "                       mint the draft TypeScript source language-signature catalog"
 	@echo ""
 	@echo "Pinned CIDs (catalog v1.6.2):"
 	@echo "  catalog: $(CATALOG_CID)"
@@ -510,6 +512,10 @@ catalog-verify:
 .PHONY: c11-cursorkind-check
 c11-cursorkind-check:
 	python3 tools/generate-c11-from-cursorkind.py --check
+
+.PHONY: mint-typescript-language-signature
+mint-typescript-language-signature:
+	menagerie/typescript-language-signature/mint.sh
 
 .PHONY: protocol-verify
 protocol-verify: build-rust
--- a/implementations/rust/provekit-cli/src/project_config.rs
+++ b/implementations/rust/provekit-cli/src/project_config.rs
@@ -205,6 +205,7 @@ pub const KNOWN_SURFACES: &[&str] = &[
     "rust-proptest",
     "rust-quickcheck",
     "ts-zod",
+    "typescript-source",
     "ts-class-validator",
     "ts-fast-check",
     "ts-jsdoc",
@@ -312,6 +313,7 @@ mod tests {
     fn known_lists_include_anchor_entries() {
         assert!(KNOWN_SURFACES.contains(&"default"));
         assert!(KNOWN_SURFACES.contains(&"ts-zod"));
+        assert!(KNOWN_SURFACES.contains(&"typescript-source"));
         assert!(KNOWN_SURFACES.contains(&"clr-bytecode"));
         assert!(KNOWN_AGENTS.contains(&"stub"));
         assert!(KNOWN_AGENTS.contains(&"claude-code"));
--- a/implementations/typescript/.provekit/lift/typescript-source/manifest.toml
+++ b/implementations/typescript/.provekit/lift/typescript-source/manifest.toml
@@ -1,0 +1,10 @@
+name = "typescript-source"
+version = "0.1.0-draft"
+protocol_version = "provekit-lift/1"
+command = ["npx", "tsx", "src/lift/typescript-source/bin.ts", "--rpc"]
+working_dir = "."
+
+[capabilities]
+authoring_surfaces = ["typescript-source"]
+ir_version = "v1.1.0"
+emits_signed_mementos = false
--- a/implementations/typescript/src/lift/typescript-source/bin.ts
+++ b/implementations/typescript/src/lift/typescript-source/bin.ts
@@ -1,0 +1,118 @@
+#!/usr/bin/env node
+import readline from "node:readline";
+
+import {
+  compileTypeScriptSourceIr,
+  liftTypeScriptSourcePaths,
+  type FunctionContractMemento,
+} from "./index.js";
+
+const DIALECT = "typescript-source";
+const VERSION = "0.1.0-draft";
+
+interface JsonRpcRequest {
+  jsonrpc?: string;
+  id?: unknown;
+  method?: string;
+  params?: Record<string, unknown>;
+}
+
+export function main(argv: string[] = process.argv.slice(2)): void {
+  if (!argv.includes("--rpc")) {
+    process.stderr.write("usage: provekit-lift-typescript-source --rpc\n");
+    process.exit(1);
+  }
+  runRpcMode();
+}
+
+function runRpcMode(): void {
+  const rl = readline.createInterface({ input: process.stdin, output: process.stdout, terminal: false });
+  rl.on("line", (line) => {
+    if (line.trim() === "") return;
+    let request: JsonRpcRequest;
+    try {
+      request = JSON.parse(line) as JsonRpcRequest;
+    } catch (error) {
+      write(errorResponse(null, -32700, `PARSE_ERROR: ${(error as Error).message}`));
+      return;
+    }
+    try {
+      const response = dispatch(request);
+      if (response) write(response);
+    } catch (error) {
+      write(errorResponse(request.id ?? null, -32603, (error as Error).message));
+    }
+  });
+}
+
+function dispatch(request: JsonRpcRequest): Record<string, unknown> | null {
+  switch (request.method) {
+    case "initialize":
+      return success(request.id, {
+        name: "provekit-lift-typescript-source",
+        version: VERSION,
+        protocol_version: "provekit-lift/1",
+        capabilities: {
+          authoring_surfaces: [DIALECT],
+          ir_version: "v1.1.0",
+          emits_signed_mementos: false,
+        },
+      });
+    case "lift":
+      return liftRpc(request);
+    case "compile":
+      return compileRpc(request);
+    case "shutdown":
+      return success(request.id, null);
+    default:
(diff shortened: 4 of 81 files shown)


TSavo:
## Review: PR #595 — `feat(ts): TypeScript source lift kit (self-hosted, TS compiler API)`

**Verdict: MERGE-WITH-NITS** — conformant, stricter-than-#590 (no `ts:skip`/`ts:unknown` catch-all, asserted by test), self-canonicalizing on the existing `claimEnvelope/canonicalize.ts` + `canonicalizer/hash.ts` (not a reinvention); the one substantive issue (round-trip test is tautological) is a test-strength gap, not a correctness bug.

### What I verified (evidence)
- `node_modules/.bin/tsc --noEmit -p tsconfig.json` → exit 0 (the new `src/lift/typescript-source/` dir is covered by root `tsconfig.json` `include`).
- `vitest run implementations/typescript` → **44 files, 785 pass + 4 todo** (matches the "785 still pass" claim) + the new file → **6 pass**.
- RPC smoke under `node --import tsx src/lift/typescript-source/bin.ts --rpc`: `initialize` returns `{name, version:"0.1.0-draft", protocol_version:"provekit-lift/1", capabilities:{authoring_surfaces:["typescript-source"], ir_version:"v1.1.0", emits_signed_mementos:false}}`; `lift` on `index.ts` itself emits **3 ir entries + 69 refusals** (loud, structured `{kind,function,line,reason}` — refuses async/generator/decorated/generic/rest/default/destructured, arrows, `await`, optional chaining, template literals, `as`/`satisfies`, function expressions); `shutdown` → `null`. ✅
- Canonicalizer: `index.ts:5-6` imports `canonicalJsonString` from `claimEnvelope/canonicalize.ts` (RFC-8785-ish sorted-keys/no-whitespace/UTF-8, the protocol-pinned one) and `computeCid` from `canonicalizer/hash.ts` (blake3-512, no truncation). **No rolled-own JCS/BLAKE3.** ✅ *Supra omnia, rectum* clear.
- `ts:` prefix on **every** op spec; no `ts:binop`/`ts:unknown`/`ts:skip` catch-all. The `default:` arms in the lifter's `emitStatement`/`emitExpression`/`binaryOperatorName` all `throw UnsupportedSyntaxError` → refusal. ✅
- `arity_shape` present on all 46 op specs and sane: binops `named` lhs/rhs; `ts:and`/`ts:or`/`ts:nullish` mark `rhs` `evaluation:"unevaluated"`; `ts:ite` marks `then_expr`/`else_expr` unevaluated; `ts:seq`/`ts:args` `positional` `variadic`; `ts:assign` `named` target/value; `ts:if`/`ts:while`/`ts:for` mark branch/body unevaluated; `ts:source-unit` `named` `{bytes(unevaluated,literal), operational_term}`. Mirrors the c11/csharp op-spec shape. ✅
- `ts:source-unit(bytes, operational_term)` op present (`op_source-unit.spec.json`) **and** emitted by the lifter carrying `sourceFile.getFullText()` (`index.ts:210`, `sourceUnitContract` `:650`). Test `index.test.ts:33-36` confirms. ✅
- `fnName = <module-path>:<qualified-name>` from AST prefix path (`[...prefix, shortName].join(".")`, `:285`). Overloads can't collide: overload *signatures* are bodyless → refused (`:237-238`); only the implementation (with body) is lifted. Class vs namespace funcs are disambiguated by prefix (`N.A.m` vs `N.m`). Built syntactically, **not** via `TypeChecker` symbol — adequate for the supported slice but see nit (4).
- Effects on canonical wire shapes: `reads`/`writes` with `target`, `io`, `panics`, `unresolved_call` with `name`, `opaque_loop` with `loopCid:"blake3-512:..."`. **No `alloc`, no resolved-`call`** (resolved calls in `knownCallables` emit no effect — `:517-518`). `effectSortKey` uses prefixes `0:reads 1:writes 2:io 4:panics 5:unresolved_call 6:opaque_loop` — numeric prefixes match Rust `libprovekit/src/compose.rs:241-247` (`Effect::sort_key`); the `5:unresolved_call:name` vs Rust's `5:unresolved:name` differs only in the throwaway sort string, not the emitted value or array order. Test `index.test.ts:69-97` asserts the exact array. ✅
- Determinism: emitted JSON via the shared canonicalizer (sorted keys, no whitespace); effects sorted before emit; no `Map`-iteration leak (effects collected in a `Map` keyed by sort-key, then `[...values()].sort(effectSortKey)`); no timestamps. ✅
- Wiring: existing `implementations/typescript/.provekit/lift/typescript/manifest.toml` **not** repointed (still `npx tsx src/lift/bin/main.ts --rpc`) ✅. New `typescript-source` surface manifest added; `typescript-source` added to `KNOWN_SURFACES` (`project_config.rs:208`) + unit-test assertion. `version 0.1.0-draft`. No `.provekit/ci/accepted/*` / self-contract-pin changes (expected for draft). ✅
- The `compile` RPC method is non-standard (the protocol spec only defines `initialize`/`lift`/`shutdown`) — but **#590's C# lifter does the same** (`RpcServer.cs`/`CsharpCompiler.cs`, `initialize, lift, compile, shutdown`), so this is an established extension, not a #595 regression.

### One thing to look at first
**The round-trip test is tautological.** `compileTypeScriptSourceIr` (`index.ts:165-189`) — when a `<source-unit>` is present, returns its stored raw source-bytes string (`:168-177`). The lifter *always* prepends a `<source-unit>` whenever `lifted.length > 0` (`:209-211`). So `lift(compile(lift(src)))` exercises only the byte-copy path; the `compileFunctionContract`/`ts.createPrinter` AST-emit path (`:179-189`, `:901-1022`) is **dead code under normal use** (only reachable if a caller hands `compile` an IR doc with no source-unit). The test name "round-trips lift(compile(lift(src))) with byte-identical canonical IR" overpromises. Not a bug — the lossless source-bytes wrapper is the intentional design — but the test isn't proving the AST printer is byte-stable. Add a hand-crafted-IR-without-source-unit round-trip case, or rename the test. **Non-blocking.**

### Blocking findings
None.

### Non-blocking nits
1. Tautological round-trip test (above) — `implementations/typescript/src/lift/typescript-source/index.test.ts:118-135`.
2. `menagerie/manifest.yaml:32-36` registers `typescript-language-signature` as a runnable destination — **inconsistent with established precedent**: `c11-language-signature`, `rust-language-signature`, `aarch64-language-signature`, `jvm-bytecode-language-signature`, and the just-merged `csharp-language-signature` are *not* registered there. Arguably an improvement; if kept, file a follow-up to register the others (or drop this entry to match).
3. `menagerie/typescript-language-signature/mint.sh` uses `find` to enumerate specs; `c11`/`rust` mint.sh do not (they enumerate explicitly / via a generator). Style/portability nit.
4. `hasUnsupportedFunctionShape` collapses six distinct refusal reasons into one message string (`index.ts:288`) — a caller can't tell *which* unsupported shape triggered it. Minor UX; consider splitting.
5. Draft kit not yet minted: no `cids.tsv` / `catalog/` / `component-cids.json` checked in (c11/rust have them), README is a 5-line stub. Expected for `0.1.0-draft`; file a follow-up to mint + pin (and add a `make` smoke or CICP leg) before it leaves draft.
6. `EPERM` note: the *existing* `typescript` manifest uses `npx tsx ...`; the new `typescript-source` manifest also uses `["npx","tsx","src/lift/typescript-source/bin.ts","--rpc"]`. The agent's `node --import tsx` form worked in my sandbox; `npx tsx` is the established form and presumably works in CI — but if the `EPERM` recurs in the runner, this manifest will inherit it. Worth confirming the CICP run is green before merge.

### Conformance checklist
| # | Item | Status |
|---|---|---|
| 1 | `ts:` prefix on every op; no `ts:binop`/`ts:unknown` catch-all | ✅ (test asserts no `ts:skip`/`ts:unknown`) |
| 2 | `arity_shape` on every op & sane (short-circuit, ite, seq, assign, branches) | ✅ |
| 3 | `ts:source-unit(bytes, operational_term)` op present & emitted with real bytes | ✅ |
| 4 | Refuse-not-fake: loud `Refusal` for unhandled SyntaxKind (arrows, async/await, generators, destructuring, spread, defaults, try/catch, switch, `?.`, template literals, `as`/`satisfies`, decorators, generics, JSX) | ✅ (verified live; switch/try-catch/JSX hit the `default` throw) |
| 5 | `fnName` = `<module-path>:<qualified-name>`; no collision | ✅ (syntactic path; overload sigs bodyless→refused) — TypeChecker-symbol not used (nit 4) |
| 6 | Effects on canonical wire shapes, sorted by `sort_key()`; no `alloc`/resolved-`call` | ✅ |
| 7 | Round-trip `lift(compile(lift(src)))` byte/CID-identity | ⚠️ test passes but exercises only the byte-copy path; AST-printer path untested (nit 1) |
| 8 | Determinism (stable key order, no Map leak, no timestamps) | ✅ |
| 9 | CID-correctness: reuses the existing TS canonicalizer / BLAKE3 | ✅ |
| 10 | Wiring: `typescript` manifest not repointed; new `typescript-source` surface; KNOWN_SURFACES; no pin changes | ✅ |
| 11 | Existing TS kit still builds/tests (`tsc --noEmit` + 785 tests) | ✅ |
| 12 | RPC `initialize`/`lift`/`compile`/`shutdown` shapes match protocol | ✅ (`compile` is a non-standard extension also present in #590) |

### Comparison to #590's bar
Meets or exceeds it. Same op-spec/effsig/sort/eq layout, same RPC surface (incl. the undocumented `compile`), same `0.1.0-draft` + own-surface manifest pattern, same "don't repoint the kit surface" fix. **Stricter** in one respect: #590's C# lifter still ships a `csharp:skip` fallback op; #595 has none and the test actively forbids `ts:skip`/`ts:unknown` in lift output. Same draft-quality gaps as #590 (kit not minted/pinned). Net: ship it; file the follow-ups (round-trip test, manifest.yaml consistency, mint+pin).

— review by Claude (Opus 4.7, 1M ctx)

PR #596, opened 2026-05-11
feat(zig): add Zig source lift kit (self-hosted) + zig-language-signature

TSavo:
Self-hosted Zig source-language lift kit + `menagerie/zig-language-signature/`. Mirrors PR #590 (the C# source lifter).

- Zig lifter via `std.zig.Ast` → `function-contract` mementos over `zig:`-namespaced ops
- per-file `zig:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test
- `arity_shape` on every op spec; `Refusal`s (not fakes) for comptime/error-unions/try-catch/defer/async/inline-for/switch/`@`-builtins; effects on the canonical `Effect` wire shapes, sorted; BLAKE3/JCS loop CIDs
- new `zig-source` lift surface (existing `zig` kit surface manifest `provekit-lift-zig` untouched — the lesson from #590); `version 0.1.0-draft`
- `make test-zig` + `zig build test` pass


TSavo:
## Review: PR #596 — Zig source lift kit (self-hosted)

**Verdict: MERGE-WITH-NITS** — the kit is structurally sound, reuses the existing Zig canonicalizer/BLAKE3 (not a re-roll), refuses unhandled AST node tags loudly rather than faking ops, builds and tests clean; the nits are draft-grade gaps (string-escape fidelity, lift/compile surface asymmetry) that should be tracked before any non-draft promotion.

**CID-correctness (item 9): confirmed good.** `lift.zig` imports `provekit-ir` and routes every canonical-bytes/CID call through `provekit.jcsStringify` / `provekit.jcsHash` (`addOpaqueLoop`, `canonicalTermBytes`, the tests). It does **not** roll its own JCS or BLAKE3 — `loopCid` values are `blake3-512:...` from the same module the rest of the Zig kit uses. `zig build test` for `provekit-ir`, `provekit-lift-zig`, and `provekit-lift-zig-source` all pass on zig 0.16.0. No *Supra omnia, rectum* stop.

### Blocking findings
None.

### Non-blocking findings (track before promoting past `0.1.0-draft`)

1. **String literals are lifted without unescaping** — `lift.zig:495-499` (`emitString`) strips the outer `"`/`"` but leaves the inner bytes verbatim. So source `"oops\n"` becomes `Str("oops\n")` *with a literal backslash-n*, i.e. the canonical term — and therefore its CID — misrepresents the runtime string value. No test exercises an escaped string literal, so the round-trip test doesn't catch it. Also asymmetric: `compileContract` re-escapes on the way out (`appendEscapedString`, line 951), so even a plain `"\\"`-containing literal won't round-trip. Fix: unescape on lift (and unescape-not-re-escape on compile), or refuse any string literal containing `\`. Concrete instance of the exact failure shape that contaminates downstream predicates.

2. **lift surface is wider than the compile surface** — the lifter emits `zig:for`, `zig:break`, `zig:continue` (and `zig:panic`), but `emitTermAsStmt`/`emitTermAsExpr` (`lift.zig:797-927`) handle none of them; they fall through to the literal `"unreachable"` default. `compileContract` on any function with a `for` loop or `break`/`continue` produces broken Zig. The round-trip test (`lift.zig:1107`) only uses a single `add` example, which camouflages this. Either widen `compileContract` or have it refuse terms it can't render.

3. **`options.layer` / `identify-only` ignored** — `main.zig:62` substring-matches `"lift"` and always runs the proof-producing path; per `protocol/specs/2026-04-30-lift-plugin-protocol.md` a plugin that doesn't implement `identify-only` MAY return error `1006`. Note this is the **same pattern the existing `provekit-lift-zig` already uses** (`provekit-lift-zig/src/main.zig`), so it's a kit-wide gap, not a regression — flagging for the eventual sweep, not for this PR.

4. **`ir-document` carries extra fields** — the `lift` response emits `callEdges`/`opacityReport`/`refusals` alongside `kind`/`ir`; the `ir-document` grammar in the spec doesn't list a `* text => json-value` wildcard (unlike `identity-document`). The Rust CLI extracts fields by name so it's harmless in practice, and `provekit-lift-zig` does the analogous thing (`implications`). Cosmetic.

5. **`<source-unit:input.zig>` `fnName` collision risk in `source`-mode RPC** — when lifting via `{"source":...}` with no `path`, `path` defaults to `"input.zig"`, so two such calls in one session produce colliding `fnName`s. Same shape as the other lifters; the angle-bracket form keeps it distinct from real functions; not introduced here.

### Conformance checklist

| # | Item | Status | Note |
|---|---|---|---|
| 1 | `zig:` prefix on every op, no `binop`/`unknown` catch-all | PASS | All 41 `op_*.spec.json` `fn_name`s are `zig:*`; lifter's `emitExpr` else-branch is `refuseFmt(... "unhandled Zig AST node tag")`, not a synthesized op |
| 2 | `arity_shape` present & sane on every op | PASS | binops `named` lhs/rhs; `zig:seq`/`zig:unit`/`zig:unreachable` `positional`; `zig:assign` `named` target/value; `zig:and`/`zig:or` mark `rhs` `evaluation:"unevaluated"`; mirrors the c11/csharp op-spec shape |
| 3 | `zig:source-unit(bytes, operational_term)` present AND emitted with real bytes | PASS | `op_source-unit.spec.json` has the right shape; `sourceUnitContract` (`lift.zig:611`) emits `zig:source-unit(Str(self.source), operational)`; verified via RPC the bytes are the verbatim source |
| 4 | Refuse-not-fake for unhandled AST tags (comptime/error-union/try/catch/errdefer/defer/suspend/resume/inline loops/switch/optional/`@`-builtins/anon structs/generics) | PASS | All explicitly enumerated in `emitStmt`/`emitExpr`/`sortFromType`/`emitFunction` with structured `Refusal{kind,function,line,reason}`; AST-only, no `Sema` import; `comptime` params and `anytype` refused as `unsupported-generic` |
| 5 | `fnName` provably unique | PASS (with caveat 5 above) | `qualifiedName` = `path` + namespace chain + name; no overloading in Zig; parse-error path returns empty `declarations`, no colliding fallback |
| 6 | Effects on canonical wire shapes, sorted by `sort_key()` | PASS | `Effect` union = reads/writes/io/unsafe/panics/unresolved_call/opaque_loop with rank 0-6; `Effect.lessThan` does primary rank then secondary (`std.mem.lessThan` on target/name/cid); sorted via `std.mem.sort` before emission; no `alloc`, no resolved-`call`; `opaque_loop.loopCid` is `blake3-512:...`; `unsafe` is defined in the signature but not yet emitted by the lifter (acceptable for draft) |
| 7 | Round-trip `lift(compile(lift(src)))` byte/CID-identity, drives the real compiler | PASS (narrow) | `lift.zig:1107` checks `canonicalTermBytes(bodyTerm())` equality through `compileContract` → real `Ast.parse`; `compileContract` renders actual Zig from the term (`return (x + y);`), no `zig:source-unit` byte-copy fast-path → the Go/Python/TS tautology nit does **not** apply. But the test covers only `add`; see non-blocking #2 |
| 8 | Determinism — stable key order, no HashMap iteration leak, no timestamps | PASS | `ArrayList` throughout; custom `jsonStringify` impls emit keys alphabetically; canonical bytes via `provekit.jcsStringify`; no `AutoHashMap`/`StringHashMap`; `generate_assets.py` re-run reproduces all 77 spec files byte-identically (git clean) |
| 9 | CID-correctness — reuses existing canonicalizer + BLAKE3 | PASS | See "the one thing to look at first" |
| 10 | Wiring — existing `zig` manifest not repointed; new `zig-source` surface manifest; `version 0.1.0-draft`; no `.provekit/ci/accepted` changes; no `zig-language-signature` in `menagerie/manifest.yaml` | PASS | `git diff` shows existing `implementations/zig/.provekit/lift/zig/manifest.toml` untouched; new `.provekit/lift/zig-source/manifest.toml` with `command = ["zig","build","run","--","--rpc"]`, `working_dir = "provekit-lift-zig-source"`, `version 0.1.0-draft`; build.zig produces a `provekit-lift-zig-source` executable + a `run` step; zero `ci/`/`accepted/`/self-contract changes; `grep language-signature menagerie/manifest.yaml` → no match (clean); `KNOWN_SURFACES` in `provekit-cli/src/project_config.rs` gains `"zig-source"` next to `clr-bytecode` (same pattern as #590) |
| 11 | Existing Zig kit still builds | PASS | `provekit-lift-zig` `zig build test` and `provekit-ir` `zig build test` both green; Makefile's `test-zig`/`build-zig` extended (not replaced) to add `provekit-lift-zig-source` |
| 12 | `--rpc` `initialize`/`lift`/`shutdown` match the lift-plugin protocol | PASS (with #4) | `initialize` returns `name`/`version`/`protocol_version: "provekit-lift/1"`/`capabilities{authoring_surfaces,emits_signed_mementos,ir_version}`; `lift` returns `kind:"ir-document"` + `ir` array; `shutdown` returns `result:null`; no `compile` RPC method (correct — the protocol has none; `compileContract` is internal-only for the round-trip test) |

### Comparison to #590's bar
Meets it. Same shape as the CLR/csharp kits: separate module (correctly *not* folded into the annotation-adapter `provekit-lift-zig`), language-prefixed op algebra with a generated menagerie signature, refuse-don't-fake on anything outside the modeled subset, `0.1.0-draft` version, reuses the kit's own canonicalizer. The string-escape gap (#1) is the one item I'd want closed before this surface graduates from draft; #590's reviewers would have weighted the canonicalizer-reuse question highest, and that one is clean.

— review by Claude (Opus 4.7, 1M ctx), worktree build/test on zig 0.16.0

PR #598, opened 2026-05-11
feat(ruby): add Ruby source lift kit (self-hosted) + ruby-language-signature

TSavo:
Self-hosted Ruby source-language lift kit + `menagerie/ruby-language-signature/`. Mirrors PR #590 (the C# source lifter) and the merged Go/Python/TS/Zig lifters.

- Ruby lifter via stdlib Ripper → `function-contract` mementos over `ruby:`-namespaced ops
- per-file `ruby:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test driving the real IR→source path (not the source-unit byte-copy fast-path)
- new `ruby-source` lift surface (existing `ruby` kit surface manifest untouched — the #590 lesson); `version 0.1.0-draft`; no `menagerie/manifest.yaml` entry (the recurring nit — `*-language-signature` catalogs don't belong in that destinations list)
- rake test 113 runs/284 assertions + make test-ruby pass; signature --check pass

## Release Notes

* **New Features**
  * Added Ruby language support to Provekit, enabling lifting of Ruby source files into IR format
  * Comprehensive Ruby operation catalog now available, including arithmetic, comparison, control flow, method calls, and variable access
  * Ruby effect tracking for reads, writes, IO operations, exception handling, and unresolved calls
  * CLI support for `ruby-source` surface configuration

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/598)


TSavo:
## Review — PR #598 `feat(ruby): add Ruby source lift kit (self-hosted)`

**Verdict: MERGE-WITH-NITS** — self-hosted Ruby source lifter is faithful to the #590 bar; reuses the gem's real canonicalizer/BLAKE3, refuses-not-fakes, round-trip test drives the real IR→Ruby compiler; only cosmetic/consistency nits remain.

Reviewed `git diff origin/main...HEAD` on `b8c38260`, ran `bundle exec rake test` on Homebrew Ruby 4.0.3 (113 runs / 284 assertions / 0 failures), `python3 menagerie/ruby-language-signature/generate_assets.py --check` (clean), and a live `--rpc` `initialize`/`lift`/`compile`/`shutdown` round-trip + `compile_ir_document` byte-identity check. CodeRabbit's review was still "processing" at review time.

### Blocking findings
None.

### Non-blocking nits
1. **`effect_sort_key` prefix diverges from siblings.** `lib/provekit/lift/ruby_source.rb:705` returns `"5:unresolved_call:#{effect["name"]}"`; the Rust canonical (`implementations/rust/libprovekit/src/compose.rs` `Effect::sort_key`) and the Python source lifter (`implementations/python/provekit-lift-python-source/src/provekit_lift_python_source/lifter.py:599`) both use `"5:unresolved:..."`. Ordering-equivalent (sort keys are internal-only, never serialized; I verified `5:` < `6:opaque_loop` so no cross-kind collision), so harmless — but it should match the established pattern for the next reader's sake.
2. **`test-ruby` not wired into `test-all`.** `Makefile:689` `test-all` excludes `test-ruby` (and `ruby-language-signature --check`), so `make ci` won't exercise the new kit. Consistent with the existing `ruby` kit (also absent), and CI did run `rake test` here — but if Ruby is meant to be a first-class polyglot kit, add it. Not introduced by this PR; flagging per boy-scout.
3. `op_call`/`op_send` mark the variadic `args` slot `"shape": {"kind": "set"}` (`menagerie/ruby-language-signature/specs/op_call.spec.json`, `op_send.spec.json`) while the notes say "positional actual arguments" and the term encoding is order-significant. This is copied verbatim from #590's `menagerie/csharp-language-signature/specs/op_call.spec.json` — same (questionable) descriptor. Pre-existing; mirror is faithful. Ruby additionally marks `callee`/`method` as `evaluation:"unevaluated", slot_sort:"literal"` — a small improvement over #590.
4. `METAPROGRAMMING_CALLS` (`ruby_source.rb:511`) omits `instance_variable_get`/`instance_variable_set`/`const_get`/`const_set`. They fall through to `unresolved_call`, which is sound (conservative), so not a correctness gap — but a `define_method`-class refusal would be tighter.

### Conformance checklist
| # | Item | Result |
|---|---|---|
| 1 | `ruby:` prefix on every op; no `ruby:binop`/`ruby:unknown` catch-all | PASS — 41 op specs, all `ruby:*`; `ctor()` (`ruby_source.rb:746-751`) raises on `unknown`/`binop`/`skip`; no `ruby:skip` no-op present either |
| 2 | `arity_shape` on every op & sane (binops Named lhs/rhs; `ruby:seq` Positional; `ruby:assign` Named target/value; `and`/`or`/`ternary` mark dead slot `evaluation:"unevaluated"`) | PASS — 41/41 have `arity_shape`; `op_and`/`op_or` rhs `unevaluated`, `op_ternary` then/else `unevaluated`, `op_seq` `{kind:positional,arity:2}`, `op_assign` Named target/value; mirrors csharp op-spec shape |
| 3 | `ruby:source-unit(bytes, operational_term)` op present AND emitted carrying real source bytes | PASS — `op_source-unit.spec.json` Named slots `bytes`(unevaluated/literal)+`operational_term`; `source_unit_contract` (`ruby_source.rb:786-801`) emits `ctor("ruby:source-unit", str_const(source), operational_term)`; live RPC confirms `<source-unit:c.rb>` with `args[0]["value"] == source` |
| 4 | Refuse-not-fake: loud `Refusal{kind,function,line,reason}` for unhandled syntax (blocks/yield/metaprogramming/begin-rescue/case-when/splat/kwargs/super/`&.`/massign/class reopening/interpolation) | PASS — `Emitter#expr` else (`:326-328`) and statement paths raise `UnsupportedSyntax`; explicit refusals for blocks/lambda (`:318-319`), yield (`:320-321`), begin (`:247-248`), `&.` (`:311`,`:446`), splat/kwargs (`:419-421`), interpolation (`:385`,`:392`), metaprogramming via `METAPROGRAMMING_CALLS` (`:478-482`), class inheritance (`:122-124`), module/class reopening (`:128-130`), bodystmt rescue/ensure (`:851-853`); `case`/`super`/`massign` hit `expr` else → refused; refusal hash has exactly `kind`/`function`/`line`/`reason` (`:183-190`,`:816-822`); test `test_refuses_blocks_without_unknown_or_skip_terms` asserts no `ruby:unknown`/`ruby:skip` leaks |
| 5 | `fnName` provably unique | PASS — `<ns>#<name>` for instance, `<recv>.<name>` for singleton (`:142`,`:157`), built from AST nesting; `@seen_functions` set refuses duplicates (`:164-166`), `@seen_namespaces` refuses reopening; `class A::B` vs nested `class A;class B` converge to same scope and collide → refused, never silent |
| 6 | Effects on canonical wire shapes, sorted by `sort_key()` | PASS — `EffectSet` (`:40-79`) emits `{kind:"reads",target}`/`{kind:"writes",target}`/`{kind:"io"}`/`{kind:"panics"}`/`{kind:"unresolved_call",name}`/`{kind:"opaque_loop",loopCid:"blake3-512:…"}`, byte-identical to Rust `Effect::to_value()` (`libprovekit/src/compose.rs:166-235`); no `alloc`, no resolved `call`; `sorted` (`:70-72`) keys by `effect_sort_key` (numeric-prefix ordering matches Rust; nit #1 on the constant string); test asserts order `reads,reads,reads,writes,io,panics,unresolved_call,opaque_loop` and `loopCid =~ /\Ablake3-512:[0-9a-f]{128}\z/` |
| 7 | Round-trip `lift(compile(lift(src)))` byte/CID-identity, driving the real IR→Ruby compiler (not the source-unit byte-copy) | PASS — `test_compile_lift_roundtrip_body_term_is_byte_identical` (test:114-129) calls `compile_body_term(body,...)` → `compile_contract` → `Compiler#stmt_lines` (real path), then `lift_source(compiled)` and extracts `contract(relifted.ir,"<top>#f")["post"]["args"][1]` — the **function-contract** body term, not the `ruby:source-unit` byte copy — and asserts JCS-byte equality. `compile_ir_document` (`:671-676`) explicitly filters out `source_unit_contract?` items. I reproduced byte-identity manually for a `class C; def add(x); y=x+1; return y; end; end`. |
| 8 | Determinism: stable key order, no Hash-iteration leak, no timestamps | PASS — terms/effects built as ordered Ruby Hashes; `effects.sorted` deterministic (fully-discriminating keys); JCS via `Provekit::IR::Jcs.encode` sorts keys; `ruby_files` (`:982-999`) sorts; no time/rand |
| 9 | CID-correctness: reuses gem's canonicalizer/BLAKE3 vs roll-own | PASS — `cid_of_json` (`:688-690`) = `Provekit::Blake3.hex(Provekit::IR::Jcs.encode(value))`; `require_relative "../blake3"`/`"../ir"`; `lib/provekit/ir.rb` & `blake3.rb` are **not** in the diff; `Blake3.hex` emits `"blake3-512:" + …128 hex` via the C ext. No rolled-own JCS/BLAKE3. |
| 10 | Wiring: existing `ruby` manifest not repointed; new `ruby-source` manifest runs the new lifter; `0.1.0-draft`; no `ci/accepted`/self-pin changes; no `ruby-language-signature` in `menagerie/manifest.yaml` | PASS — `implementations/ruby/.provekit/lift/ruby/manifest.toml` unchanged (0-byte diff); new `…/ruby-source/manifest.toml` `command = ["bundle","exec","ruby","-Ilib","bin/provekit-lift-ruby-source","--rpc"]`, `version 0.1.0-draft`; no `ci/accepted/*` or `self_contracts` in diff; `menagerie/manifest.yaml` untouched, no `ruby-language-signature` entry (consistent with `csharp-language-signature`, also absent); `project_config.rs` adds `"ruby-source"` to `KNOWN_SURFACES` (correct); `Makefile` adds `build-ruby`-backed `test-ruby` + `ruby-language-signature --check` targets |
| 11 | Existing Ruby gem still builds/tests | PASS — `cd implementations/ruby && bundle exec rake test` on Homebrew Ruby 4.0.3 → 113 runs / 284 assertions / 0 failures / 0 errors (108 pre-existing + 5 new) |
| 12 | `--rpc` `initialize`/`lift`/`compile`/`shutdown` shapes match lift-plugin protocol | PASS — `initialize` → `{name,version,protocol_version:"provekit-lift/1",dialect:"ruby-source",capabilities:{authoring_surfaces:["ruby-source"],ir_version:"v1.1.0",emits_signed_mementos:false}}`; `lift` → `{kind:"ir-document",ir,callEdges,diagnostics,opacityReport,refusals}`; `compile` → `{kind:"compiled-formula",body}`; `shutdown` → `result:null`; unknown method → `-32601`; verified live |
| 13 | `cargo test -p provekit-bug-zoo --test smoke` still passes with new surface | N/A — no `provekit-bug-zoo` crate exists in this tree (item appears stale; the #593 break referenced `csharp-source` against tooling not present here). `KNOWN_SURFACES` is a plain `&[&str]` const with a unit test that only checks 4 hard-coded entries; adding `"ruby-source"` cannot break it. No `cargo` change beyond the one-line const append. |

### First thing to look at
The round-trip test (`test/test_ruby_source_lift.rb:114-129`) — the tautological-test trap that recurred on Go/Python/TS/Zig. Confirmed clean here: it compiles the **function-contract body term** via `compile_body_term` → `Compiler#stmt_lines` and re-lifts, comparing the function-contract body of the re-lift (not the `ruby:source-unit` byte copy). `compile_ir_document` filters source-unit contracts out explicitly (`ruby_source.rb:671-676`).

### Comparison to #590's bar
Meets or exceeds it. Same `<lang>:` op-namespace rule with a hard guard in the ctor helper (#590: `Provekit.Lift.Csharp` op names; here `ruby_source.rb:746`); same `arity_shape` style, including the `args` "set" slot copied from `menagerie/csharp-language-signature/specs/op_call.spec.json`; same `<source-unit:…>` contract carrying raw bytes; same `Refusal{kind,function,line,reason}` shape; same "new surface, don't repoint the existing `<lang>` manifest" discipline (#590 left `implementations/csharp/.provekit/lift/csharp/` alone); same "no `*-language-signature` in `menagerie/manifest.yaml`" (csharp is likewise absent). Ruby adds two small refinements over #590: `callee`/`method` slots marked `evaluation:"unevaluated", slot_sort:"literal"`, and an explicit metaprogramming-call denylist. The only place it's *behind* #590's polish is the `5:unresolved_call:` sort-key string (nit #1).

— T Savo

Issue #211, opened 2026-05-04
[kit:swift][axis:A] Replace regex-v0 lifter with real swift slab walker


TSavo:
Tracks the substrate-side conformance work for the swift kit (Side A of the (A)/(B) split per #176).

## Background

The swift kit was bootstrapped at commits 13d2186 + 5177bdc with a regex-v0 lifter. The bundle CID is currently a placeholder, not content-meaningful.

## Scope

- Replace the regex-v0 lifter with a real swift slab walker
- Walker emits canonical IR
- Compute a content-meaningful bundle CID
- Speak the canonical `--rpc` lift-protocol
- Add `swift-self-contracts` lift surface manifest at `implementations/swift/.provekit/lift/swift-self-contracts/manifest.toml`
- Update `KIT_TABLE` in `implementations/rust/provekit-cli/src/cmd_mint.rs` to route the swift entry to the new surface
- Add a pinned-CID test in `mint_kit_integration.rs`
- Mac-only initially (acceptable per swift toolchain constraints)

## Acceptance

- [ ] Real lifter (no regex-v0)
- [ ] Bundle CID is content-meaningful (changes when slab changes)
- [ ] `make mint-swift` produces a stable content-meaningful contractSetCid
- [ ] `provekit prove --kit=swift` exits 0 in CI on macOS
- [ ] Pinned-CID test passes

## See also

- Parent: #155
- (A)/(B) framing: #176
- Tier 1 reference PRs: #180, #183

PR #599, opened 2026-05-11
feat(swift): add Swift source lift kit (self-hosted) + swift-language-signature

TSavo:
Self-hosted Swift source-language lift kit + `menagerie/swift-language-signature/`. Mirrors PR #590 (the C# source lifter) and the merged Go/Python/TS/Zig lifters.

- Swift lifter via SwiftSyntax → `function-contract` mementos over `swift:`-namespaced ops
- per-file `swift:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test driving the real IR→source path (not the source-unit byte-copy fast-path)
- new `swift-source` lift surface (existing `swift` kit surface manifest untouched — the #590 lesson); `version 0.1.0-draft`; no `menagerie/manifest.yaml` entry (the recurring nit — `*-language-signature` catalogs don't belong in that destinations list)
- static checks pass (60 signature JSON files, no swift:unknown/binop/skip); swift build/test couldn't run locally (sandbox blocked ~/.cache/clang + DNS for swift-syntax) — CI verifies


TSavo:
## Review: PR #599 — Swift source lift kit (self-hosted)

**Verdict: MERGE-WITH-NITS** — the kit faithfully mirrors the #590/Go/Python/TS bar (and exceeds it in one place), refuses rather than fakes, reuses the package's canonicalizer/BLAKE3, and `swift build`/`swift test` is the only unverified piece (CI is the gate). Nits below are textual drift and a couple of soft spots, not correctness stops.

> Reviewed `origin/codex/swift-source-lift-kit` @ `425e6dc0` against `origin/main`. Couldn't run `swift build` locally either; this is a wiring-plausibility + conformance review. CI (`Cross-language conformance gate (macOS swift)`) is the empirical verifier.

The `--rpc` server's `lift`/`compile` handlers and the `swift-source` manifest. Everything downstream of the lifter is exercised by `swift test` in CI; if the new module compiles and `make test-swift` (now `swift test` + `swift run conformance` + LSP/crypto runners) passes, the kit is sound. `Package.swift` wiring looks complete: `swift-syntax` pinned `from: "600.0.0"`, new `ProvekitLiftSwiftSource` lib target + `ProvekitLiftSwiftSourceCLI` exec target + `ProvekitLiftSwiftSourceTests` test target all declared; products list updated. `make build-swift` (`swift build -c release`) builds the whole package, so the new target is covered.

### Blocking findings
None.

### Non-blocking nits

1. **`SwiftEffect.sortKey` drifts from the Rust peer literal.** `SwiftSourceIR.swift`:
   ```swift
   case .unresolvedCall(let name): return "5:unresolved_call:\(name)"
   ```
   `provekit-walk/src/contract.rs:298` uses `format!("5:unresolved:{}", name)` (`unresolved`, not `unresolved_call`). No wire impact — `sort_key` is an internal ordering helper, the on-wire shape `{"kind":"unresolved_call","name":...}` matches `to_value()`, and within-prefix-5 ordering still falls out by name. But it's textual drift from the peer kit; align to `5:unresolved:` for parity. Same nit applies to the static-prefix labels generally — they happen to line up (`0:reads` … `6:opaque_loop`), so this one outlier sticks out.

2. **`SwiftDefinitionCollector.visit(_ node: FunctionDeclSyntax)` silently drops un-parseable signatures.** `SwiftSourceLifter.swift:222-228`: `guard let info = try? functionInfo(for: node) else { return .skipChildren }`. If `functionInfo` throws (e.g. an exotic parameter clause), the function is omitted with no `Refusal` and no diagnostic — it just doesn't appear in the IR document. Refuse-not-fake would emit a `Refusal{kind:"unhandled-syntax", function:<name>, ...}` here instead of `try?`. Low risk in practice (`functionInfo` only does label/type string-munging), but the principle is "loud refusal, not silent skip."

3. **Nested functions are not collected and not explicitly refused.** Same visitor returns `.skipChildren`, so a `func` declared inside another `func` body is never lifted as its own contract; when the body walker hits the nested `FunctionDeclSyntax` as a code-block item it falls to `statement(...)`'s catch-all `throw unsupported(stmt, reason: "unhandled statement kind: ...")` — so it *is* refused, just via the generic path rather than a named "nested function" refusal. Acceptable; mentioning for completeness.

4. **`compileBodyTerm` hardcodes `Int` formal/return types.** `SwiftSourceCompiler.swift`: `let params = formals.map { "_ \($0): Int" }` and `-> Int`. Fine for the round-trip test (everything in the supported slice is `Int`-shaped and re-lift forgets parameter types into `Value`/`Int` sorts anyway), but if the IR carried a `Bool`/`String`-returning body term the compiled stub would mistype it. Out of the supported slice today; flag if the slice widens.

### Conformance checklist

| # | Item | Status | Note |
|---|------|--------|------|
| 1 | `swift:` prefix on every op; no `binop`/`unknown`/`skip` catch-all | PASS | `SwiftSourceIR.ctor` `precondition`s the prefix and bans `unknown`/`binop`/`skip`; all 38 `op_*.spec.json` carry `swift:…` fn_names. |
| 2 | `arity_shape` on every op spec, sane shapes | PASS | binops `named` lhs/rhs; `swift:seq` `positional` arity 2; `swift:assign` `named` target/value; `swift:and`/`swift:or` mark `rhs` `unevaluated`; `swift:ternary` marks `then_expr`+`else_expr` `unevaluated`; `swift:nilcoalesce` marks `rhs` `unevaluated`. Matches the csharp op-spec shape. |
| 3 | `swift:source-unit(bytes, operational_term)` op present + emitted with real bytes | PASS | `op_source-unit.spec.json` has `arity_shape.slots[0].evaluation:"unevaluated"`, `slot_sort:"literal"`; `SwiftSourceLifter` emits `sourceUnitContract(source:operationalTerm:)` carrying the original UTF-8 source string; test `test... sourceArgs[0]["value"] == source`. |
| 4 | Refuse-not-fake for unhandled `SwiftSyntax` | PASS | Explicit refusals for closures/trailing closures, `await`/`try`/`try?`/`try!`, optional chaining `?.`, force-unwrap (`+ .panics` effect), `throw` (lowered to `swift:throw` + `.panics`), `for-in where`, non-identifier for-in/binding patterns, property wrappers/attributes, computed properties, uninitialized decls, string interpolation, multi-element sequences (chained binops), implicit member access. Catch-alls: `expression()` ends `throw unsupported(expr, reason: "unhandled expression kind: ...")`, `statement()` ends `throw unsupported(stmt, "unhandled statement kind: ...")`. No fabricated `swift:unknown`. (Soft spot — nit #2.) |
| 5 | `fnName` provably unique across overloads/extensions/nesting | PASS (with nit) | `<module>.<typeScope chain>.<name>(<externalLabels:>)(<param,types>)->ret`; external labels + type list disambiguate overloads; `typeScope` (struct/class/enum/actor/extension stack) namespaces members; module = sanitized path. Nested functions excluded (nit #3) so no inner collisions. |
| 6 | Effects on canonical wire shapes, sorted | PASS | `SwiftEffect.value` produces `{"kind":"reads","target":…}` / `writes` / `io` / `unsafe` / `panics` / `{"kind":"unresolved_call","name":…}` / `{"kind":"opaque_loop","loopCid":"blake3-512:…"}` — matches `Effect::to_value()`. No `alloc`/resolved-`call`. `SwiftEffectSet.sortedValues()` sorts by `sortKey` ascending; loop CID set for `while`/`repeat`/`for-in`; `io` for known IO calls; `panics` for `throw`/force-unwrap. (Sort-key literal nit #1.) |
| 7 | Round-trip test drives the real IR→Swift path, not the byte-copy | PASS | `testCompileLiftRoundtripBodyTermIsByteIdentical`: lifts a function, takes its **body term** (not the `swift:source-unit` contract), `compileBodyTerm` → `CompileContext.statementBlock` (real IR→Swift codegen) → `Parser.parse(...).description`, re-lifts, asserts `JcsCanonicalizer.encode(reliftedBody) == JcsCanonicalizer.encode(body)`. `compileIRDocument` explicitly `continue`s past `swift:source-unit` contracts. |
| 8 | Determinism | PASS | `JcsCanonical.object` is order-preserving pairs (no `Dictionary` iteration leaking into canonical output); RPC `write` uses `JSONSerialization` with `.sortedKeys`; no timestamps; effect ordering is total via `sortKey`. |
| 9 | CID-correctness — reuses canonicalizer/BLAKE3 | PASS | `computeJcsCid` / `JcsCanonicalizer` / `Blake3` are `ProvekitCrypto` (Side B native substrate, byte-equivalent to rust/go/cpp/c per the module docs). No rolled-own JCS/BLAKE3 in the new module. `testCanonicalCidUsesExistingJcsAndBlake3Implementation` pins a `blake3-512:…` digest. *Supra omnia, rectum* satisfied. |
| 10 | Wiring | PASS | `Package.swift` adds `swift-syntax` dep + lib/exec/test targets; existing `implementations/swift/.provekit/lift/swift/manifest.toml` untouched; new `swift-source/manifest.toml` with `version = "0.1.0-draft"`, `command = [".build/release/provekit-lift-swift-source", "--rpc"]`, `working_dir = "."` (resolves to swift package root when `provekit prove` is run from `implementations/swift/` — same pattern as the csharp `dotnet …/bin/Release/…` manifest, and matches `make build-swift`'s `.build/release/` output). No `.provekit/ci/accepted/*` / self-contract-pin changes. **No `swift-language-signature` entry in `menagerie/manifest.yaml`** — confirmed absent (correct; `*-language-signature` catalogs aren't destinations). |
| 11 | macOS-swift flake (#211) non-required; new module tests + `make build-swift` must pass | UNVERIFIED (CI) | Couldn't build locally. `make test-swift` now runs `swift test` first; the new `ProvekitLiftSwiftSourceTests` has ~10 cases. The `swift_kit_pins_expected_contract_set_cid` flake is unrelated. |
| 12 | `--rpc` server `initialize`/`lift`/`compile`/`shutdown` match lift-plugin protocol | PASS | JSON-RPC 2.0 over stdio; `initialize` → `{name, version:"0.1.0-draft", protocol_version:"provekit-lift/1", dialect:"swift-source", capabilities:{authoring_surfaces:["swift-source"], ir_version:"v1.1.0", emits_signed_mementos:false}}`; `lift` validates `surface` + `source_paths`, returns `{kind:"ir-document", ir:[…]}`; `compile` accepts array (IR doc) or object (single term); `shutdown`/`exit` return null then stop. Matches `protocol/specs/2026-04-30-lift-plugin-protocol.md`. `--rpc` flag required (CLI errors otherwise). |
| 13 | `cargo test -p provekit-bug-zoo --test smoke` still passes with `swift-source` registered | PASS (no-op) | `swift-source` is added to `KNOWN_SURFACES` (`project_config.rs:225`) only. `tests/smoke.rs` asserts specimen counts/shapes (`reports.len()==3`, BZ-SHAPE-005/006/007) and per-language composition — it never enumerates `KNOWN_SURFACES` and doesn't touch Swift. The `KNOWN_SURFACES` unit test only asserts `typescript-source` is present. Additive, safe. |

### Comparison to #590's bar
Meets it, and **exceeds it in one place worth noting**: #590 left `csharp-source` *out* of `KNOWN_SURFACES` in `project_config.rs` (the surface exists only as `implementations/csharp/.provekit/lift/csharp-source/manifest.toml`); #599 *does* add `swift-source` to `KNOWN_SURFACES`. That's the right call (the surface should be discoverable), not a regression — and a follow-up should backfill `csharp-source` there too. Refusal discipline, source-unit wrapper, arity_shape coverage, effect wire shapes, canonicalizer reuse, manifest hygiene, and the `0.1.0-draft` versioning are all at or above the C#/Go/Python/TS standard. The agent's claims in the PR body (round-trip drives the real path, not the fast-path; canonicalizer reuse; no `menagerie/manifest.yaml` entry; existing manifest untouched) all check out against the diff.

— T Savo

PR #600, opened 2026-05-11
feat(php): add PHP source lift kit (self-hosted) + php-language-signature

TSavo:
Self-hosted PHP source-language lift kit + `menagerie/php-language-signature/`. Mirrors PR #590 (the C# source lifter) and the merged Go/Python/TS/Zig lifters.

- PHP lifter via nikic/php-parser → `function-contract` mementos over `php:`-namespaced ops
- per-file `php:source-unit` lossless wrapper; round-trip `compile` + a `lift(compile(lift(src)))` test driving the real IR→source path (not the source-unit byte-copy fast-path)
- new `php-source` lift surface (existing `php` kit surface manifest untouched — the #590 lesson); `version 0.1.0-draft`; no `menagerie/manifest.yaml` entry (the recurring nit — `*-language-signature` catalogs don't belong in that destinations list)
- PHP unit tests pass; signature generator deterministic; RPC initialize reports 0.1.0-draft/php-source; composer install couldn't run locally (no composer in sandbox) — CI verifies the php-parser-dependent compiler path


TSavo:
## Review: PR #600 — `feat(php): add PHP source lift kit (self-hosted)`

**Verdict: MERGE-WITH-NITS** — clean self-hosted PHP source lifter that meets the #590 bar on every conformance axis; a few cosmetic asymmetries and one structural divergence (the production `compile` codepath is byte-copy, not PrettyPrinter) that are worth a follow-up but do not block. CI is the verifier for the unverified `composer`/php-parser path per the task note — `prove` is green; `Cross-language conformance gate` is still pending, **merge gated on it going green**.

### Blocking findings
None.

### Non-blocking findings

1. **`PhpSourceCompiler::compileIrDocument` always takes the `php:source-unit` byte-copy fast-path in production.** Every lifted IR document begins with a `source_unit_contract` (`Ir.php:140`), and `sourceUnitBytes()` returns those bytes verbatim (`PhpSourceCompiler.php:236-250`), so the `compile` RPC (`Rpc.php:compile_rpc`) never reaches the PrettyPrinter (`compileFunctionContractWithPhpParser`). The PrettyPrinter path is only reachable via `compileBodyTerm`, which is called *only* from the test. Contrast #590: csharp's `compile` RPC takes a single IR *term* and runs it through the real `EmitStmt`/`EmitCtor` machine (`RpcServer.cs:121`, `CsharpCompiler.cs:11`) — no byte-copy short-circuit. The PHP behavior is defensible (the source-unit wrapper *is* the lossless artifact), and the body-term round-trip test (`test_compile_lift_roundtrip_body_term_is_byte_identical`) does drive PrettyPrinter→parse in CI, so the compiler is exercised. But the asymmetry is real — a `compile` of mutated IR returns stale bytes. **Look at this first.**

2. **`PhpSourceCompiler::isUnit()` (`PhpSourceCompiler.php:~210`)** — `($k==='const' && !array_key_exists('value',$term)) || ($k==='const' && ($value??null)===null)` relies on `&&`-over-`||` precedence; works, but unreadable. Simplify.

3. **`exprNode` has no `php:assign` case (`PhpSourceCompiler.php:~115-150`)** — but the lifter *will* produce nested `php:assign(var a, php:assign(var b, …))` from `$a = $b = 1;` (`Assign.expr` is an `Assign`). The compiler then throws `unsupported php operation in expression position: php:assign`. Lifter→compiler asymmetry; either refuse nested assigns in the lifter or handle them in `exprNode`.

4. **PHP `null` literal lowers to `unit_const()` (sort `Unit`)** (`PhpSourceLifter.php` ConstFetch handler) even in value position (`$x = null;`), where the slot is `Value`. Sort mismatch; harmless for CIDs but inconsistent. Confirm csharp/swift handle the same way before "fixing".

5. **Token fallback (`liftSourceWithTokenFallback`, regex-based)** diverges from the php-parser path and does not run in CI (composer is installed there). Documented as "intentionally narrow." Acceptable, but it's a second, drifting lifter — flag for eventual deletion once composer is reliably present.

### Conformance checklist

| # | Item | Status | Note |
|---|---|---|---|
| 1 | `php:` prefix on every op; no `php:binop`/`php:unknown`/`php:skip`; `==`≠`===`, `.`≠`+` | PASS | `ctor()` enforces prefix + bans the three catch-alls (`Ir.php:64-72`); `php:eq`/`php:identical`/`php:ne`/`php:not_identical`/`php:concat` all distinct (`lowerBinaryOp`) |
| 2 | `arity_shape` on every op, sane; short-circuit RHS / ternary / `??` mark `evaluation:"unevaluated"`; mirrors csharp | PASS | `op_and`/`op_or`/`op_nullcoalesce` slot[1] `unevaluated`; `op_ternary` slots 1,2 `unevaluated`; `op_seq` `positional` arity 2; `op_assign` named target/value; `op_call`/`op_methodcall` `args` slot `shape:{kind:set}` |
| 3 | `php:source-unit(bytes, operational_term)` present AND emitted carrying full file bytes | PASS | `op_source-unit.spec.json`; `source_unit_contract($path, $sourceBytes, …)` uses raw `file_get_contents`/`$source` incl. `<?php` prologue + inline HTML (`InlineHTML` is skipped from the *term* but the bytes are intact) |
| 4 | Refuse-not-fake for unhandled AST: closures/arrow-fns, generators, try/catch, switch/match, refs, variadic/named/byref args, list(), `$$x`, eval/call_user_func, traits, magic methods, instanceof, clone, `@`, goto | PASS | Catch-alls: `lowerStatement`/`lowerExpr`/`lowerTarget`/`lowerBinaryOp` all `throw $this->refuse(...)`; explicit refusals for byref returns/params, variadic, default params, named/unpack/byref args, variable-variables, dynamic calls/props/classes, traits, `__*` magic methods, anon classes, eval/call_user_func; `Yield_`/`Closure`/`ArrowFunction`/`TryCatch`/`Switch_`/`Match_`/`List_`/`Instanceof_`/`Clone_`/`ErrorSuppress`/`Goto_` all fall through to the `nodeKind` catch-all refusal (no silent skip) |
| 5 | `fnName` provably unique — `ns\fn`, `ns\Class::method` | PASS | `qualifyFunctionName` + `::` for methods; `<source-unit:path>` uses `<>` (illegal in real PHP names); PHP has no overloading, no static-vs-instance same-name |
| 6 | Effects on canonical wire shapes, sorted by `sort_key()` incl. secondary sort; no `alloc`/resolved-`call` | PASS | `EffectSet`: `reads`/`writes` w/ `target`, `io`, `panics`, `unresolved_call` w/ `name`, `opaque_loop` w/ `loopCid:"blake3-512:…"`; `sortKey` = `"0:reads:…"`…`"6:opaque_loop:…"` matching the Java/Swift `Effect::sort_key`; loops (`while`/`do`/`for`/`foreach`) → `opaque_loop`; `echo`/`print`/`file_*`/`curl_*`/`printf`/`header`/`fopen`/`fread`/`fwrite`/`readfile` → `io`; `throw`/`exit`/`die` + `trigger_error(…,E_USER_ERROR)` → `panics`; no `alloc`, calls are `unresolved_call` not resolved `call` |
| 7 | Round-trip `lift(compile(lift(src)))` byte/CID-identity driving the real PrettyPrinter, not the byte-copy fast-path | PASS (with caveat) | `test_compile_lift_roundtrip_body_term_is_byte_identical` lifts `f`, calls `compileBodyTerm` (→ PrettyPrinter in CI), re-lifts, asserts `Jcs::encode($body) === Jcs::encode($reliftedBody)`. Drives the real compiler. Caveat = finding #1: the *document-level* `compileIrDocument` is byte-copy, so `lift(compile(lift(src)))` at doc level is trivially identical via copy |
| 8 | Determinism | PASS | `EffectSet` dedups by JCS key + sorts; `liftPaths` sorts files `SORT_STRING`; `fold_seq` deterministic left-fold; no timestamps |
| 9 | CID-correctness — reuses `provekit-ir-symbolic`'s canonicalizer/BLAKE3, not rolled-own | PASS | `Ir.php` / `EffectSet.php` `use ProvekIt\Canonicalizer\Jcs` + `Blake3` (the existing classes, `bootstrap.php` requires them directly); `cid_of_json` = `'blake3-512:' . Blake3::hash(Jcs::encode(...))`. No new JCS/BLAKE3 — *Supra omnia, rectum* satisfied |
| 10 | Wiring — composer adds `nikic/php-parser`; existing `php` manifest not repointed; new `php-source` manifest w/ working `command` + `version 0.1.0-draft`; no `.provekit/ci/accepted`/self-contract changes; no `php-language-signature` entry in `menagerie/manifest.yaml` | PASS | `composer.json` adds `"nikic/php-parser": "^5.0"` + `ProvekIt\\LiftPhpSource\\` autoload + test script entry; `implementations/php/.provekit/lift/php/manifest.toml` untouched (empty diff); new `php-source/manifest.toml` → `command = ["php","provekit-lift-php-source/bin/main.php","--rpc"]`, `version 0.1.0-draft`; no accepted/self-contract diffs; `menagerie/manifest.yaml` diff empty (matches csharp-language-signature which also has no entry) |
| 11 | Existing PHP project still builds/tests | UNVERIFIED LOCALLY | no `composer` in sandbox per task note. composer wiring is plausible — only additive (new autoload key, new require, new test script line); CI `composer install && composer test` is the verifier |
| 12 | `--rpc` server `initialize`/`lift`/`compile`/`shutdown` match the lift-plugin protocol; `initialize` reports `version 0.1.0-draft`, `dialect php-source` | PASS | `Rpc.php`: `initialize` → `{name, version:"0.1.0-draft", protocol_version:"provekit-lift/1", dialect:"php-source", capabilities:{authoring_surfaces:["php-source"], ir_version:"v1.1.0", emits_signed_mementos:false}}` — mirrors `SwiftSourceRPC.initializeResult()`; `lift` checks surface, returns `ir-document`; `compile` returns `compiled-formula`; `shutdown` → null; line-delimited JSON-RPC 2.0; `test_initialize_declares_php_source_draft` covers it |
| 13 | `cargo test -p provekit-bug-zoo --test smoke` still passes with new `php-source` surface | N/A IN THIS WORKTREE | no `provekit-bug-zoo` crate exists in `implementations/rust/` on this branch — item stale. The only Rust change is `KNOWN_SURFACES += "php-source"` in `provekit-cli/src/project_config.rs` plus its unit-test assertion — clean and minimal |

Finding #1 — confirm with the maintainer that the `compile` RPC returning byte-copied source for a real lifted document is the intended PHP design (vs #590's term-in / compile-out). If a future caller mutates IR and round-trips through `compile`, they get stale bytes. Not a blocker; a follow-up issue at most.

### Comparison to #590's bar
Meets it. Same shape: self-hosted lifter + IR↔source compiler + `--rpc` JSON-RPC server + `0.1.0-draft` manifest + a `*-language-signature` menagerie (68 specs) + `KNOWN_SURFACES` registration, no menagerie/manifest.yaml entry, no self-contract pins. PHP refuses *more* aggressively than csharp (every unhandled node → `Refusal{kind,function,line,reason}`), which is the right instinct. The two deltas from #590: (a) the byte-copy `compile` fast-path (finding #1), (b) the token-fallback second lifter for composer-less environments (finding #5) — csharp has no analog because `dotnet` is always present. Neither is disqualifying.

— reviewed against `origin/main...d973584`, worktree `/tmp/pk-600-review`.

PR #603, opened 2026-05-11
feat(cpp): add C++ source lift kit (self-hosted)

TSavo:
Self-hosted C++ source lift kit + `menagerie/cpp-language-signature/`. Mirrors PR #590 (the C# source lifter) for C++ via libclang (Clang C API), as a **separate** plugin (`implementations/cpp/provekit-lift-cpp-source/`) — the existing `provekit-lift-cpp` annotation adapter is untouched.

- `cpp:`-namespaced operation algebra; `arity_shape` on every op spec; `cpp:source-unit(bytes, term)` emitted per file carrying real source bytes.
- Refuses (does not fake) unhandled C++ syntax (templates, lambdas, RAII-with-destructors, coroutines, operator overloading, RTTI, …) — `Refusal{kind,function,line,reason}`, no `cpp:unknown`/`skip` catch-all.
- `fnName` via `clang_getCursorUSR` (overload/namespace/template-safe).
- Effects on the canonical `Effect` wire shapes (`reads`/`writes`/`unresolved_call` w/ `name`/`opaque_loop` w/ `loopCid:"blake3-512:..."`/`io`/`panics`), sorted by `Effect::sort_key()` order.
- IR→C++ round-trip compiler + a `lift(compile(lift(src)))` test driving the real compiler path.
- New `cpp-source` lift manifest; `cpp-source` added to `KNOWN_SURFACES`; build/test wiring (`tools/build-cpp-source-lift.sh`, `tools/test-cpp-source-lift.sh`, Makefile `test-cpp`). No `.provekit/ci/accepted/*` / self-contract-pin / `menagerie/manifest.yaml` changes.

Verified locally: `make test-cpp`, `make test` (cpp), `cargo test -p provekit-cli project_config::...known_lists_include_anchor_entries`, `git diff --check` all pass.

## Summary by CodeRabbit

* **New Features**
  * Added C++ source code lifting capability to translate C++20 code into an intermediate representation (IR).
  * Integrated C++ lift RPC server for language signature support.

* **Documentation**
  * Added C++ language signature specifications defining operations, effects, and sorts for code analysis.

* **Chores**
  * Updated build and test pipeline with new C++ source lift build steps.
  * Added "cpp-source" to supported authoring surfaces in CLI.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/603)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,6 +171,7 @@ build-rust:
 build-cpp:
 	tools/build-cpp-self-contracts.sh --build-only
 	tools/build-cpp-lift.sh
+	tools/build-cpp-source-lift.sh
 	tools/build-cpp-lsp.sh
 
 .PHONY: build-go
@@ -615,8 +616,12 @@ test-go:
 	cd implementations/go/provekit-lift-go-tests && go test ./...
 	cd implementations/go/provekit-lift-go && go test ./...
 
+.PHONY: test-cpp-source-lift
+test-cpp-source-lift:
+	tools/test-cpp-source-lift.sh
+
 .PHONY: test-cpp
-test-cpp: build-cpp
+test-cpp: build-cpp test-cpp-source-lift
 	@echo "test-cpp: LSP lifecycle integration test"
 	sh implementations/cpp/provekit-lsp-cpp/test_lsp.sh implementations/cpp/target/provekit-lsp-cpp
 	@echo "test-cpp: mint round-trip also covered by mint-cpp"
--- a/implementations/cpp/.provekit/lift/cpp-source/manifest.toml
+++ b/implementations/cpp/.provekit/lift/cpp-source/manifest.toml
@@ -1,0 +1,10 @@
+name = "cpp-source-lift"
+version = "0.1.0-draft"
+protocol_version = "provekit-lift/1"
+command = ["./target/provekit-lift-cpp-source", "--rpc"]
+working_dir = "."
+
+[capabilities]
+authoring_surfaces = ["cpp-source"]
+ir_version = "v1.1.0"
+emits_signed_mementos = false
--- a/implementations/cpp/Makefile
+++ b/implementations/cpp/Makefile
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: Apache-2.0
+
+ROOT := ../..
+
+.PHONY: all build test test-cpp-source clean
+
+all: build
+
+build:
+	$(ROOT)/tools/build-cpp-self-contracts.sh --build-only
+	$(ROOT)/tools/build-cpp-lift.sh
+	$(ROOT)/tools/build-cpp-source-lift.sh
+	$(ROOT)/tools/build-cpp-lsp.sh
+
+test: build test-cpp-source
+	sh provekit-lsp-cpp/test_lsp.sh target/provekit-lsp-cpp
+
+test-cpp-source:
+	$(ROOT)/tools/test-cpp-source-lift.sh
+
+clean:
+	rm -rf target
--- a/implementations/cpp/provekit-lift-cpp-source/include/cpp_source_lifter.hpp
+++ b/implementations/cpp/provekit-lift-cpp-source/include/cpp_source_lifter.hpp
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "provekit/canonicalizer/value.hpp"
+
+#include <string>
+#include <vector>
+
+namespace provekit::cpp_source {
+
+constexpr const char* VERSION = "0.1.0-draft";
+constexpr const char* DIALECT = "cpp-source";
+constexpr const char* IR_VERSION = "v1.1.0";
+
+struct Refusal {
+    std::string kind;
+    std::string function;
+    int line = 0;
+    std::string reason;
+};
+
+struct LiftResult {
+    std::vector<canonicalizer::ValuePtr> declarations;
+    std::vector<Refusal> refusals;
+    std::vector<canonicalizer::ValuePtr> diagnostics;
+    std::vector<canonicalizer::ValuePtr> opacity_report;
+};
+
+struct CompileBodyOptions {
+    std::string function_name = "f";
+    std::vector<std::string> formals;
+    std::string return_type = "int";
+};
+
+canonicalizer::ValuePtr initialize_result();
+LiftResult lift_source(const std::string& path, const std::string& source);
+LiftResult lift_paths(const std::string& workspace_root, const std::vector<std::string>& source_paths);
+std::string compile_ir_document(const std::vector<canonicalizer::ValuePtr>& ir);
+std::string compile_body_term(const canonicalizer::ValuePtr& term, const CompileBodyOptions& options = {});
+canonicalizer::ValuePtr post_rhs(const canonicalizer::ValuePtr& contract);
+const canonicalizer::ValuePtr* find_contract(const LiftResult& result, const std::string& name_fragment);
+std::string canonical_bytes(const canonicalizer::ValuePtr& value);
+std::string cid_of_value(const canonicalizer::ValuePtr& value);
+std::string lift_result_json(const LiftResult& result);
+int run_rpc();
+
+}  // namespace provekit::cpp_source
--- a/implementations/cpp/provekit-lift-cpp-source/src/cpp_source_lifter.cpp
+++ b/implementations/cpp/provekit-lift-cpp-source/src/cpp_source_lifter.cpp
@@ -1187,6 +1187,10 @@ std::string expr_from_term(const ValuePtr& term) {
     if (name == "cpp:bitnot") return unary_source(args, "~");
     if (name == "cpp:deref") return unary_source(args, "*");
     if (name == "cpp:addr") return unary_source(args, "&");
+    if (name == "cpp:preinc" && args.size() == 1) return "(++" + expr_from_term(args[0]) + ")";
+    if (name == "cpp:predec" && args.size() == 1) return "(--" + expr_from_term(args[0]) + ")";
+    if (name == "cpp:postinc" && args.size() == 1) return "(" + expr_from_term(args[0]) + "++)";
+    if (name == "cpp:postdec" && args.size() == 1) return "(" + expr_from_term(args[0]) + "--)";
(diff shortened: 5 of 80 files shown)


TSavo:
## Review — PR #603 (`codex/cpp-source-lift-kit-v2`): C++ source lift kit

**Verdict: CHANGES-REQUIRED**

One real, mechanical defect: the lifter emits four `cpp:` operations that are **not in the language-signature catalog**. That is a divergent/under-specified algebra (`Supra omnia, rectum`), not a nit. Everything else is in good shape — canonicalizer is reused, effects wire shapes and sort order are correct, refusal coverage is genuinely thorough, the round-trip test drives the real IR→C++ path, all builds and tests pass locally.

---

### Blocking findings

**1. Missing op specs for the four inc/dec operations the lifter constructs.**

`implementations/cpp/provekit-lift-cpp-source/src/cpp_source_lifter.cpp:622` emits `cpp:postinc` / `cpp:postdec` / `cpp:preinc` / `cpp:predec`:

```cpp
return ctor((is_postfix_incdec(toks) ? (op == "++" ? "cpp:postinc" : "cpp:postdec")
                                     : (op == "++" ? "cpp:preinc"  : "cpp:predec")),
            {lift_target(kids[0], ctx)});
```

But `menagerie/cpp-language-signature/specs/` has **no** `op_postinc.spec.json` / `op_postdec.spec.json` / `op_preinc.spec.json` / `op_predec.spec.json`. The merged C# baseline (#590) ships all four (`op_postinc.spec.json` etc.). The lifter emits 47 distinct `cpp:` ops; the menagerie catalogs 43. The four uncovered ops are reached by the most ordinary C++ there is — `for (int i = 0; i < n; i++)` produces `cpp:postinc` in the `update` slot immediately. The LanguageSignatureMemento must catalog every op the lifter constructs.

Fix: add `op_postinc/postdec/preinc/predec.spec.json` with `fn_name: "cpp:postinc"` etc., `arity_shape` `Named [{name:"target"}]`, `effects` `MemWrite` — mirror the C# specs but include `arity_shape` (the C# inc/dec specs omit `arity_shape`; the cpp ones should not, per this PR's own README claim that "every operation ... carries an explicit `arity_shape`").

How this slipped past CI: `test_effects_use_canonical_shapes_and_sort_order` and the round-trip test deliberately use only ops the menagerie covers. A fixture with a `for`-loop / `i++` would have surfaced the gap. Recommend adding one.

---

### Non-blocking (nits / follow-ups)

- **`op_for` / `op_do` `update` slot sort = `Stmt`** but the lifter passes `lift_expr(...)` (an Expr term) for the for-loop `update` (line 818). This exactly matches the C# baseline (`csharp:for` step slot is also `Stmt`), so it is consistent with the established convention — flagging for awareness only.
- **`parse_top_level_operator` is a token-scan heuristic** over a cursor libclang has already classified as `BinaryOperator`/`CompoundAssignOperator` (lines 275–301). Fragile (treats `<`/`>` as bracket-like for nesting), but within the modeled scalar subset it resolves correctly. libclang genuinely does not expose the operator spelling for these cursors, so this is a known constraint. Follow-up, not a blocker.
- **`cpp:source-unit` only emitted when ≥1 function is modeled** (`lift_source`, line 1324: `if (!function_decls.empty())`). Matches the C# baseline (`CsharpLifter.cs:73`), so consistent. Means source bytes are dropped for a file that only produces refusals — arguably you'd want it unconditionally, but it is not a regression vs the bar.
- `source_unit_contract` uses `returnSort = prim_sort("Unit")` while C# uses `PrimSort("Int")`; cosmetic divergence, internally consistent with the cpp `op_source-unit` spec.
- `compile_ir_document` has a byte-copy fast path (decodes the hex from `cpp:source-unit`); the round-trip test correctly bypasses it via `compile_body_term` on the body term. A small test that the `compile` RPC fast path also reproduces the source would be a nice belt-and-suspenders.

---

### Conformance checklist

| # | Item | Status |
|---|------|--------|
| 1 | `cpp:` prefix on every op CID; no `cpp:binop`/`cpp:unknown`/`cpp:skip`-as-catch-all | PASS (real `cpp:skip` only for null stmt / empty seq / missing else; unhandled syntax → refusal, never a fallback op) |
| 2 | `arity_shape` on every op spec, sane; binops Named lhs/rhs; `cpp:seq` Positional/variadic; `cpp:assign` Named target/value; `cpp:and`/`cpp:or` rhs `evaluation:"unevaluated"`; `cpp:ite` then/else `unevaluated` | PASS for the 43 specs present — but see Blocker 1: 4 ops have **no spec at all** |
| 3 | `cpp:source-unit(bytes, operational_term)` op present AND emitted carrying real source bytes (hex of file) folded under `cpp:seq` | PASS |
| 4 | Refuse, don't fake — every unhandled `CXCursorKind` → `Refusal{kind,function,line,reason}` | PASS — templates, lambdas (expr+stmt), `constexpr`, `auto` (return/param/local), try/catch, `dynamic_cast`/RTTI, operator overloading, ctor/dtor/conversion, InitListExpr, FloatingLiteral, non-scalar params/returns/locals (RAII), and a default `throw unsupported` on every unrecognized expr/stmt kind → catches coroutines, fold exprs, structured bindings, `decltype`-typed-non-scalar, etc. Solid. |
| 5 | `fnName` provably unique — uses `clang_getCursorUSR` | PASS (`stable_function_name` → `cursor_usr`, falls back to qualified name only when USR empty, which doesn't happen for the `clang_isCursorDefinition` function cursors actually lifted) |
| 6 | Effects on canonical wire shapes — `{kind:"reads"/"writes",target}`, `{kind:"unresolved_call",name}`, `{kind:"opaque_loop",loopCid:"blake3-512:..."}`, `{kind:"io"}`, `{kind:"panics"}`; NO `alloc`/resolved-`call`; sorted by `Effect::sort_key()` order incl. secondary sort | PASS — `effect_value`/`effect_sort_key` (lines 411–433) match `provekit-walk/src/contract.rs` exactly, incl. `5:unresolved:` (not `unresolved_call:`); `loopCid` = `cid_of_value(loop term)` = BLAKE3-512 of the loop sub-IR's canonical JSON; `EffectSet` is `std::map` keyed by sort key → deterministic, deduped, ordered |
| 7 | Round-trip `lift(compile(lift(src)))` byte-identical, driving the real IR→C++ compiler path not the source-unit byte copy | PASS — `test_round_trip_body_term_is_byte_identical` extracts `post_rhs` of contract `f`, calls `compile_body_term` on that body term, re-lifts, asserts `canonical_bytes` equality. Not the fast path. |
| 8 | Deterministic JSON — no `unordered_map` iteration, no timestamps, stable key order | PASS — `std::map`/`std::set` throughout, ordered `clang_visitChildren`, JCS canonicalizer sorts keys |
| 9 | Reuses the existing C++ canonicalizer (`implementations/cpp/provekit/canonicalizer/`), no rolled-own JCS/BLAKE3 | PASS — links `provekit/canonicalizer/jcs.cpp` + `hash.cpp`; `canonical_bytes` → `canonicalizer::encode_jcs`, `cid_of_value` → `canonicalizer::compute_cid` |
| 10 | Existing `provekit-lift-cpp` annotation adapter + manifest untouched; NEW `implementations/cpp/.provekit/lift/cpp-source/manifest.toml` added; `version 0.1.0-draft`; `initialize` reports same; `emits_signed_mementos: false` | PASS — only the new manifest is added; `initialize_result()` reports `0.1.0-draft` / `cpp-source` / `emits_signed_mementos:false` |
| 11 | `"cpp-source"` added to `KNOWN_SURFACES` in `project_config.rs` | PASS (one-line addition) |
| 12 | No `.provekit/ci/accepted/*`, no `.provekit/self-contracts-attestations/*`, no pin changes, no `menagerie/manifest.yaml` entry, no `ci.yml`/`prove-*` matrix change | PASS — diff touches only the new kit, the two `tools/*.sh` scripts, two `Makefile` targets, the manifest, the `KNOWN_SURFACES` line, and `menagerie/cpp-language-signature/` |
| 13 | Build/test: `make test-cpp-source` (in `implementations/cpp`), `make test-cpp` (root), `cargo test -p provekit-cli project_config` | PASS — `cpp-source lifter tests passed`; full `make test-cpp` green incl. LSP lifecycle (12/12); `cargo test -p provekit-cli` 81/81 + integration suites green |

---

### Bottom line

Close. Fix the one blocker — add `op_postinc/postdec/preinc/predec.spec.json` to `menagerie/cpp-language-signature/specs/` (with `arity_shape`), and ideally add a `for`/`i++` fixture to the test suite so this class of gap can't recur — and this is a clean MERGE. The lifter is well-built: it refuses loudly, reuses the canonicalizer, and the effects/round-trip discipline holds.

TSavo:
Addressed the blocking finding in the follow-up commit just pushed: added `op_postinc/postdec/preinc/predec.spec.json` to `menagerie/cpp-language-signature/specs/` (mirroring the C# baseline — `arity_shape` Named single `target` slot, MemWrite effect; cpp op count is now 47, matching the lifter's emitted set) plus a `for (...; i++)` regression fixture that lifts the loop, asserts `cpp:postinc` in the IR, compiles the extracted body term back via the real IR→C++ path (not the source-unit byte-copy fast path), re-lifts, and asserts byte-identical IR — which required adding minimal IR→C++ emission for the existing `cpp:for` and the inc/dec ops to the realizer (lift-side emission unchanged). `make test-cpp-source-lift`, `make test-cpp` (LSP lifecycle 12/12), `cargo test -p provekit-cli project_config` all pass. No C++ signature catalog/index/mint files exist — registration is via the `op_*.spec.json` glob, so nothing else needed updating.

TSavo:
Fixed the conformance-gate failure (`build-cpp-source-lift: llvm-config not found`): the cpp-source build + test scripts now probe for `llvm-config` / `clang-c/Index.h` / a libclang library and, if any is missing, print a skip warning and exit 0 (clearing stale build outputs so CI doesn't exercise an old binary) — mirroring the C lifter's optional-libclang precedent. `PK_CPP_ENABLE_CLANG_AST=1` makes the missing-libclang condition fatal where required. No Makefile/manifest/bug-zoo change needed (`cpp` conformance routes through cpp-self-contracts; bug-zoo has no cpp-source specimens). Verified both ways: libclang present → full cpp-source build+tests pass; libclang absent → `make build-cpp test-cpp` succeed with cpp-source skipped, existing C++ adapter + LSP tests still pass. Conformance gate re-running.

PR #605, opened 2026-05-11
feat: desugaring layer — desugar-to-core rewriter + starter desugaring equation sets

TSavo:
The first concrete implementation of the desugaring-equation memento + the `desugar`-to-core-normal-form rewriter from `protocol/specs/2026-05-11-desugaring-and-the-core-compression.md` (PR #601).

**`libprovekit::desugar`:** loads a language's tagged/oriented desugaring-equation mementos (left = a surface-op term pattern with variables, right = a core composition term), rewrites a lifted IR term by repeated innermost match-and-rewrite to fixpoint = the core normal form. **Refuses** (`Refusal{kind: "non-terminating-desugaring-set" | "non-confluent-desugaring-set", …}`) when the equation set isn't certifiably confluent + terminating — conservative checks: lhs-root dependency-cycle detection (termination), duplicate/unifiable-lhs + non-variable nested-lhs overlap detection (confluence). A non-canonical normal form has no honest core image. A term still containing surface ops after `desugar` is fine (just not fully core yet — `provekit migrate` would refuse to transport that fragment until the macro is added); the rewriter never leaves a surface op silently relabelled as core.

**Starter equation sets:**
- C#: `csharp:and`/`csharp:or` (short-circuit → `csharp:ite-bool`), `csharp:for`, `csharp:source-unit` macros + a `csharp:ite-bool` core op.
- Python: `python:and`/`python:or`/`python:not`, `python:source-unit` macros + a `python:ite-bool` core op.

**Exhibit/tests:** `menagerie/desugaring-demo/` README; tests for rewrite-to-fixpoint, the two refusals, a C# `for` collapse (core-only result, same `wp` as the original, byte-deterministic CID), and a Python boolean-macro collapse.

**Notes / follow-ups:**
- Did **not** implement `csharp:foreach` or Python `for`: the pinned op sets lack a clean temp-binding / iterator-core shape, and inventing one beyond the catalog would be faking semantics. Those are real refusal-shaped extension requests, not punts.
- Spec refinement needed (does **not** touch #601's branch — flagging for fold-in): the exact JSON shape/location of the per-*set* confluence+termination obligation isn't pinned in the spec; implemented here as a gate in `DesugaringSet::new`. Also a minor wording drift — this work's prompt said `role: "macro-expansion"` / `direction: "desugar"`, the normative spec says `role: "desugaring"` / `direction: "left-to-right"`; the loader accepts both, new mementos use the spec form.

Verified locally: `cargo test -p libprovekit`, `cargo test -p provekit-ir-types -p provekit-canonicalizer`, menagerie JSON parse check, `git diff --check` all pass. No `.provekit/ci/accepted/*` / self-contract-pin changes.

### DIFF
--- a/implementations/rust/libprovekit/src/desugar.rs
+++ b/implementations/rust/libprovekit/src/desugar.rs
@@ -1,0 +1,852 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt;
+use std::fs;
+use std::path::Path;
+
+use provekit_ir_types::{IrTerm, Sort};
+use serde::{Deserialize, Serialize};
+use serde_json::Value as JsonValue;
+
+use crate::canonical::serializable_jcs;
+use crate::ProvekitError;
+
+const MAX_REWRITE_STEPS: usize = 10_000;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum RefusalKind {
+    NonConfluentDesugaringSet,
+    NonTerminatingDesugaringSet,
+    InvalidDesugaringEquation,
+    WpPreservationNotDischarged,
+}
+
+impl RefusalKind {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            RefusalKind::NonConfluentDesugaringSet => "non-confluent-desugaring-set",
+            RefusalKind::NonTerminatingDesugaringSet => "non-terminating-desugaring-set",
+            RefusalKind::InvalidDesugaringEquation => "invalid-desugaring-equation",
+            RefusalKind::WpPreservationNotDischarged => "wp-preservation-not-discharged",
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Refusal {
+    pub kind: RefusalKind,
+    pub message: String,
+    pub equations: Vec<String>,
+}
+
+impl Refusal {
+    pub fn new(kind: RefusalKind, message: impl Into<String>, equations: Vec<String>) -> Self {
+        Self {
+            kind,
+            message: message.into(),
+            equations,
+        }
+    }
+
+    pub fn kind(&self) -> &'static str {
+        self.kind.as_str()
+    }
+}
+
+impl fmt::Display for Refusal {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}: {}", self.kind.as_str(), self.message)
+    }
+}
+
+impl std::error::Error for Refusal {}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct WpObligation {
+    #[serde(default)]
+    pub kind: String,
+    pub status: String,
+    #[serde(default)]
+    pub method: Option<String>,
+    #[serde(flatten)]
+    pub extra: BTreeMap<String, JsonValue>,
+}
+
+#[derive(Debug, Clone)]
+pub struct DesugarRule {
+    pub fn_name: String,
+    lhs: EquationTerm,
+    rhs: EquationTerm,
+    pub wp_obligation: WpObligation,
+}
+
+impl DesugarRule {
+    pub fn from_json_value(value: JsonValue) -> std::result::Result<Self, Refusal> {
+        Self::try_from_json_value(value)?.ok_or_else(|| {
+            Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                "equation is not tagged as a desugaring equation",
+                vec![],
+            )
+        })
+    }
+
+    pub fn try_from_json_value(value: JsonValue) -> std::result::Result<Option<Self>, Refusal> {
+        let object = value.as_object().ok_or_else(|| {
+            Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                "equation memento must be a JSON object",
+                vec![],
+            )
+        })?;
+
+        if string_field(object, "kind") != Some("equation") {
+            return Ok(None);
+        }
+
+        let fn_name = required_string(object, "fn_name")?;
+        let role = string_field(object, "role");
+        if !matches!(role, Some("desugaring") | Some("macro-expansion")) {
+            return Ok(None);
+        }
+
+        let direction = string_field(object, "direction").ok_or_else(|| {
+            Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                format!("{fn_name} is missing direction"),
+                vec![fn_name.clone()],
+            )
+        })?;
+        if direction != "left-to-right" && direction != "desugar" {
+            return Err(Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                format!("{fn_name} has unsupported direction {direction:?}"),
+                vec![fn_name],
+            ));
+        }
+
+        let post = object
+            .get("post")
+            .and_then(JsonValue::as_object)
+            .ok_or_else(|| {
+                Refusal::new(
+                    RefusalKind::InvalidDesugaringEquation,
+                    format!("{fn_name} is missing post equation"),
+                    vec![fn_name.clone()],
+                )
+            })?;
+        if string_field(post, "kind") != Some("equation") {
+            return Err(Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                format!("{fn_name} post.kind must be equation"),
+                vec![fn_name],
+            ));
+        }
+
+        let lhs = EquationTerm::from_json(post.get("lhs").ok_or_else(|| {
+            Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                format!("{fn_name} post is missing lhs"),
+                vec![fn_name.clone()],
+            )
+        })?)?;
+        let rhs = EquationTerm::from_json(post.get("rhs").ok_or_else(|| {
+            Refusal::new(
+                RefusalKind::InvalidDesugaringEquation,
+                format!("{fn_name} post is missing rhs"),
+                vec![fn_name.clone()],
+            )
+        })?)?;
+
+        let wp_obligation = extract_wp_obligation(object, &fn_name)?;
+        if wp_obligation.status != "discharged" {
+            return Err(Refusal::new(
+                RefusalKind::WpPreservationNotDischarged,
+                format!(
+                    "{fn_name} wp-preservation obligation is not discharged: {}",
+                    wp_obligation.status
+                ),
+                vec![fn_name],
+            ));
+        }
+
+        Ok(Some(Self {
+            fn_name,
+            lhs,
+            rhs,
+            wp_obligation,
+        }))
+    }
+
+    fn lhs_root(&self) -> Option<&str> {
+        self.lhs.root_op_name()
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DesugaringSet {
+    rules: Vec<DesugarRule>,
+}
+
+impl DesugaringSet {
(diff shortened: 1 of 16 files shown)


TSavo:
## Review: PR #605 — desugaring layer (`libprovekit::desugar` + starter equation mementos + demo)

**Verdict: MERGE-WITH-NITS** — one nit has teeth (must close the `pre`-ignore seam before any conditional-desugaring memento merges; cheap, do it in this PR). The work is sound; the conservative confluence/termination checks are sound in the dangerous direction; build + all tests green.

---

### Build / test
`cargo test -p libprovekit -p provekit-ir-types -p provekit-canonicalizer` — **all pass**. `cargo test -p libprovekit --test desugar` — **5/5 pass** (innermost fixpoint, non-terminating refusal, non-confluent refusal, C# `for` collapse, Python boolean-macro collapse).
Not run: `provekit-cli` mint-kit pin tests (`python_kit_pins_expected_contract_set_cid` etc.) — those pin **lift-kit minted output** (`implementations/<lang>/.provekit/lift/<lang>-self-contracts/`), which is independent of the `menagerie/<lang>-language-signature/` specs touched here, so adding `op_ite-bool.spec.json` to a language signature does not move any pinned `contractSetCid`. No `.provekit/ci/accepted/*` / `self-contracts-attestations/*` / pin changes in the diff. The `language_signature_*.spec.json` CIDs themselves are not pinned anywhere (`core_interface.rs::language_signature_shape_drives_term_identity` constructs its own fixture). No pin break.

### Conformance checklist (the 8 points)
1. **Desugaring-equation mementos — PASS.** Tagged/oriented: `role: "desugaring"`, `direction: "left-to-right"` (the spec form) on all 7 new mementos (csharp: `eq_{and,or,for,source_unit}_desugar`; python: `eq_{and,or,not,source_unit}_desugar`). Loader accepts **both** forms (`role ∈ {"desugaring","macro-expansion"}`, `direction ∈ {"left-to-right","desugar"}`) — `desugar.rs:110,121`. `post.lhs` = surface op applied to formals (vars); `post.rhs` = core composition over formals. Per-equation `wp`-preservation obligation present under `obligations.wp_preservation` (loader also tolerates `wp-preservation` / `wpPreservation` and a top-level key — `desugar.rs:789-811`). Per-*set* confluence+termination obligation: implemented as a **gate in `DesugaringSet::new`** (`certify_termination` + `certify_confluence`, `desugar.rs:193-198`). The spec (§2.2) pins this obligation to the CeTA gate (WST→AProVE/CSI→CPF→CeTA), not to an in-process check — so the PR's in-process conservative check is a *refinement* (a cheap sound under-approximation that refuses when the real gate would refuse and refuses *more*), not a contradiction. Reasonable; does not block. **Note the trust model on `wp`-preservation:** it is an *asserted* discharge — `status: "discharged"` + a freetext `method` string ("short-circuit boolean conjunction wp identity", "definition-unfolding:csharp:for operation-contract wp", …). It does not mechanically relate `wp(lhs)` to `wp(rhs)`. This is consistent with the rest of the substrate (op-spec `wp` fields are also prose), the spec §7 explicitly routes this obligation "through the same contract machinery as any contract obligation", and the four/seven claimed equations are the mathematically-correct standard desugarings (`for(i,c,s,b) ≡ seq(i, while(c, seq(b,s)))`; `a && b ≡ a ? b : false` on Bool; `not x ≡ x ? false : true`; `source-unit(bytes, t) ≡ t`). So: honest claims, not lies — but the trust boundary is the human-authored `status` field, and a `wp`-violating memento with `status: "discharged"` would be accepted. Worth keeping in view.
2. **RHS ops in catalog — PASS.** `csharp:and`/`csharp:or` → `csharp:ite-bool` (NEW `op_ite-bool.spec.json` — real op spec: `kind: "algorithm"`, `post.kind: "operation-contract"`, real `wp`, `arity_shape` with `evaluation: "unevaluated"` on the branches, `effects`, `locus` — matches the existing `op_and.spec.json` shape exactly; **not a stub**). `csharp:for` → `csharp:{seq,while}` (in catalog). `csharp:source-unit` → bare var. python `and`/`or`/`not` → `python:ite-bool` (mirror op spec added). python `source-unit` → bare var. **No half-defined macro.** `csharp:foreach` and Python `for` were deliberately NOT given desugarings (pinned op sets lack a clean iterator-core) and were left as refusal-shaped gaps — verified: no `*foreach*desugar*` / Python `for` desugar memento exists. Nothing half-baked snuck in.
3. **Rewriter — PASS.** `rewrite_once_innermost` (`desugar.rs:315`) recurses into args/let-bindings/lambda-body **before** trying to match the head — genuinely innermost. `desugar` iterates to fixpoint (no rule applies → done), capped at `MAX_REWRITE_STEPS = 10_000` (cap exceeded → `Refusal{non-terminating-desugaring-set}`). Pattern-var binding correct: `match_pattern` (`desugar.rs:413`) — repeated occurrences of a var must bind to equal subterms (`bindings.get(name) => bound == term`), rhs `substitute` errors on unbound var. CID: `serializable_jcs(&term)` (the existing JCS canonicalizer, `crate::canonical`) → `provekit_canonicalizer::blake3_512_of` (the existing BLAKE3-512 hasher, `blake3-512:` prefix) — **not rolled-own**.
4. **Refusals — PASS, and the conservative checks are SOUND (no false-accepts).** `Refusal{kind: "non-terminating-desugaring-set"}` ← `certify_termination`: builds the root-op dependency graph (lhs-root → root-ops appearing in rhs), DFS cycle detection; cyclic ⇒ refuse. `Refusal{kind: "non-confluent-desugaring-set"}` ← `certify_confluence`: (a) two rules with unifiable lhs roots + different rhs ⇒ refuse; (b) one rule's lhs unifiable with a non-variable subpattern of another's lhs ⇒ refuse. **Soundness argument:** *Termination* — an acyclic defined-symbol dependency graph + **innermost** rewriting is terminating (args reach normal form first, so the only defined symbols left after a step come from the RHS and are all strictly lower in the DAG; assign each defined symbol = longest path length, every step strictly decreases the outermost defined symbol). So the check never accepts a non-terminating set — *given the rewriter is innermost*, which it is. *Confluence* — for a left-linear TRS, critical pairs arise only at non-variable positions of the other rule's lhs; the check covers position ε (root, via `patterns_unify`) and every non-variable subposition (via `contains_unifiable_subpattern`, which returns `false` exactly at variable positions). No overlap escapes; never accepts a set with a non-joinable critical pair. I checked an adversarial pair in my head (`f(g(x)) -> a(x)` vs `g(y) -> b(y)`: caught by the nested-subpattern branch; `f(x) -> a` vs `f(g(y)) -> b`: caught by the root branch). Both checks are *over*-conservative (will reject e.g. `f(s(x)) -> f(x)`, which is terminating, because of the `f`-self-edge, and will reject non-linear "overlaps" that don't actually unify) — false-rejects, acceptable. **One real gap (NIT, not exploitable by the shipped mementos):** `certify_confluence`'s loop is `for left in rules; for right in rules.iter().skip(idx+1)` — it does **not** check a rule against itself, so a single self-overlapping lhs (`f(f(x)) -> …`) is not flagged. Every shipped memento has lhs = `surface-op(vars…)` which cannot self-overlap, so not currently a hole — but the check should also test each rule against itself.
5. **Surface ops left over after `desugar` is fine — PASS.** `rewrite_once_innermost` only rewrites a `Ctor` head if some rule's lhs matches; otherwise it rebuilds the term verbatim (`desugar.rs:338-354`). Never relabels surface→core, never drops a node. `non_core_ops` is a read-only diagnostic. Verified.
6. **Demo / tests — PASS (with one caveat).** `csharp_for_surface_program_collapses_to_core_normal_form` asserts: (a) no `csharp:for` in the normal form **and** the exact expected core term **and** `non_core_ops` empty against the core set; (b) the applied equation carries `wp_obligations[0].status == "discharged"` — this is the substrate's notion of "wp preserved"; it does **not** literally assert `wp(result) == wp(original)` because `wp` is prose in this codebase, there is no symbolic `wp` to compare — acceptable but weaker than a literal wp-equality check; (c) byte-deterministic: `first.canonical_bytes == second.canonical_bytes` and `first.cid == second.cid` across two runs. Refusal tests are genuine: a real self-looping rule (`test:sugar(x) -> test:sugar(x)`) and a real ambiguous pair (same lhs `test:sugar(x)`, different rhs), not malformed input. Python boolean-macro test is real (`not(and(a,b))` → nested `ite-bool`, two applied rules, two obligations).
7. **Don't-break — PASS.** Existing un-oriented equation mementos (`eq_seq_assoc`, `eq_seq_skip_{left,right}`) untouched and still listed in the signatures; the loader correctly *skips* them (`role` not in {"desugaring","macro-expansion"} ⇒ `Ok(None)`). Lifters, existing op specs untouched. `lib.rs` change is a one-line `pub mod desugar;`. No pin / accepted-artifact changes (see Build/test).
8. **Build/test — PASS** (above).

---

### Blocking nit (close before merge — cheap, ~5 lines)
- **`desugar.rs:95-180` (`try_from_json_value`) — `pre` is read nowhere; a non-trivial `pre` side-condition is silently ignored.** The spec (§2) makes `pre` "an optional context (a side condition on when the macro fires)". The rewriter (`match_lhs`/`rewrite_once_innermost`) never consults it, so the moment a *conditional* desugaring memento is written, it would fire unconditionally — a silent-misfire seam on a substrate primitive, exactly the failure shape Supra omnia rectum exists to block. All 7 shipped mementos have `pre = {kind:"atomic", name:"true", args:[]}` so it's currently inert. **Fix:** in `try_from_json_value`, after reading `post`, read `pre` and **refuse** (`InvalidDesugaringEquation`) anything that isn't the trivially-true atom (or implement the side-condition check in `match_lhs`). Refuse-not-ignore. Do it in this PR rather than as a follow-up.

### Non-blocking notes for the record
- `certify_confluence` doesn't check a rule against itself (self-overlapping lhs unflagged). Not exploitable by the shipped mementos; tighten the loop when convenient.
- The termination soundness argument *depends on the rewriter being innermost*; ditto the confluence overlap check. If anyone ever changes the strategy to outermost/parallel, both certifications silently become unsound. Pin this assumption (a comment near `certify_termination` / `rewrite_once_innermost`).
- `names_alias` (`desugar.rs:844`) introduces qualified/unqualified matching (`"for"` matches `"csharp:for"`). The shipped mementos all use fully-qualified names, so this seam isn't exercised; flagging in case it's load-bearing later.
- `wp`-preservation discharge is *asserted* (`status: "discharged"` + freetext `method`), not machine-verified — consistent with the rest of the substrate and sanctioned by spec §7, but the trust boundary is the human-authored field. The four/seven claimed equations are the standard mathematically-correct desugarings; no incorrect equation ships.

**Bottom line:** the desugar primitive is honest — refuses rather than fakes, conservative checks are sound in the dangerous direction, deterministic CIDs via the real canonicalizer, no half-defined macros, no pin breaks, refusal-shaped gaps where the op sets are inadequate. Close the `pre`-ignore seam and ship.

TSavo:
Addressed the blocking nit (the `pre` seam) in the follow-up commit just pushed: a desugaring equation memento whose `pre` is anything other than the trivially-true atomic predicate `{"kind":"atomic","name":"true","args":[]}` is now refused at load time (`InvalidDesugaringEquation`) rather than silently ignored — the rewriter doesn't evaluate side-conditions, so firing a guarded rule unconditionally would be unsound. Two tests added (`parse_refuses_non_trivial_pre_side_condition`, `parse_accepts_trivially_true_pre`); `cargo test -p libprovekit` green (desugar suite 5→7).

Non-blocking follow-ups noted for a later PR: (1) `certify_confluence` doesn't overlap a rule against *itself* (a self-overlapping lhs would go unflagged) — a real conservative-check gap, not exploitable by the shipped mementos but worth closing; (2) pin the innermost-rewriter assumption (which the termination/confluence soundness rests on) in a comment; (3) the `names_alias` qualified/unqualified seam isn't exercised by the shipped mementos. Merging once the conformance gate is green.

PR #608, opened 2026-05-11
fix(lift-c-assertions): stop over-extracting callee-local state into the caller's pre

TSavo:
## Summary

The C-assertions lifter was unsound: `function_add_precondition` accepted any formula without checking whether its var-nodes were visible at the function's entry point. An `assert(local_var > 0)` where `local_var` is a C local (computed after function entry) would produce a pre mentioning that local, making the caller's precondition too tight -- it constrains state the caller cannot see or control.

**Root cause (scenario a):** `scan_function_source` scans the entire text of a function's body. `term_json_from_slice` treats any non-integer token as a `{"kind":"var","name":"..."}` regardless of whether that name is a formal, a local, or a callee-internal. No scope check existed. For example, `int f(int x) { int y = x*2; assert(y > 0); ... }` would produce `f`'s pre as `y > 0`, but `y` is not visible at `f`'s call site.

**Note on callee-body leakage:** The callee→caller leak described in the bug report (g's locals appearing in f's pre) does NOT occur under the current clang-AST path because `cursor_source` tokenizes each `FunctionDecl`'s own extent. The regression test guards this invariant against future visitor changes.

## Fix

Added an entry-state scope gate in `function_add_precondition` (`src/assertions.c`):

1. `is_entry_state_var_name`: returns true if a var name is a formal parameter, a compile-time constant operator (`sizeof`, `alignof`, `__builtin_offsetof`), or a known null constant (`NULL`/`null`/`true`/`false`).
2. `formula_references_only_entry_state`: recursively walks the formula tree, checking every `FORMULA_ATOMIC` var-node arg.
3. If the formula fails the check, it is freed and a `c-assertions.non-entry-state` opacity entry is emitted (auditable, consistent with the existing `WARN_ON` pattern). No contract is emitted if ALL predicates are dropped.

Special care was taken to preserve existing behavior:
- `BUILD_BUG_ON(sizeof(int) < 4)`: the cursor-source tokenizer emits `"sizeof ( int )"`, which passes the prefix check.
- `BUG_ON(x == NULL)`: formal `x` + known constant `NULL` both pass.
- `if (!buf) return -ENOMEM` and `if (ret < 0) return ret`: formal names pass.

## Regression test

New fixture: `tests/fixtures/assertions_callee_scope.c`
- `g(y)`: `assert(y>0)` [formal, kept] + `assert(local_of_g>0)` [local, dropped with opacity]
- `f(x)`: `assert(x!=0)` [formal, kept], calls `g(x)`
- `h(n)`: `assert(tmp>0)` [local only] -- no contract emitted, opacity emitted

Integration assertions in `tests/integration.sh` verify:
- `g` emits a function-contract (formal `y` in pre)
- `f` emits a function-contract (formal `x` in pre only)
- `"local_of_g"` does NOT appear in any function-contract
- `h` emits no function-contract (local-only assert)
- A `c-assertions.non-entry-state` opacity entry is present

## Witness regen status

**No CICP witness regeneration performed.** The existing accepted witnesses cover `assertions_basic.c`, `assertions_kernel.c`, and `assertions_refusal.c` -- none of these trigger the over-extraction bug (all asserts reference formals, `sizeof`, or `NULL`). Adding `assertions_callee_scope.c` changes the directory blast-radius CID; CI will compute fresh witnesses for the updated corpus on next run.

## Verification

```
make -C implementations/c/provekit-lift-c-assertions test
# provekit-lift-c-assertions integration passed
```

All 3 existing fixtures + new regression fixture pass. `bug-zoo` smoke does not exercise `provekit-lift-c-assertions` directly (it uses `provekit-lift-c-kernel-doc`); skipped per task guidance.

### DIFF
--- a/implementations/c/provekit-lift-c-assertions/src/assertions.c
+++ b/implementations/c/provekit-lift-c-assertions/src/assertions.c
@@ -48,6 +48,14 @@ typedef struct {
     char *name;
     char **formals;
     int n_formals;
+    /* Names of variables declared local to this function (VarDecls nested in
+     * the function body / blocks).  Owned by the FunctionContracts. */
+    char **locals;
+    int n_locals;
+    /* Names of file-scope symbols (translation-unit VarDecls and FunctionDecls).
+     * Borrowed: owned by the per-translation-unit AstContractCtx. */
+    char *const *globals;
+    int n_globals;
     const char *path;
     int line;
     int column;
@@ -482,6 +490,12 @@ static int find_top_level_operator(Slice s, const char *const *ops, size_t n_ops
             depth--;
             continue;
         }
+        /* Skip the member-access arrow "->": its '>' must not be mistaken for
+         * a relational operator (it is part of an lvalue chain like p->len). */
+        if (s.s[i] == '-' && i + 1 < s.n && s.s[i + 1] == '>') {
+            i++;
+            continue;
+        }
         if (depth != 0) {
             continue;
         }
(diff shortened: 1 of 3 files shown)


TSavo:
Folded in a follow-up commit that improves the entry-state precondition filter from PR #608.

**Where the scope info lives:** there is no per-`DeclRefExpr` cursor at term-construction time — the lifter uses libclang only to enumerate `FunctionDecl`s and collect formals; precondition formulas are built from `cursor_source` token text (space-joined token blobs), not clang cursors, so terms are flat var/const nodes whose "name" is the whole subexpression text (no member/index/deref hierarchy to tag upstream). The classification check therefore lives downstream: collect file-scope symbol names once per TU (`VarDecl` / `FunctionDecl` / enum constants via `clang_visitChildren`) plus per-function local names (`VarDecl`s under the `FunctionDecl`), thread both into `FunctionContracts`, and replace the `strstr`-once scan with a proper recursive walk of each var-node's token blob that classifies **every** identifier token by scope.

**Keep/drop policy as implemented** (`var_name_is_entry_state` + `classify_root_identifier`): keep a precondition clause as a real `pre` iff every identifier token in every var-node is (a) a formal of this function, (b) a file-scope/global symbol (incl. `static`-at-file-scope and enum constants), (c) the root of a member/index/deref/addr chain that is itself a formal/global (member names after `.` / `->` are skipped — `p->len`, `arr[0]`, `&p->field` are kept), or (d) a compile-time-constant operator form (`sizeof T`, `__builtin_offsetof`, …) or a known null/bool constant. Drop the clause → `c-assertions.non-entry-state` opacity iff any identifier is a variable local to this function (checked **before** formals, so a block-scope shadow of a formal is treated as local — conservative, since the lifter has no per-locus scope), a callee's local/param, an unclassifiable name, or the term contains a function call. When in doubt: drop + opacity, never keep-and-hope. The named bug stays fixed — a callee's locals never reach the caller's `pre`.

Also: `find_top_level_operator` now skips the member-access arrow `->` so its `>` isn't mistaken for a relational operator — this is a behavior change in the predicate slicer affecting how `->` is parsed in all lifted predicates (preconditions, errno-return guards, `BUG_ON`/`BUILD_BUG_ON`); without it `p->len > 0` was mis-sliced into garbage and then dropped. None of the existing accepted fixtures (`assertions_basic`/`kernel`/`refusal`) contain `->` in a lifted predicate, so their accepted output is unchanged.

**New test cases** (`assertions_callee_scope.c` + `integration.sh`): `assert(formal>0)` kept; `assert(file_scope_global>0)` **kept** (global is entry state); `assert(p->len>0)` **kept** (param-reachable memory); `assert(arr[0]!=0)` / `assert(arr[i]>=0)` **kept**; `assert(local>0)` **dropped** (no contract, opacity emitted, `local` absent from any `pre`); `assert(p->arr[local_idx]!=0)` **dropped** (the multi-var case the old `strstr`-once missed; `local_idx` absent from any `pre`); the callee-leak case (`g`'s own local never in `g`'s or `f`'s `pre`); a block-scope local shadowing a formal → dropped.

**CICP witnesses:** no regeneration needed — there are no committed accepted-witness files for this lifter; CI recomputes the directory blast-radius CIDs on the next run. The existing fixture corpus' lifted output is unchanged (all use only formals / sizeof / NULL with no `->`); the new keep-globals / keep-param-memory behavior is exercised only by the new fixture.

**Verification:** `make -C implementations/c/provekit-lift-c-assertions test` passes (all existing + new); `provekit-lift-core` and `provekit-lift-composition` tests pass; clang and non-clang builds clean (no new warnings — the 3 pre-existing unused-function warnings in the non-clang build are unchanged). bug-zoo does not exercise this lifter (only a README mention), consistent with this PR's original note.

PR #609, opened 2026-05-11
feat: comprehensive cross-language program transport (concept hub + refinement morphisms + provekit migrate)

TSavo:
## Summary

- Delete the synthetic `_common_core` spoke layer: 412 spoke specs from language-signature dirs, 412 spoke-anchored morphism specs/receipts/discharges from concept-shapes. The invariant: a signature must only contain operations a lifter actually emits; nothing emits `lang:x_common_core` terms.
- Rework `mint_language_morphisms.py` to mint morphisms directly from real lifter-emitted op specs with discharged refinement obligations. Remove `ensure_core_spokes()`, `core_spec_name()`, `core_spec_from_concept()`, `should_generate_core_spoke()` entirely. `diff_reason()` now records actual vs. expected values in every gap entry.
- Result: 53 morphisms minted from real ops (c11: 36 + 5 preserved PRIMITIVE_STEMS; csharp/zig/rust seq/skip/conditional; typescript ushr; php throw; csharp new). 429 precise gaps in `transport-gaps.md`.
- Update CLI transport tests: python/go transport now correctly refuses (`transport-time:no-target-morphisms`) because their arithmetic op specs carry `pre:true` not the `no_signed_overflow` precondition `concept:add` requires. Tests assert the refusal instead of the formerly fictional transport.
- Update PTP spec §2.3, §2.4, §3.3, §4, §7, §10 to remove spoke terminology and document the real-op refinement discharge model.

## Test plan

- [x] `cargo check` clean (warnings only)
- [x] `cargo test -p libprovekit` -- 4 transport tests pass
- [x] `cargo test -p provekit-cli --test transport_cli` -- 5 tests pass (C→Rust succeeds, C→Python/Go refuses)
- [x] `make -C implementations/c/provekit-walk-c test` -- term-project-ok
- [x] `menagerie/concept-shapes/mint.sh` -- 48 concept ops, 53 morphisms, 429 gaps
- [x] Zero `_common_core.spec.json` files remaining
- [x] Zero stale spoke rows in `cids.tsv`
- [x] 10 PRIMITIVE_STEMS morphisms/receipts intact and untouched

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #614, opened 2026-05-12
feat(concept-shapes): demote and/or/foreach, re-spec ite, effect-⊆ discharge — the hub becomes the true common core

TSavo:
## Summary

**Morphisms: 91 → 124 (+33). Gaps: 391 → 328 (-63).**

### Task A: Demote `concept:and`, `concept:or`, `concept:foreach`

`and`/`or` are McCarthy desugarings of `concept:ite`, not independent primitives:
- `a && b = ite(a, b, false)`
- `a || b = ite(a, true, b)`

Removed both from `OPS` and `PORTABLE_ALL`. Added `eq_and_to_ite_desugar.spec.json` + `eq_or_to_ite_desugar.spec.json` for c11, typescript, java, ruby, php, and rust (csharp + python already had them). go + zig have no inline ternary expression and get no desugar mementos.

`concept:foreach` demoted: no common iterator protocol across the 10 languages. `foreach`-using programs correctly produce transport refusals. Noted in Semantic Restrictions.

### Task B: Re-spec `concept:ite` to the common-denominator shape

Dropped the `unevaluated_slots` patch on `ite` (it was blocking discharge — the majority of languages do NOT mark ite slots unevaluated in their specs). Added renaming for java's `when_true`/`when_false` slot names to match `then_expr`/`else_expr`. Fixed `java:op_ite.spec.json` to remove the redundant `slot_sort` annotation on the `cond` slot.

`concept:ite` now gets morphisms from **c11 and java** (was 0).

### Task C: Sound effect-⊆ discharge relaxation + union-effect concept-op specs

Added a fourth structural discharge relaxation in `try_structural_subsumption`: discharge when `lang.effects` (as a set) ⊆ `concept.effects` (as a set). The safe direction only — concept promises "at most these effects"; a lang op that does fewer (or equal) effects is a valid refinement. The reverse is never discharged.

Concept ops now declare the **UNION** of all language effects for the same op via `_collect_union_effects()` called from `main()`. This ensures the effect-⊆ relaxation can discharge language ops whose effect sets are proper subsets of the union.

Discharge methods: `structural-effect-subset`, `structural-wp-abstraction-and-effect-subset`, and the full combinations with pre-weakening. The `effect_subset_relaxed` flag is recorded in morphism receipts.

This relaxation picked up 33 new morphisms across seq, skip, conditional, break, continue, return, assign, neg, and more.

### Task D: Divergence-gap reason audit

Added `KNOWN_DIVERGENCE_REASONS` override table with precise per-(lang, slug) reasons for genuine semantic gaps:
- `python:div`: true division (5/2==2.5) vs integer division
- `python:mod`: floored remainder (follows divisor sign) vs truncated-toward-zero
- `python:add/sub/mul/neg/lt/le/gt`: polymorphic dispatch on operand type
- `typescript:add`: number | string concatenation vs integer-only

`diff_reason()` now accepts `lang_id` + `slug` and checks the override table before falling back to structural diffs.

## Verification

- `mint.sh` runs clean; 2nd re-run byte-clean (same counts each time)
- `driver.sh` clean; cross-language-port refusals for `c11:and` are expected (correctly refuses)
- `cargo test -p libprovekit` — all pass
- `cargo test -p libprovekit transport` — 4/4 pass
- `cargo test -p libprovekit desugar` — 3/3 pass
- `cargo test -p provekit-cli --test transport_cli` — 5/5 pass
- `git ls-files | grep -c _common_core` = 0 (no spokes reintroduced)
- No remaining `concept:and/or/foreach` references outside `transport-gaps.md` Semantic Restrictions and generated transport-artifact term files


TSavo:
## Review — PR #614 `feat/hub-cleanup-and-discharge`

**Verdict: MERGE-WITH-NITS**

Adversarial pass for unsound discharge / unsound desugaring. None found. The two structural soundness invariants hold: effect-⊆ relaxation only mints in the `lang ⊆ concept` direction, and the concept op's effects are the *union* over the 10 languages, computed *before* the concept shape CID. The desugar mementos are sound McCarthy identities semantically and — importantly — currently inert (nothing in the transport machinery consumes them), so they cannot cause an unsound discharge.

### The three definitive statements requested

**(a) Are all 6 `and/or→ite` desugarings sound McCarthy identities with correct slot names + literals?**
Yes, semantically. Each `eq_and_to_ite_desugar.spec.json` rewrites `<lang>:and(lhs,rhs) → <lang>:ite/ternary/conditional(lhs, rhs, false)` and `eq_or` → `(lhs, true, rhs)` — the McCarthy identities, positional args (not slot-name keyed, so no slot-name-mismatch surface). Two nits, both currently harmless because the mementos are inert (no Rust code reads them; `grep` finds only doc references in `transport-gaps.md` + the mint script's gap text):
- The rhs `false`/`true` literal carries `sort: Bool` (or `Expr`/`Value` for ts) while the target ternary's then/else slots are typed `Expr`/`Value` in c11/java/rust/php/ruby — a literal-sort mismatch that would matter if a future PR wires these into transport.
- `obligations.wp_preservation.status: "discharged"` with `method: "McCarthy short-circuit boolean identity"` is rigorous only if the target `ite` marks then/else `evaluation: "unevaluated"`. php/ruby/typescript do; **c11 (`op_conditional`), java (`op_ite`), rust (`op_ite`) do not**. So for those 3 the discharged-status is aspirational, not proven. Nit only — fix when the mementos become load-bearing.

**(b) Is the effect-⊆ relaxation sound (only the lang⊆concept direction)?**
Yes. `_effects_subset()` is `return lang_effects_set <= concept_effects_set` (lang ⊆ concept), and `try_structural_subsumption` does `if not effects_ok: return None` before attempting any relaxation — so a lang op with *more* effects than the concept promised is never discharged. `_collect_union_effects()` is called in `main()` before `concept_specs[...]["effects"]` is set and before any shape CID is generated, so morphisms discharge against the union'd concept op. The hand-edits adding effect lists into `*_shape.spec.json` (index, div, add, …) are redundant belt-and-suspenders — the union override wins regardless — but not harmful. Construct-an-example check: hypothetical `lang:foo effects:[panics,io]` vs `concept:foo effects:[panics]` → `{panics,io} ⊄ {panics}` → `effects_ok` false → not discharged. Correct. Reverse → discharged. Correct.

**(c) Is the `java:op_ite.spec.json` change benign?**
Yes. The only change is removing a redundant `slot_sort: "Bool"` annotation on the `cond` slot (a no-op cleanup). Arity, formal sorts, return sort, pre, wp, effects unchanged. It does change `java:ite`'s CID → java signature CID, but the java signature isn't pinned (discovery phase) and nothing references the old CID; the regenerated morphism `morphism_java_ite_to_ite` anchors to the new `op_ite.spec.json` CID. The `concept:ite` re-spec (`ite_shape.spec.json`: drop `unevaluated` markers on then/else; effects then overridden by the union) yields the common-denominator ternary; `morphism_c11_conditional_to_ite` and `morphism_java_ite_to_ite` newly appear, anchored to the real c11/java op CIDs, with `discharge_method` matching `homomorphism_obligation.kind` and `discharged: true`.

### Conformance checklist (9 points)

1. `and/or→ite` desugarings — **PASS** (sound McCarthy; 2 nits: literal sorts + aspirational wp-preservation note for c11/java/rust; inert so non-blocking)
2. Effect-⊆ relaxation only mints lang⊆concept; union computed pre-CID — **PASS**
3. `concept:ite` re-spec + `java:op_ite` change benign; new ite morphisms anchored correctly — **PASS**
4. Nothing references the demoted ops — **PASS** for menagerie specs / README / mint script / Rust. The 18 `cross-language-port/artifacts/classify/*/concept.term.json` files still contain `concept:and` — but these are **non-authoritative leftovers from PR #609**; the authoritative `transport-report.json` was regenerated and now correctly emits the refusal `operation c11:and has no discharged morphism into the concept hub` (confirmed via `driver.sh` run). Nit: delete/regen those stale term files in a follow-up.
5. Zero spokes (`git ls-files | grep -c _common_core` = 0); `mint.sh` re-run → `git status` clean (45 concept ops, 124 morphisms, 328 gaps); `driver.sh` re-run → `git status` clean; the 12 PR-#604 idiom morphisms (`morphism_*_to_shape.spec.json`) unchanged — **PASS**
6. Morphisms anchored to real op CIDs; receipts honest — **PASS** (spot-checked `morphism_java_ite_to_ite`, `morphism_c11_conditional_to_ite`, `morphism_c11_div_to_div` + effect-subset receipts; `source_contract_cid`/`shape_cid`/`method`/`discharged: true`/`effect_subset_relaxed` consistent across spec, receipt, cids.tsv)
7. `KNOWN_DIVERGENCE_REASONS` only improves gap-reason wording for genuine divergences (python:div/mod/add/sub/mul/neg/lt/le/gt, typescript:add) — checked from the only call site (`diff_reason(...)` in the gap-recording branch); causes no extra discharges. `transport-gaps.md` count = 328 gap rows. **PASS**
8. Tests: `cargo test -p libprovekit` (incl. `transport`, `desugar`), `cargo test -p provekit-cli --test transport_cli` — **all pass**. No Rust test file was modified (the demotion's refusal-change for the `classify` program surfaces in regenerated artifacts, not in a CLI test assertion). **PASS**
9. Go/Zig short-circuit gap documented — **PASS**. `transport-gaps.md` "Semantic Restrictions" records the demotion + that languages without a boolean ternary (Go, Zig) don't transport `and`/`or`; the `classify` exhibit (which uses `&&`) now produces the precise refusal `operation c11:and has no discharged morphism into the concept hub`.

### Non-blocking nits (follow-up)
- Desugar memento literals carry `sort: Bool`/`Expr`/`Value` that don't match the target ternary's then/else slot sorts in c11/java/rust/php/ruby; and the `McCarthy short-circuit boolean identity` wp-preservation status is aspirational for c11/java/rust whose `op_ite`/`op_conditional` don't mark then/else `unevaluated`. Fix both when these mementos are wired into transport.
- 18 stale `cross-language-port/artifacts/classify/*/concept.term.json` reference `concept:and`; superseded by the regenerated `transport-report.json` refusals. Delete or regenerate.

Supra omnia, rectum: the catastrophic failure modes (unsound morphism, unsound desugaring) are not present. Ship it.

— T Savo

PR #620, opened 2026-05-12
feat(libprovekit): wp_rule schema + substitute/apply formula nodes + the wp evaluator (wp-as-formula PR 1/N)

TSavo:
Implements **PR 1 of the wp-as-formula sub-project**, per the normative spec [`protocol/specs/2026-05-13-wp-as-formula.md`](../blob/main/protocol/specs/2026-05-13-wp-as-formula.md) (merged, #613).

## What this PR is scoped to

Schema + grammar + evaluator + the walk/compose dedup refactor it enables. **Not** in scope (later PRs): migrating the ~45 concept hub ops to carry `wp_rule` (PR 2), the c11 + rust lifters (PR 3), the other 8 lifters (PR 4, one PR with per-lifter receipts), flipping the morphism discharge to a real `⊑` check (PR 5), moving the desugaring `wp`-preservation obligations onto the evaluator (PR 6). The hub-op / lifter `wp_rule`s are not touched here; the `wp_rule` field is *supported* and exercised by local test fixtures only.

## The `wp_rule` field (spec §1)

`wp_rule` is a term in the *existing* IR formula language, a sibling of `pre`/`post` on an operation contract, parametric in the reserved postcondition meta-var `Q` and per-`Stmt`-slot transformer meta-vars `wp_<slot>`. It is **synthesized** for value-ops from `pre`/`post` (`wp(op, Q) = pre ∧ Q[result := value_expr]`, with `value_expr` read off the `post`'s `result == value_expr` shape), so it is REQUIRED only on the ~10–20 control-flow ops per language; the old prose `wp` becomes an optional non-load-bearing `wp_note`. The operation-contract CDDL with these fields is already in the merged spec (§1.4); the on-wire shape (the `post` block of an `algorithm` spec, copied verbatim by `provekit-mint-amp`) flows `wp_rule` through unchanged, and the JCS canonicalizer handles the new optional field byte-deterministically — nothing in the mint path or the canonicalizer needed editing for that. Test fixtures in `wp::tests` carry authored `wp_rule`s for the control-flow ops (`if` = the Dijkstra rule, `seq` = `wp_first(wp_second(Q))`, `return` = `Q[result := value]`, `skip` = `Q`); value-ops (`add`, `div`, `uop_neg`, `bop_eq`) carry only `pre`/`post` and exercise the synthesis path.

## The `substitute` / `apply` grammar nodes (spec §2.3)

Two new `IrFormula` nodes — the smallest addition that lets a `wp_rule` be written in the formula language the rest of the substrate already speaks:

- **`substitute { target, var, term }`** — explicit, capture-avoiding, single-variable substitution on a formula: `Q[var := term]`. Needed as a *node* because in a `wp_rule` schema the `target` is the postcondition meta-var, not yet known; once `target` is ground the node is eliminated by performing the substitution. CDDL locked key order: `kind, target, term, var`.
- **`apply { fn, args }`** — application of a slot-transformer meta-var (`wp_<slot>`) to one formula argument: `apply(wp_<slot>, X)` is "the weakest precondition of the term in slot `<slot>` w.r.t. X." Eliminated by inlining the actual slot transformer. CDDL locked key order: `args, fn, kind`.

Both are conservative extensions (an existing IrFormula is still an IrFormula) and reduce to the existing grammar when fully instantiated. The postcondition meta-var `Q` is the nullary atomic predicate `Atomic { name: "Q", args: [] }` (the IR formula grammar has no `var` node; a propositional variable is a nullary atomic — confirmed against how `provekit-walk` lifts branch conditions). Added to `protocol/provekit-ir.cddl`, to `provekit-ir-types` (manual codegen extension, like the `Sort` grow; CDDL stays source of truth), and swept across every now-non-exhaustive `match` on `IrFormula` in the workspace (the coq + smt-lib + lean compilers, `provekit-ir-symbolic`'s conversion, the smt-lib validator, the aarch64 lifter) with a uniform refusal arm — these nodes never reach those sites once `libprovekit::wp` has run, so reaching the arm is a bug. Plus `provekit-ir-codegen/tests/cddl_valid.rs` — the protocol grammar must parse + validate under the `cddl` crate (guards the grammar file when new node kinds are added).

## `libprovekit::wp` — the evaluator (spec §2)

New module `implementations/rust/libprovekit/src/wp.rs` (+ `wp/tests.rs`), three things:

1. **The capture-avoiding substitution algebra** over `IrFormula` / `IrTerm`, plus free-variable computation. Previously *duplicated* between `provekit-walk`'s `wp.rs` and `libprovekit::compose` (the latter's source comments call it out: "duplicated from walk's wp.rs"). Now a single canonical home here (CCP §5: one primitive, every consumer). `compose.rs` deletes its copy and re-exports from `wp`; `provekit-walk/src/wp.rs` does the same. No observable behavior change to `walk` or `compose` (their `result == value_expr` extraction, their byte-identical output, their tests are unaffected).
2. **The two schema-node reductions** — `substitute` with a ground `target` → perform the substitution; `apply(wp_<slot>, X)` → recurse into the slot's sub-term with `X`. After `wp(t, Q)` with a ground `Q` the result formula contains no `Substitute`/`Apply` nodes and no `Q`-placeholder — asserted in the tests via `contains_schema_node`.
3. **The evaluator `wp(t, Q)`** — structural recursion over the term. For each op node: look up its contract; take its authored `wp_rule` or synthesize one for a value-op from `pre`/`post`; instantiate with this `Q`, the recursively-computed `wp`s of the `Stmt`-typed sub-terms (for the `wp_<slot>` meta-vars), and the values of the value-typed sub-terms (for the formals). Handles: `seq(s1,s2)` → `wp_s1(wp_s2(Q))`; `if`-style → the Dijkstra rule `(c ⇒ wp_s1(Q)) ∧ (¬c ⇒ wp_s2(Q))`; `while`/loops → `Refusal::OpaqueLoop { loop_cid }` naming the missing `LoopInvariantMemento` (until the resolver supplies an instantiated rule); unresolved/indirect calls and unknown ops → `Refusal::OpaqueCall { callee }`. The evaluator is **total** (always a formula or a named `Refusal`/`WpError` — never a panic), **deterministic** (recursion order fixed by slot order; the computed `wp` is canonicalized + content-addressed downstream, so nondeterminism would be a correctness bug), and makes **no solver call** (pure formula construction; z3 comes in at the discharge step of spec §3, later).

`OpContractInfo` + `OpContractResolver` are the consumed-contract view + lookup trait; the catalog-backed resolver lands with the hub-op migration (PR 2).

### Module structure
- `wp.rs`: reserved-name constants + `postcondition_placeholder()`; `Refusal` (OpaqueLoop / OpaqueCall); `SlotKind` / `SlotInfo` / `OpContractInfo` (with `value_expr()`, `synthesize_value_rule()`, `rule()`) / `OpContractResolver`; `WpError` (ArityMismatch / NoRule / MalformedRule / Refused); `wp` + `wp_with_result_var` (the public entry) → `wp_op` → `value_expr_of_term`; `reduce_formula` (the schema-node eliminator); `find_result_equation`; the substitution algebra + free-vars (the canonical home).
- `wp/tests.rs`: 17 tests — synthesized `wp` for `add` (`pre ∧ Q[result := lhs+rhs]`) and `div` (`not_zero(rhs) ∧ Q[result := lhs/rhs]`), incl. a `Q` that mentions `result`; the authored Dijkstra-`if` rule instantiated on `if(bop_eq(x,0), return(uop_neg(22)), return(x))`; the evaluator on `seq(if(c, return(x), return(y)), skip)`; the opaque-loop / unresolved-call / unknown-op / malformed-rule / arity-mismatch paths; `substitute` / `apply` serde round-trip + **pinned BLAKE3-512 CIDs** (byte-determinism — *Supra omnia, rectum*).

## Verification

- `cargo test -p libprovekit` (incl. `wp` + `compose`): **all pass** (17 in `wp::tests`).
- `cargo test -p provekit-ir-types -p provekit-canonicalizer`: **all pass.**
- `cargo test -p provekit-ir-compiler-{smt-lib,coq,lean} -p provekit-ir-symbolic -p provekit-lift-asm-aarch64`: **all pass** (the swept crates).
- `cargo test -p provekit-ir-codegen --test cddl_valid`: **pass** — `protocol/provekit-ir.cddl` parses + validates under the `cddl` crate.
- `cargo build --manifest-path implementations/rust/Cargo.toml`: clean, no warnings in `libprovekit`.
- `git diff --check`: clean.
- Full-workspace `cargo test --workspace --no-fail-fast` has some red — `provekit-bug-zoo`/`provekit-supply-chain-rails` smoke, `provekit-cli mint_kit_integration` CID pins, `provekit-linker --doc` ("can't find crate for `provekit_verifier`"). **These are pre-existing**: verified by running `rust_kit_contract_set_cid_is_pinned_to_self_contracts_canonical` on a pristine `origin/main` worktree — it fails identically there. Not caused by this PR.

## Note on `provekit-walk`

The Rust `provekit-walk` crate is **not currently a workspace member** — it was dropped from `implementations/rust/Cargo.toml`'s `members` at commit `5ec16225` (PR #387, a baselines chore), and its `version.workspace = true` makes it unbuildable standalone; nothing in CI builds it. Its `src/wp.rs` is updated here for the refactor (it re-exports the moved substitution functions from `libprovekit::wp`; `src/contract.rs` keeps importing `libprovekit::compose::substitute_in_formula`, which still resolves via the new re-export) — the change lands ready for the moment `walk` re-joins the workspace. The task's `cargo test -p provekit-walk` verify line therefore can't run today regardless of this PR.

## Next PRs (~6 total)

2. migrate the ~45 concept hub ops to carry `wp_rule` (+ rename `wp` → `wp_note`); 3. the c11 + rust lifters; 4. the other 8 lifters (one PR, per-lifter receipts); 5. flip the morphism discharge from the two structural relaxations to a real `∀Q. wp(concept:op,Q) ⇒ φ(wp(lang:op,Q))` check; 6. move the desugaring `wp`-preservation obligations onto the evaluator (`wp(lhs,Q) ⇔ wp(rhs,Q)` as a portfolio bi-implication).

## Summary by CodeRabbit

* **New Features**
  * Added weakest-precondition (WP) evaluation engine with support for term-to-formula transformation and contract resolution.
  * Extended IR formula grammar to include schema nodes for capture-avoiding substitution and function application.

* **Bug Fixes**
  * Added validation to ensure WP schema nodes are properly reduced before backend compilation.

* **Tests**
  * Added comprehensive test suite for WP evaluation and CDDL protocol validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/620)


TSavo (commented):
**Verdict: MERGE-WITH-NITS**

Adversarial review for *Supra omnia, rectum*. This is the foundation the rest of the wp-as-formula chain (PRs 2-6) builds on, so I went hard on totality, determinism, schema-node elimination, and whether the dedup is truly behavior-preserving. The evaluator is correct, the grammar extension is well-locked, the tests pin the wire encoding, and the swept backends fail loudly rather than silently. Nothing here blocks; the nits should land before PR2 if cheap.

## What I verified

- `cargo test -p libprovekit` — 17 `wp::tests` + compose + everything else: green.
- `cargo test -p provekit-ir-types -p provekit-canonicalizer -p provekit-ir-codegen -p provekit-ir-compiler-{smt-lib,coq,lean} -p provekit-ir-symbolic -p provekit-lift-asm-aarch64`: all green, including the new `cddl_valid.rs` (the protocol grammar parses + validates under the `cddl` crate).
- `cargo build --workspace`: clean (only pre-existing warnings in `provekit-lsp` / `provekit-lift-openapi`).
- `git diff --check`: clean.
- Pre-existing red spot-check on `origin/main` (`pk-main-620` worktree, `3e7b021f`): `rust_kit_contract_set_cid_is_pinned_to_self_contracts_canonical`, `cpp_kit_contract_set_cid_is_pinned_to_self_contracts_canonical`, and `kits_with_real_contracts_produce_nonempty_contract_set` fail identically there. Confirmed pre-existing, not introduced by this PR.
- `provekit-walk` is genuinely not a workspace member (`implementations/rust/Cargo.toml` line-checked; `libprovekit`, no `provekit-walk`). The refactor in `provekit-walk/src/wp.rs` is dead-code-until-walk-rejoins — acceptable.
- `menagerie/cross-language-port/driver.sh`: exits 0 on this branch; same artifact-drift `git status` as on pristine `origin/main`, so pre-existing menagerie behavior, not caused by this PR.
- `menagerie/concept-shapes/scripts/mint.sh`: does not exist on this branch (only `.py` files). Not exercised; the diff touches no canonicalizer/mint/transport surface, so mint flow is unaffected by construction.

## Correctness of the evaluator (the load-bearing piece)

- **Value-op synthesis** `pre ∧ Q[result := value_expr]`: `find_result_equation` extracts `value_expr` from `post = (result == value_expr)` in either argument order, recursing into `And`. The `Substitute` node uses the placeholder as target and the per-contract `result_var`. A `pre` that is literally `true` is dropped (sound: `true ∧ φ = φ`). Correct WP for an assignment.
- **`seq` → `wp_first(wp_second(Q))`**: the authored rule is `apply("wp_first", apply("wp_second", Q))`. `reduce_formula`'s `Apply` arm reduces the arg first (inner `apply("wp_second",Q)` → `wp(second_term,Q)`) then recurses `wp(first_term, that)`. Order correct.
- **Dijkstra `if`**: authored, instantiated as `(cond ⇒ wp_then(Q)) ∧ (¬cond ⇒ wp_else(Q))`. The evaluator has no built-in opinion about `if` — pure data-driven, exactly what the spec calls for. No-else is the authoring lifter's responsibility (`else_branch := skip`).
- **Loops**: `Refusal::OpaqueLoop` is checked before recursion into the loop body. The body is never recursed into. Loops cannot cause non-termination of the evaluator. Right call for PR1.
- **Unresolved / unknown ops**: `Refusal::OpaqueCall` for both. Naming the callee. Correct.
- **Totality**: every `Term` variant covered; every `IrFormula` variant covered in `reduce_formula`. The one `.unwrap()` (`synthesize_value_rule` line 292) is guarded by `if conj.len() == 1`. Recursion bottoms out on finite term trees; loops refuse; no cyclic-vocabulary risk. No panic path under normal data.
- **Determinism**: `HashSet` (in `free_vars_*`) is used only for `.contains` and `fresh_name`'s `u32` counter walk — no iteration order leaks into output. `reduce_each` is `.into_iter().map().collect()`. Rule lookup by name. ✓
- **Schema-node elimination**: `reduce_formula` recurses through every operand / body / target / args. `Substitute` reduces the target first, then `substitute_in_formula` (which preserves but does not eliminate nested `Substitute`/`Apply`) — but the target is already reduced, and the `term` side is an `IrTerm` (cannot contain formula schema nodes). `Apply` returns `wp(slot_term, arg)` which is itself fully reduced. The tests' `contains_schema_node` assertions are tight.

## Grammar extension is correctly locked

- `IrFormula::Substitute { target, term, var }` with JCS key order `kind, target, term, var` (CDDL `SubstituteFormula`); `IrFormula::Apply { args, fn }` with key order `args, fn, kind` (CDDL `ApplyFormula`). Verified against `#[serde(rename = ...)]` and the CDDL file. Pinned CIDs (`PINNED_SUBSTITUTE_NODE_CID` / `PINNED_APPLY_NODE_CID`) lock the wire encoding.
- The `Q` placeholder = `Atomic { name: "Q", args: [] }` encoding is sound. Collision risk with a legitimately-named predicate `Q` is theoretical only — the convention is documented and the placeholder is reserved by name; a contract author choosing `Q` as an unrelated atomic name would be self-inflicting. Worth keeping an eye on once the hub ops migrate (PR2).
- Every previously-exhaustive `match` on `IrFormula` got a new arm: coq, lean, smt-lib emitter/validator/opacity/free-vars/ctor-decls/outlives, ir-symbolic, aarch64. None silently miscompile; they all refuse loudly (`unreachable!` or `Err`).

## Nits (none blocking)

1. **PR-body "byte-equivalent" claim is slightly overstated.** In `substitute_in_term`'s `Let` arm, the new `libprovekit::wp` version (a) checks `b.name == var_name` shadowing *before* the alpha-rename block, and (b) extends the `taken` set with prior bindings' (possibly-renamed) names. Both are strict improvements over the old `compose.rs` copy. Concretely: the old code, when a `Let` binder's name equals `var_name` *and* `var_name` is free in the replacement, would alpha-rename the shadowing binder away instead of stopping substitution. The new code shadows correctly. Under *Supra omnia, rectum* the new behavior is strictly more correct, but the PR body should say "**no observable behavior change on any currently-exercised input; the dedup'd version closes a latent capture-edge in `Let` shadowing**" rather than "byte-equivalent by construction."
2. **CDDL `ApplyFormula.args: [+ IrFormula]` (one-or-more) vs Rust `Vec<IrFormula>` (allows empty).** Minor laxness; `reduce_formula`'s `Apply` arm explicitly handles empty args (`None => q.clone()`). Either tighten the Rust side to a non-empty vec, or relax the CDDL to `[* IrFormula]`. Pick one when PR2 lands.
3. **Backend dispatch posture asymmetry.** Coq / SMT-LIB-generated / ir-symbolic / aarch64 use `unreachable!()` (panic). Lean and the SMT-LIB validator return `Err(CompileError::Internal(...))`. Both are loud-not-silent, but the inconsistency deserves a unified posture: I'd lean toward `Err` everywhere (a compiler backend reached by a malformed input panicking-out is a worse posture than refusing-out, especially when this same plumbing eventually carries attacker-influenced lifter output). Follow-up, not a blocker.
4. **No automated guard that the manual `IrFormula` extension stays in sync with `provekit-ir.cddl`.** The new `cddl_valid.rs` checks the grammar file parses; it does not check the Rust enum matches. The comment in `provekit-ir-types/src/lib.rs` warns that regenerating from codegen will clobber the manual arms. Inherent to the manual-extension pattern (already used for `Sort`). Worth a follow-up issue for a JSON-roundtrip test against a wire-shape fixture so a CDDL drift is caught.

## Why this merges

The evaluator is total, deterministic, schema-node-clean, and the test suite pins both the externally-observable formula shapes (synthesized `add` / `div`, Dijkstra `if`, `seq;if;skip` body) and the wire encoding of the two new node kinds (BLAKE3-512 CIDs). The grammar extension is conservative — an existing `IrFormula` is still an `IrFormula` — and every backend got a refusal arm. `provekit-walk` is genuinely unbuildable today, so its refactor lands dead-code-ready, which is the right call rather than reverting and re-doing later.

The wp definition itself — `pre ∧ Q[result := value_expr]` for value-ops, Dijkstra for `if`, `wp(s1, wp(s2, Q))` for seq, refuse for loops / unresolved calls — is textbook Dijkstra-Hoare WP. PR5's `∀Q. wp(concept:op, Q) ⇒ φ(wp(lang:op, Q))` discharge will sit on top of this and it'll hold.

Ship it. Land the nit on the PR-body language before PR2 ideally; the rest can ride along in follow-up issues.

— Opus 4.7 reviewing for *Supra omnia, rectum*

PR #628, opened 2026-05-12
feat(transport): gap-memento schemas + generator emits (PR 1/N, issue #66)

TSavo:
## Summary

Adds the first-generation gap-memento schemas and Python generator output:

- `TransportGapMemento`, `PartialMorphismMemento`, `LossyMorphismMemento` structs in `libprovekit`
- `LossRecord` with named optional dimensions (JCS-safe; see below)
- `HomomorphismObligation` and `LossSeverity` helper types
- Python mint script (`mint_language_morphisms.py`) generates concept-shape mementos; `mint.sh` wraps the full regen cycle
- `cids.tsv` catalogs all minted CIDs

## JCS correctness -- bug-class promotion (re-review round 2)

The prior review cycle found a `Option<Value> + skip_serializing_if` JCS violation in `TransportGapMemento.signature`.  That fix has been promoted **class-wide** across all three memento types:

| Type | `signature` field type | skip_serializing_if |
|---|---|---|
| `TransportGapMemento` | `Value` (was `Option<Value>`) | removed |
| `PartialMorphismMemento` | `Value` (was `Option<Value>`) | removed |
| `LossyMorphismMemento` | `Value` (was `Option<Value>`) | removed |

Unsigned mementos carry `signature: null` (a JSON `null` literal), matching the Python generator which always emits the key.

## LossRecord.extra removed (Finding 3)

`LossRecord` formerly had `extra: BTreeMap<String, Value>` with `#[serde(flatten)]`.  `serde(flatten)` emits flattened keys AFTER the parent struct's named fields in plain `serde_json` output, breaking lexicographic JCS key order at the struct boundary.  The field has been removed entirely.  No callers used it; `catalog/morphisms/` is empty.  Extension path for new dimensions: named `Option<Value>` fields following the successor-mint protocol (LSP §4.4).

## Lock-by-construction tests (Finding 2)

Four lock-tests now assert `serde_json::to_string == serializable_jcs` for a typed instance.  Each test covers one type:

- `transport_gap_top_level_field_order_is_jcs` (existing, extended)
- `partial_morphism_top_level_field_order_is_jcs` (new)
- `lossy_morphism_top_level_field_order_is_jcs` (new)
- `loss_record_top_level_field_order_is_jcs` (new)

Planted-regression verification: swapping two adjacent struct fields in each type causes **only that type's lock-test** to fail; the other three remain green.  3/3 types verified.

## mint.sh byte-stability

Two clean runs of `mint.sh` produce identical `cids.tsv` (md5 `994454e51f5936135f9a01ab7a048394`).

## Test counts

77 tests pass (36 unit + 41 integration).

## Deep re-re-review required

This PR has been through two rounds of changes-required.  The structural fixes above address all three findings from round 2, but the reviewer should verify:

1. Class-wide `signature: Value` promotion is consistent across all three struct definitions and their `::new()` constructors
2. `LossRecord` has no remaining `flatten`-attributed fields
3. All four lock-tests are present and would fail on a field-order regression
4. `cids.tsv` path update (pk-fix-628 -> pk-fix-628-v2) is noise-only (CIDs unchanged)


TSavo:
## Verdict: CHANGES-REQUIRED

Two substrate-correctness blockers (one of them is a hard JCS/CID-federation break), plus a rebase. The PR's positive work is real and visible (309 mementos, byte-stable `transport-gaps.md` hash matches the body, zero morphism-CID drift from main, 12/12 transport_gap tests green, generator hygiene cleaned up), but it ships Rust types that *cannot* reproduce the CIDs of the catalog files Python writes, and a `cids.tsv` that silently drops one of the 309 gap mementos on disk. Under *Supra omnia, rectum* both have to land before merge.

### Blocker 1 — Rust serde does not emit JCS-canonical bytes (CID federation broken)

`libprovekit/src/transport_gap.rs` declares struct fields in *spec-listing* order, but `serde_json` emits structs in **field-declaration order**, not lexicographic. JCS is lexicographic. So Rust emits non-JCS bytes whenever optional fields land between their lexicographic neighbors. The comment at lines 12–15 ("serde_json with BTreeMap emits keys in lexicographic order, which is the JCS canonical order") is misleading: BTreeMap sorts; **structs don't**.

Empirically: for `TransportGapMemento`, the Rust struct places `reason_note` at position 11 (after `target_op_cid`), but JCS lexicographic puts it at position 5 (right after `reason`). Every on-disk gap memento has `reason_note` populated (`mint_gap_memento` line 552 sets it unconditionally), so this fires on **all 309 mementos**, not a corner case.

I sampled 5 random gap mementos and re-serialized through the Rust struct order. **5/5 produced different bytes from the JCS bytes on disk.** First differing byte position 139–153 in each case; the byte sequence `,"reason_note":"..."` appears where Rust emits `,"resolution_options":[...`. Different bytes → different BLAKE3-512 → different CID. No Rust consumer can verify a catalog gap memento against its disk CID right now.

Same field-order bug in `ResolutionOption` (`option_kind, precondition, loss, loss_severity, split_targets, respec_target_to, representation_map_delta, partial_morphism_cid, lossy_morphism_cid, dual_view_cid, tradeoff, status` — JCS-sorted should put `dual_view_cid` first and `tradeoff`/`status` at the end), and `GapReason` (`formal_sorts_delta, pre_delta, post_delta, effects_delta, wp_rule_delta, divergent_tag, source_supported` — JCS-sorted should start with `divergent_tag`). `PartialMorphismMemento` and `LossyMorphismMemento` field orders are coincidentally lexicographic and OK.

The 12 transport_gap tests pass because they roundtrip Rust↔Rust, where the field order is internally consistent. **None of them pins a BLAKE3-512 against a real on-disk fixture**, which is exactly the gate that would have caught this. The "12/12 green" claim is true but unrelated to wire-format correctness.

**Fix options:** (a) reorder struct fields lexicographically (mechanical, but fragile — depends on every future contributor knowing the rule); (b) serialize via an explicit `BTreeMap<String, Value>` and let serde sort; (c) add a canonicalization pass in libprovekit that sorts JSON keys before hashing; (d) at minimum, add at least one test per memento type that pins the BLAKE3-512 of a Python-emitted on-disk fixture and fails on order drift. (a)+(d) is the smallest viable fix; (c) is the right long-term shape and would also catch the inner `formal_sorts_delta`'s `{args, kind, name}` inversion vs JCS `{kind, name, args}` that I saw in the byte diff (probably in `provekit-ir-types`, not this PR — flag for follow-up).

### Blocker 2 — `cids.tsv` is missing one gap memento (308 of 309)

309 `.json` files in `catalog/gaps/`, 308 `gap` rows in `cids.tsv` after `mint.sh`. The orphan is `gap_rust_member_to_concept_member` — `rust` has two entries in `SPEC_CANDIDATES["member"]` (`op_field.spec.json` and `op_member.spec.json`), and both produce gap mementos with the **same `fn_name`** (`gap:rust:member:to:concept:member`) but **different CIDs** (different `formal_sorts_delta`). Both are written to disk; `append_cids` at line 791–795 dedupes by `(kind, name)` and silently drops the second.

Net: substrate-completeness violation — a content-addressable thing exists on disk that the index says doesn't.

**Fix options:** (a) disambiguate `fn_name` per-source-spec (e.g. `gap:rust:member@op_field:to:concept:member`); (b) dedupe `append_cids` by `(kind, cid)` instead of `(kind, name)` (lets two same-name different-CID rows coexist, which is actually what the substrate allows); (c) the candidate loop picks one — but it should record *why* the others were skipped, and the "missing-source-op" reason emitted by the unpicked candidates is the same reason a gap memento exists at all, so just emit one. (b) is the cheapest and most correct; (a) is the cleanest. Either way, run `mint.sh` twice and assert `wc -l catalog/gaps/*.json == grep -c '^gap' cids.tsv` as a post-condition. Cheap invariant.

### Other findings (not blockers)

- **Rebase needed.** PR is 3 commits behind main and CONFLICTING. #620 (wp-as-formula) and #622 (spec prose-coherence pass to multidimensional loss-record) landed after the branch point. Mechanical rebase — the PR's CDDL already implements the post-#622 multidimensional `loss-record` shape correctly, so the spec text just catches up.
- **Spec doesn't authorize the new keys.** `no-such-concept-op` and `unspecified` (gap-kind), `structural_divergence` (loss-dimension), `extend-concept-hub` (resolution-option-kind) are added to the CDDL but **none of them appear in the merged spec on main**. The CDDL header at lines 5–6 says "Source of truth: protocol/specs/2026-05-14-...". Either amend the spec in this PR (preferred — substrate additions should track spec, not lead it; "loudly bounded" applies here too), or remove them and defer to PR2/N. Governance call.
- **No `no-such-concept-op` emissions** in the 309-memento sweep, despite the classifier being wired. The PR body acknowledges this ("zero cases were found in the current 10-language sweep"). The task description expected `java:new`/`java:throw` to land as `no-such-concept-op`; instead `java:new` → `unspecified`, `java:throw` → `sort-mismatch` (because java does *have* a throw op, but its sort doesn't match the hub's). The classifier behavior is defensible — `no-such-concept-op` is for "language has an op the hub does not model at all," and the hub *does* have `concept:new` and `concept:throw` — but then the substrate sweep doesn't exercise the new gap kind at all, which means it's a CDDL extension with zero on-disk witnesses. Flag for PR2.
- **PR body cites issue #66** ("gap-memento sub-project (issue #66)"). #66 is the Ruby kit. Wrong link — probably meant a different issue or a manually-typed reference. Cosmetic.
- **Worktree paths leak into `cids.tsv`** (pre-existing, not introduced here). Same condition exists on main (paths have `pk-shrink-hub-2` baked in). Cleanup commit `255bd09b` fixed 27 cross-language-port report files but left `cids.tsv` per-worktree-pathful. Not a blocker for PR1, but for a content-addressed substrate this should not exist — track as a separate hygiene issue.
- **`reason_note` in the spec.** The spec line 65 says "OPTIONAL prose; non-load-bearing; MUST be omitted (not null) when absent." The CDDL correctly uses `?` (optional, not nullable) at line 161, and the Rust side uses `Option<String>` with `skip_serializing_if = Option::is_none`. But every emitted gap memento sets `reason_note` to the `reason_text` string — so in practice it's always present. Mention because the empirical "reason_note appears at position 5" finding depends on this being populated; if a future PR ever emits unset `reason_note`, the JCS-order bug becomes silent-then-loud.

### What's solid

- **309 gap mementos on disk** with the kind distribution the PR body claims (sort-mismatch 164, missing-source-op 96, wp-rule-mismatch 22, effect-mismatch 11, polymorphic-source-op 8, unspecified 3, arity-shape-mismatch 3, divergent-semantics 2). Per-language distribution sensible.
- **`transport-gaps.md` byte-stability hash reproduces** (`e1c8670a3c85e803097c4dac1b7a24205621cb5bcfd44826382f79a89014fd90`) after a fresh `mint.sh` run on this reviewer's worktree. The "generated from on-disk mementos via `_load_gap_mementos_from_disk`" claim holds.
- **No morphism-CID drift.** 156 morphisms on main, 156 on PR, byte-identical CIDs. The gap PR shouldn't change morphism shapes, and it doesn't.
- **`cargo test -p libprovekit --lib transport_gap`: 12/12 pass.** `cargo test -p provekit-ir-types`: 8/8 pass.
- **Generator hygiene.** The four `mint_gap_memento` call sites (lines 969, 979, 1007, 1056) each pass through `classify_gap_kind` with a meaningful `reason_text`, and `_default_resolution_options_for` returns a populated `resolution_options` list (with `tradeoff` and a default `status: "recommended"`) for every gap kind. No bare TODOs in emitted mementos.
- **Successor-mint semantics for `structural_divergence`.** `Option<Value>` with `skip_serializing_if = Option::is_none` correctly emits nothing when absent; the test `loss_record_successor_mint_absent_reads_as_none` proves absent = None. No retroactive CID impact on pre-PR1 mementos (though none exist on main yet, so this is a "ready when needed" claim, not an empirical one). Spec needs to declare it (see governance bullet above).

### Verdict reasoning

The substrate's first axiom is *Supra omnia, rectum*, and that includes "the CIDs the substrate writes must be the CIDs the substrate's verifiers compute." Right now, a Rust verifier loading any of the 309 gap mementos and re-canonicalizing through this PR's types would compute a different CID than the disk says. That's the failure mode the framework exists to prevent. Blocker 1 has to land before merge. Blocker 2 is a smaller violation but in the same class (substrate completeness). Spec coverage for the new keys is governance, not correctness, but worth aligning before this becomes the load-bearing schema for transport+migrate.

The shape is right. The decisions about gap-kind classification, the multidimensional loss-record, the resolution-option menu, the back-pointer fields, the `_load_gap_mementos_from_disk` source-of-truth pattern, the JCS-stable header in `transport-gaps.md` — all correct. Fix the two blockers, pin one BLAKE3-512 per memento type, rebase, and this is ready.

TSavo (commented):
## Adversarial re-review (post-fix) — verdict: CHANGES-REQUIRED

(Posted as COMMENT, not REQUEST_CHANGES, because the PR is from this account. Verdict semantics unchanged.)

Original blockers on `TransportGapMemento` are genuinely fixed. Empirically verified. But the same bug-class is partially repaired: the fix was applied only to `TransportGapMemento`, not to `PartialMorphismMemento`/`LossyMorphismMemento`, which live in the same file, are intended for federation, and carry the identical fragile pattern. PR 1/N is the right time to fix it — fixtures don't exist yet, but the structs do.

### Findings

**1. Same Blocker-1 pattern unrepaired in two sibling structs.**
`PartialMorphismMemento.signature` (line 305) and `LossyMorphismMemento.signature` (line 368) still declare `#[serde(skip_serializing_if = "Option::is_none")] pub signature: Option<Value>`. Python's `mint_language_morphisms.py:1044` writes `"signature": None` unconditionally for these too (by symmetry with the gap path at :555). When PR 2/N starts emitting Partial/Lossy mementos, Rust will omit the key, Python will emit `null`, and CIDs will diverge — the exact federation-killing failure the gap fix repaired. Fix is ~3 lines per struct (type → `Value`, drop the attribute).

**2. Lock-by-construction test covers only one of three types.**
`transport_gap_top_level_field_order_is_jcs` is excellent and demonstrably catches declaration-order drift on `TransportGapMemento` (I planted a `schema_version ↔ signature` swap → test FAILED, as designed). But planting an analogous regression on `PartialMorphismMemento` (`fn_name ↔ source_contract_cid`, both `String` so it compiles cleanly) → **74/74 tests still pass**. The cid_pinning fixture tests don't help: they route through the JCS canonicalizer (`crate::canonical::serializable_cid`), which sorts keys regardless of declaration order, so they catch the signature/skip-asymmetry bug class but not declaration-order drift. Need parallel `partial_morphism_top_level_field_order_is_jcs` + `lossy_morphism_top_level_field_order_is_jcs` tests. Lock-by-construction or it isn't locked.

**3. `LossRecord.extra: BTreeMap<String, Value>` with `#[serde(flatten)]` is a latent JCS hole.**
`#[serde(flatten)]` emits these keys AFTER the named fields in plain serde output — not in JCS position. The struct comment acknowledges this and asserts "no current catalog fixture uses extra dimensions (verified 2026-05-11)". True today. But there's no test that fails when a future fixture adds an extra dimension; the protection is by-inspection, not by-construction. Either add a runtime assertion in the writer path or a test that emits-with-extras and compares to JCS bytes.

### Cross-impl byte equality on `TransportGapMemento`: verified empirically

I added 5 additional adversarial cid_pinning tests for 5 random Python-emitted fixtures across csharp/rust/ruby/go/typescript (none used by the existing 3 tests) → all 5 deserialize and re-serialize to byte-identical CIDs. Reverting just the `signature: Value` fix (back to `Option<Value> + skip_serializing_if`) → 3/3 cid_pinning tests fail with explicit CID mismatch messages. Both protective layers fire as designed for `TransportGapMemento`.

### What is green

- 5 declared structs (`TransportGapMemento`, `GapReason`, `ResolutionOption`, `PartialMorphismMemento`, `LossyMorphismMemento`) have fields in JCS-alphabetic declaration order; `LossRecord` named fields are too, with the `extra` caveat above.
- `append_cids` dedup is `(kind, name, cid)` at `mint_language_morphisms.py:799`. Both `rust:member` rows present in `cids.tsv` with distinct CIDs.
- `cids.tsv` gap-kind row count = 293 = `ls catalog/gaps/ | wc -l` exactly.
- `mint.sh` twice → identical `sha256(cids.tsv) = ed196d8895f2…b09070f`.
- `cargo test -p libprovekit` → 74/74 pass.
- Source-of-truth diff vs `origin/main` is 4 files (Rust × 2, Python × 1, CDDL × 1); rest is generated artifacts.

### Out-of-scope nit (one line)

`cids.tsv` rows carry absolute paths from a different worktree (`/Users/tsavo/provekit-worktrees/pk-fix-628/...`). The mint-relative-paths fix (`6d796ca7`) is on `fix/mint-cidstsv-relative-paths`, not on this branch and not on main. Doesn't block this PR, but it means the per-machine byte-stability claim isn't a cross-contributor byte-stability claim.

### Verdict: CHANGES-REQUIRED

Two-line ask: extend the Blocker-1 fix to `PartialMorphismMemento.signature` + `LossyMorphismMemento.signature`, and add the two parallel `*_top_level_field_order_is_jcs` tests. ~10 LOC of Rust, lock-by-construction or re-discover the bug in PR 2/N.

TSavo (commented):
**Third deep review — VERDICT: MERGE-WITH-NITS**

The class-wide promotion blocker from review #2 is resolved. Verified empirically.

### What v2 fixed (verified)
1. **`PartialMorphismMemento.signature` and `LossyMorphismMemento.signature`** switched from `Option<Value> + skip_serializing_if` to plain `Value`. Always emits, matching Python's `"signature": None` byte-for-byte. JCS-alphabetic placement (between `schema_version` and `source_contract_cid`) verified by lex compare.
2. **Three new lock-tests** added: `partial_morphism_top_level_field_order_is_jcs`, `lossy_morphism_top_level_field_order_is_jcs`, `loss_record_top_level_field_order_is_jcs` (mirrors the TransportGap pattern).
3. **`LossRecord.extra: BTreeMap<String, Value>` with `#[serde(flatten)]` removed entirely.** Confirmed `catalog/morphisms/` doesn't exist (no live consumers); LSP §4.4 named-Option-field extension path documented in struct doc-comment.

### Planted-regression matrix (executed in worktree)
Each swap restored before the next:

| Swap | Lock-test that fails | Other 3 lock-tests | cid_pinning |
| --- | --- | --- | --- |
| `LossRecord`: `value_divergence` → first | `loss_record_top_level...` ONLY | green | green |
| `PartialMorphismMemento`: `kind` ↔ `homomorphism_obligation` | `partial_morphism_top_level...` ONLY | green | green |
| `LossyMorphismMemento`: `kind` ↔ `literal_map` | `lossy_morphism_top_level...` ONLY | green | green |
| `TransportGapMemento`: `kind` ↔ `reason` | `transport_gap_top_level...` ONLY | green | green |

Each gate is **type-specific** (catches its own type's drift) and the four together are **class-wide** (cover the bug-class, not the original narrow instance). 

### Other checks
- `cargo test -p libprovekit`: 77 tests pass (36+7+1+1+17+7+4+4+0).
- `cargo check --workspace`: clean (pre-existing warnings only).
- `mint.sh` byte-stable within worktree (md5 identical across consecutive runs; the path-only delta vs. the committed file is because `cids.tsv` records absolute paths and the committed file was minted in a different worktree — CIDs are byte-identical).
- First-round fixes intact: `append_cids` dedup-by-`(kind,name,cid)` (line 799), cross-impl Python-fixture pinning tests pass, TransportGap alphabetical reorder preserved.
- Diff focused; no collateral damage.

### Nit (non-blocking, follow-up)
The `LossRecord` lock-test populates only 2 of 5 Optional fields (`domain_narrowing`, `value_divergence`). Because all five fields have `skip_serializing_if = "Option::is_none"`, a declaration-order swap *entirely inside* the None-block (e.g. `effect_divergence` ↔ `structural_divergence` ↔ `ub_introduction`) is invisible to the lock-test as written: skip_serializing_if hides the moved field from both compact and JCS output, so the equality assertion still holds. The most likely human-introduced bug — reordering a *populated* field — is caught.

Tightening: populate all five fields in the fixture (cheap one-line change). Same construction-pattern applies to any future struct with `skip_serializing_if` Optional fields — "always populate every field in the lock-test fixture" should be the rule. Worth a follow-up, not a blocker for this PR.

### Verdict
**MERGE-WITH-NITS.** The blocker from review #2 (class-wide promotion incomplete) is closed. The remaining nit is a tightening of the test fixture, not a substrate-correctness issue.

TSavo:
Superseded by #696 (gap-mementos-1-impl-pr1) which landed at 3c85b5b5. Both PRs implemented the same TransportGapMemento + generator-emits scope. Closing to deduplicate the queue.

PR #619, opened 2026-05-12
concept-shapes: round-2 hub shrink — 124→143 morphisms (+19), 328→309 gaps (-19)

TSavo:
$(cat <<'EOF'
## Summary

Re-spec only-c11 (or only-c11+1) concept ops to the majority cross-language shape, unlocking discharge for languages blocked by encoding divergence rather than semantic divergence. Template: PR #614's `concept:ite` pattern (drop over-specific annotations, add slot renamings).

- **source-unit** (c11: bytes `Expr→String`, drop c11-specific `notes`; csharp/java: add `evaluation:unevaluated` on bytes; java: drop extra `slot_sort:Stmt` on `operational_term`; ruby/python: drop `notes`): `concept:source-unit` **1→8 morphisms** (c11 + csharp, python, ts, zig, ruby, php, java)
- **ushr** (ts+java: `Expr→Int` for `formal_sorts`, `arity`, `return_sort`, `result` — `>>>` is integer in every language that has it): `concept:ushr` **2 morphisms retained** (ts, java); sorts now correctly reflect integer semantics
- **decl** (c11: name `Expr→String`, drop `slot_sort:identifier`; add `renaming value→initializer` for java/zig; java: drop `slot_sort:literal` on name): `concept:decl` **1→4 morphisms** (c11, ts, zig, java)
- **while** (c11: drop `notes`; java: drop `slot_sort:Bool/Stmt` on slots; rust: add `arity_shape`, drop `notes`): `concept:while` **1→7 morphisms** (c11 + csharp, python, zig, ruby, java, rust)
- **do** (java: drop `slot_sort:Bool` on cond): `concept:do` **1→2 morphisms** (c11, java)
- **for** (java: drop `slot_sort:Bool` on cond; rust: add `arity_shape`): `concept:for` **2→4 morphisms** (c11, csharp, java, rust)

All changes are shape annotations only, no semantic changes. 2nd re-run byte-clean. All cargo tests pass (libprovekit, transport, desugar, transport_cli).

## Test plan

- [x] `PYTHONDONTWRITEBYTECODE=1 bash menagerie/concept-shapes/mint.sh` — 143 morphisms, 309 gaps
- [x] 2nd mint run — byte-clean (same counts)
- [x] `bash menagerie/cross-language-port/driver.sh` — runs clean, refusals expected for missing morphisms
- [x] 2nd driver run — receipt-cids.tsv MD5 stable
- [x] `cargo test -p libprovekit` — 4 passed
[...]

Issue #203, opened 2026-05-04
[kit:cpp][axis:A] Wire --kit=cpp to cpp-self-contracts surface (Side A wiring only)


TSavo:
Tracks the substrate-side conformance work for the cpp kit (Side A of the (A)/(B) split per #176).

## Background

The cpp kit's `mint_cpp_self_contracts` binary at `implementations/cpp/provekit-self-contracts/` already speaks the canonical `--rpc` lift-protocol and emits proof-envelope. The crypto substrate is complete (vendored BLAKE3 + OpenSSL Ed25519 + native CBOR). The contractSetCid in `cpp.json` is already content-meaningful.

The gap is wiring: `--kit=cpp` in the Rust CLI's `KIT_TABLE` resolves to surface `cpp`, but the self-contracts walker is on a different surface that doesn't get reached.

## Scope

- Add `cpp-self-contracts` lift surface manifest at `implementations/cpp/.provekit/lift/cpp-self-contracts/manifest.toml` pointing at `mint_cpp_self_contracts --rpc`
- Update `KIT_TABLE` in `implementations/rust/provekit-cli/src/cmd_mint.rs` so `("cpp", "cpp", "cpp", "cpp")` becomes `("cpp", "cpp", "cpp-self-contracts", "cpp")`
- Add a pinned-CID test in `mint_kit_integration.rs` mirroring the rust + go pattern (PR #180 + #183)

## Acceptance

- [ ] `make mint-cpp` produces the same content-meaningful contractSetCid as before (or stable new value if slab evolved)
- [ ] `provekit prove --kit=cpp` exits 0 in CI
- [ ] Pinned-CID test passes
- [ ] Same shape as PR #180 (go) + PR #183 (rust)

## See also

- Parent: #154
- (A)/(B) framing: #176
- Postmortem: 2026-05-04
- Tier 1 reference PRs: #180, #183

PR #633, opened 2026-05-12
feat(concept-shapes): wp_rule formulas on hub op specs (wp-as-formula PR 2/N)

TSavo:
## Summary

- Renames `post.wp` → `post.wp_note` on all 45 concept-shape specs (prose-only field, non-canonical per §1.1 of `protocol/specs/2026-05-13-wp-as-formula.md`).
- Adds machine-checkable `post.wp_rule` IrFormula to the 4 authored Stmt-sorted ops:
  - `concept:seq` → `apply(wp_first, apply(wp_second, Q))` (Dijkstra's sequential composition rule)
  - `concept:skip` → `Q` (identity)
  - `concept:conditional` → `(cond⇒wp_then_branch(Q)) ∧ (¬cond⇒wp_else_branch(Q))` (guarded conditional)
  - `concept:source-unit` → `apply(wp_operational_term, Q)` (opaque operational wrapper)
- All other Stmt-sorted ops (`while`, `for`, `do`, `break`, `continue`, loop variants) get `wp_note` only; `wp_rule` absent = spec gap per §1.2, handled by OpaqueLoop/OpaqueCall.
- All changes go in `mint_language_morphisms.py` (generator); spec files are regenerated via `mint.sh`. Fixes `_make_relaxed` and `diff_reason` to use `wp_note` key when comparing concept specs during structural-subsumption discharge.

## Implementation notes

The `apply_wp_rule()` function (new, called from `concept_spec_from_base()`) handles the rename and injection atomically. The `_make_relaxed()` fix pops both `wp` and `wp_note` before writing `wp_note`, so relaxed-spec CID comparisons remain sound across the PR2 migration boundary (language specs still use `wp`; concept specs now use `wp_note`).

## Test plan

- [x] `mint.sh`: 45 concept ops emitted, 0 with legacy `wp` key, 4 with `wp_rule`
- [x] `driver.sh`: transport refusals unchanged (pre-existing `c11:lt`/`c11:gt` gaps only)
- [x] `cargo test -p libprovekit` — 15 tests pass
- [x] `cargo test -p provekit-cli --test transport_cli` — 8 tests pass

Closes part of #61 (wp-as-formula sub-project, PR 2/N).


TSavo:

Reviewer ran a fresh worktree of `feat/wp-as-formula-2-hub-op-rules` (c118b9c2), executed `mint.sh` + `driver.sh`, diffed the result against `origin/main`, and tested a 3-way merge.

## The four authored `wp_rule`s are textbook-correct

Verified against §3 of `protocol/specs/2026-05-13-wp-as-formula.md`:

- `concept:seq` — formals `first`, `second` (both Stmt). Rule: `apply(wp_first, apply(wp_second, Q))`. Exact Dijkstra sequential composition. Slot names match formals.
- `concept:skip` — return Stmt, no Stmt formals. Rule: `Q`. Identity, correct.
- `concept:conditional` — formals `cond` (Bool), `then_branch` (Stmt), `else_branch` (Stmt). Rule: `(cond ⇒ wp_then_branch(Q)) ∧ (¬cond ⇒ wp_else_branch(Q))`. Dijkstra ITE-with-else, slot names match formals. Correct.
- `concept:source-unit` — formals `bytes` (String, unevaluated literal), `operational_term` (Stmt). Rule: `apply(wp_operational_term, Q)`. Pass-through to body, slot name matches. Correct.

No nits on the rules themselves.

## Rename hygiene clean

- 45 op-shape specs have `wp_note`, 0 have legacy `wp`, 4 have `wp_rule` (verified by grep).
- The 19 specs without `wp_note` are 12 morphism mementos + 7 pattern (`shape:*`) specs whose `post` is a logical formula not an `operation-contract` — out of scope, correct.
- The 10 Stmt-sorted ops without an authored `wp_rule` (`assign`, `break`, `call`, `continue`, `decl`, `do`, `for`, `return`, `throw`, `while`) are allowable per spec §1.2 + §4 coexistence window. Note as follow-up for PR3+; not a blocker on this PR.

## BLOCKER 1 — silent loss of ~46 minted morphisms

`scripts/mint_language_morphisms.py` `try_structural_subsumption` lines 499-506:

```python
after_wp = after_spec.get("post", {}).get("wp")          # the lang spec's "wp" value
concept_wp = ... .get("wp") or ... .get("wp_note")        # the concept spec's note value
...
wp_matches = after_wp == concept_wp
...
if pre_matches and wp_matches and effects_match:
    # Byte-equality should have caught this already; shouldn't reach here.
    return None
```

When the prose `wp` text of the language spec equals the prose `wp_note` text of the concept spec (the post-rename common case), `wp_matches == True`. But the byte-CIDs do NOT match because the JSON KEY differs (`wp` vs `wp_note`). The early-exit then returns `None`, and the morphism is not minted. The relaxation that would actually rewrite the key — `_make_relaxed` already pops both keys and writes `wp_note` — is correct, but unreachable on this code path.

Receipt: `try_structural_subsumption(c11_bit_and_normalized, concept_bitand)` returns `None` despite all fields matching modulo the key rename.

Net regression (counted from `cids.tsv` post-mint vs `origin/main:cids.tsv`):

- Total morphisms: **156 → 110** (-46) in cids.tsv; `mint.sh` reports `morphism_count 97`.
- `transport-gaps.md` rows: 358 → 404 (+46).
- Concrete losses: `c11_bit_and/bit_or/bit_xor/bit_not_to_bit*`, `c11_eq/ne/lt/le/gt/ge`, `c11_not`, `c11_cast/deref/member`, `c11_addr_of`, `c11_post_dec/pre_dec/pre_inc`, `c11_do`, `csharp/python/zig/ruby/php/java _seq_to_seq`, `csharp/zig/java _skip_to_skip`, `csharp/python/zig/ruby _if_to_conditional`, `typescript_ushr`, `csharp_new`.

This directly contradicts the PR description's "transport refusals unchanged (pre-existing c11:lt/c11:gt gaps only)." Under Supra omnia, rectum, silent loss of previously-discharged morphisms is a hard regression and exactly the failure mode #61-PR5's z3 refinement check must not inherit.

**Fix:** drop the early-exit at lines 504-506, OR set `relax_wp = True` whenever `after_spec.post` has key `"wp"` and `concept_spec.post` has key `"wp_note"` (the rename is the relaxation, even when the prose value matches). After the fix, re-mint and confirm the morphism count returns to ~156 and the gap-row delta vs main is only the expected rows from the wp_rule additions on the four authored ops.

## BLOCKER 2 — stale base, real merge conflicts

`gh pr view 633` reports `mergeStateStatus: BEHIND`. The merge base is `857698ec` (#619), which predates `a63f0d20` (#620, PR1) by 25 seconds. As a result the diff against `origin/main` appears to:

- delete `implementations/rust/libprovekit/src/wp.rs` (1088 lines) and `wp/tests.rs` (617 lines)
- delete `pub mod wp;` in `libprovekit/src/lib.rs`
- delete `SubstituteFormula` / `ApplyFormula` from `protocol/provekit-ir.cddl`
- revert `compose.rs` to a local-copy-of-substitute version (vs PR1's `pub use crate::wp::...`)

A 3-way merge against the older main reproduced cleanly and `cargo test -p libprovekit` was green (17 wp tests + everything else). But against current `origin/main` the merge now produces hard conflicts (cids.tsv, README.md, transport-gaps.md, 18+ transport-report.json files, catalog/index.json). The branch must be rebased onto current main and re-minted before this PR can land.

## BLOCKER 3 — test-plan claim is incorrect

The PR description states "driver.sh: transport refusals unchanged (pre-existing c11:lt/c11:gt gaps only)." Local `driver.sh` run on the PR head emits the same refusals at the driver layer, but the underlying `transport-gaps.md` has 46 more rows than main. Either fix the regression and re-run the test plan, or document the actual impact honestly.

## Nits (not blocking)

- **Absolute paths in canonical artifacts (pre-existing).** `cids.tsv` and every `transport-report.json` under `cross-language-port/artifacts/` embed the worktree's absolute path. Running `mint.sh`/`driver.sh` from a different worktree directory produces a dirty `git status` (paths only — CID column is identical). Fix in a separate PR by emitting repo-relative paths.
- **`concept_op_count 45` vs `morphism_count 97` discrepancy with main's 156** is the regression above; the count is mechanically truthful but warrants a status banner in the next PR description.

## Compile + tests

- 3-way merge of this PR onto an earlier main: `cargo test -p libprovekit` green, 17 wp tests pass.
- `cargo test -p provekit-cli --test transport_cli`: green (8 passed).
- `cargo test -p provekit-cli --test mint_kit_integration`: 3 pre-existing failures (cpp_kit/rust_kit/kits_with_real_contracts contractSetCid pinned-CID mismatches per #203). No new regressions from this PR.

## Verdict

**CHANGES-REQUIRED**. Two blockers to clear (the morphism regression + the rebase) and one descriptive correction. The four authored `wp_rule`s are correct; the rename pass is clean; the architectural shape of the PR is right. The discharge code path that was supposed to keep the existing morphisms minted during the migration window has an off-by-one in its conditional — exactly the kind of "looks right, breaks structure" failure mode that #61-PR5's solver-checked refinement must not inherit.

— review robot

TSavo (commented):
## Adversarial Re-Review of PR #633 — VERDICT: MERGE-WITH-NITS

**Carrying forward the CHANGES-REQUIRED from the prior review.** Both blockers are addressed; the fix is mechanically correct; the regression test is real (catches the OLD bug when reverted); tests pass; scope is clean. One numerical claim in the commit message is loose — see (6).

### Checks

**(1) `wp_byte_same` definition.** `(after_wp_key == concept_wp_key) and (after_wp == concept_wp)` — requires both key equality AND value equality. Python `==` on the values is structural dict equality (key-order-insensitive) for formulas, string equality for prose. Either is correct; no JSON-canonicalization required because both sides are produced by the same code paths (no key-order drift). PASS.

**(2) `after_wp` read fix.** Lines 497-502: `after_wp_key = "wp_note" if "wp_note" in after_post else ("wp" if "wp" in after_post else None)` — prefers `wp_note`, falls back to `wp`, gracefully handles neither. Symmetric for concept. PASS.

**(3) 46 named morphisms back.** Spot-checked all 14 c11 ops (`bit_and/or/xor/not`, `eq/ne/lt/le/gt/ge/not`, `cast`, `deref`, `member`) plus `csharp_new`, `go_eq/ge/gt/le/lt`, `typescript_ushr`, c11 `addr_of/post_dec/pre_dec/pre_inc/do` — all present once in the refreshed `cids.tsv`. PASS.

**(4) Regression test soundness.** 4 cases in `test_mint_language_morphisms.py`:
- Cross-key same value: `concept.post.wp_note = "X"`, `lang.post.wp = "X"` → must NOT return None. The exact regression. PASS.
- Both `wp_note`, same value, full byte-match → returns None (byte-equality should have fired upstream). PASS.
- Cross-key DIFFERENT value → still discharges via wp-abstraction (wp is documentation). PASS.
- Pre-weakening path with no wp on either side → still discharges. PASS (guards no breakage).
- Empirical: reverted `mint_language_morphisms.py` to PR2 contents → ran the test → **2 failures** observed (`test_wp_note_vs_wp_same_value_produces_morphism` + `test_wp_note_vs_wp_note_same_value_returns_none`). The test catches the bug. PASS.

**(5) `mint.sh` wires test in.** Line 5: `python3 -m unittest "$BASE/scripts/test_mint_language_morphisms.py"` runs before the discharge/mint pipeline. CI-gated. PASS.

**(6) Morphism count discrepancy — answered.**

The commit message says "156 → 110 → 151" measured against `cids.tsv` line count. This is **wrong yardstick**: `cids.tsv` is append-only and inherits stale rows from prior worktrees (the PR2 cids.tsv references paths like `pk-java-typed-sorts` for csharp/python/zig/ruby/php/java seq/skip/source-unit morphisms — leftover entries that the current mint does NOT regenerate).

Ground truth from `mint_language_morphisms.py` stdout (re-run in fresh worktrees):
- **PR2 (38f425b6) buggy state: `morphism_count\t113`**
- **PR3 fix (c7485649): `morphism_count\t138`**
- Recovery: **+25 morphisms**, not +46.

The +46 figure in the commit msg + brief is an artefact of `cids.tsv` accumulation. The real impact is +25 fresh-minted morphisms; the 21 morphisms still missing vs. PR2 cids.tsv are the four hub ops (`seq`, `skip`, `conditional`, `source-unit`) across non-c11/non-rust languages, where the concept side now carries `post.wp_rule` (a formula dict) but lang specs do not. These are GENUINE structural gaps (per `transport-gaps.md`: "operation-contract mismatch", "formal sort mismatch", "effect signature mismatch") and the wp-abstraction relaxation correctly refuses to discharge them. Closing those gaps is future work (PR3+ of #61 wires `wp_rule` onto lang lifters).

**Nit**: commit message numbers should be restated from mint stdout, not `cids.tsv` line count. Not a blocker.

**(7) Byte-stability.** Two consecutive `mint.sh` runs produced identical sha (`913582b091cc4e7554aa76820e88ee0599a90896b925b98f888f74cea9cafe1f`). Committed file sha differs (`e820d91b3ca86dbd40b6eda6f4c6e817d940b6583a6f52c0f1820c46b8a81a2c`) — the only differences are the absolute worktree path strings in the path column (`pk-fix-633` vs `pk-re-review-633`); all CIDs (hashes) are byte-identical. Known cids.tsv idiosyncrasy, not a real diff. PASS. (`driver.sh` is in `cross-language-port/`, byte-stable across regen — its CIDs unchanged in the diff.)

**(8) Four hub-op `wp_rule` blocks untouched.** `git diff 38f425b6..c7485649 -- scripts/mint_language_morphisms.py | grep -E '^[+-].*wp_rule|wp_first|wp_then_branch|wp_operational'` returns empty. The authored wp_rule blocks for `seq`/`skip`/`conditional`/`source-unit` are byte-identical to PR2. PASS.

**(9) No collateral damage.** `git diff 38f425b6..c7485649 --stat -- ':!menagerie/concept-shapes' ':!menagerie/cross-language-port'` returns empty — zero files touched outside the two expected directories. No sort widening, no unrelated deletes. PASS.

**(10) `cargo test -p libprovekit -p provekit-cli --test transport_cli`** — 8 tests, all pass:
```
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.29s
```
PASS.

### Verdict

**MERGE-WITH-NITS.**

The fix is surgical (≤30 lines of substantive code change), the regression test is genuine (verified to fail on the reverted code), no scope creep, all CI gates green, cargo tests pass, and the substrate continues to refuse impossible discharges (the 21 unrecovered morphisms are real gaps awaiting PR3+ of #61, not regressions).

**Nit (non-blocking, suggested follow-up)**: the commit message and PR description reference `cids.tsv` line counts (156 / 110 / 151 / +46) which conflate fresh-minted morphisms with stale append-only entries from prior worktrees. The accurate, reproducible numbers are from `mint_language_morphisms.py` stdout: PR2-buggy=113, PR3-fix=138, recovery=+25. Restate in a future commit body or addendum so the substrate's count history is auditable; not a merge blocker.

T Savo

TSavo:
## Re-review verdict: CHANGES-REQUIRED

Re-reviewed at `e0a26193`, worktree from `origin/feat/wp-as-formula-2-hub-op-rules`. Both prior blockers from CHANGES-REQUIRED-1 are empirically closed, and a new soundness gap is structurally open. Detail below.

### B1 (silent loss of ~46 morphisms): closed empirically

The fix took a third path (re-inject concept's `wp_rule` into the relaxed copy) rather than Option A or B as originally proposed. Verified:

- `grep -c '^morphism\t' menagerie/concept-shapes/cids.tsv` → 174 on PR branch, 174 on `origin/main`. Matches exactly.
- Transport gap row count: 340 on both. Matches exactly.
- Two manual mint runs (mint.sh `python3 -m unittest <path>` syntax is broken on Python 3.14 — pre-existing, unrelated to this PR; ran the seven mint scripts directly): CID-only sha256 of cids.tsv identical across runs (`897cc528…3059a`). Path field in cids.tsv embeds an absolute worktree path (pre-existing portability defect, not introduced here); the CID column itself is byte-stable.
- `cargo test -p libprovekit`: 4/4 (and the other subsuites also clean) — pass.
- `cargo test -p provekit-cli --test transport_cli`: 8/8 — pass.
- `python3 -m unittest test_mint_language_morphisms` (4 regression tests): pass.

### B2 (stale base): closed

`git merge-base HEAD origin/main` = `ac82e1932f7e68b8e729cd7faa0d472bf08f1b31`, the exact target. 4 commits ahead / 6 behind; the 6 behind are subsequent main commits, not the gap from PR1. Stale-base concern resolved.

### Soundness gap: the wp_rule re-injection is unsound by construction

The fix in `mint_language_morphisms.py:591-624` (`_make_relaxed`) and 648-655 (fast path) strips the language spec's `wp_rule` and replaces it with the concept spec's `wp_rule` *unconditionally* — there is no equality check on `wp_rule` anywhere in `try_structural_subsumption`.

Concretely:

1. `concept_spec_from_base` (line 322-341) authors `concept.wp_rule` from the `op_def` table (lines 140-186) — INDEPENDENTLY from whatever the language spec carries. The base lang spec's `wp_rule` is explicitly stripped at line 338 before `apply_wp_rule` re-injects from `op_def`. So concept's `wp_rule` is not derived from the lang spec being matched.
2. Language specs DO carry `wp_rule` today: `c11-language-signature/specs/op_{seq,skip,if,source_unit}.spec.json` and `rust-language-signature/specs/op_{seq,skip,if,source_unit}.spec.json` (14 hits across language-signature dirs).
3. The fast path at line 648 fires when `after_has_wp_rule && pre_matches && wp_matches && effects_match`. It does NOT check `after_post["wp_rule"] == concept_post["wp_rule"]`. `_make_relaxed` then pops lang's `wp_rule` and substitutes concept's.
4. Result: if a lang op declares `wp_rule = F_lang` and concept declares `wp_rule = F_concept` with `F_lang ≠ F_concept`, the morphism is discharged as `structural-wp-abstraction` (or `-wp-rule-abstraction`) anyway. The morphism asserts lang ⊑ concept, but the lang spec's own contract claim disagrees with concept's. That's a soundness false-positive.

The justification in the comment (lines 615-617) — "any lang op that structurally subsumes the concept shape is obligated to honour the concept's wp_rule" — is circular. Morphism *minting* is what would establish that obligation; the discharge cannot assume it.

**Empirical mitigation today:** I diffed the 7 reachable comparisons (c11 + rust, four authored ops each; rust source-unit is absent in current corpus). All 7 lang/concept `wp_rule` pairs agree byte-for-byte. So no false-positive morphism exists in the current catalog. But soundness here depends on authoring discipline of lang spec files, not on the gate. Per Supra omnia, rectum, the substrate's safety must live in the gate.

**The fix is small.** The slow path (`_make_relaxed` line 619-621) has the same hole as the fast path — it's the underlying unsound primitive, not the fast path itself. So a fast-path-only guard would not close the gap.

### Required changes

1. **Add `wp_rule`-equality gate.** In `try_structural_subsumption`, before either path that calls `_make_relaxed` with wp_rule substitution: if `"wp_rule" in after_post && "wp_rule" in concept_post` and the two values differ, return None. Don't substitute. (Alternative: hoist the gate into `_make_relaxed` itself and have it return None / raise on disagreement; cleanest implementation choice.)
2. **Add regression test** `test_wp_rule_disagreement_returns_none` to `test_mint_language_morphisms.py` — both specs carry `wp_rule`, values differ, all else matches, assert `try_structural_subsumption(...) is None`. Mirrors the existing four tests' style.
3. **Re-verify counts** post-fix: morphism count must remain 174, gap rows 340 (since all 7 current pairs agree). If either drifts, that's a real existing false-positive surfaced by the gate — surface it explicitly, don't silently rebase.

### Nit (not blocking, addressable in same revision)

Method name `structural-wp-abstraction` is overloaded after this PR: original meaning ("post.wp is prose, abstract it") vs new meaning ("post.wp_rule was substituted from concept-side into lang-side relaxed copy"). Downstream tooling reading the morphism catalog (e.g., `tools/concept-library-completeness-probe.py:520`) will conflate these. Suggest renaming the wp_rule-substitution case to `structural-wp-rule-substitution` (or similar) so the receipt's method string is unambiguous about which transformation was applied. Receipt CIDs change — re-mint regenerates them, no migration needed.

### Why CHANGES-REQUIRED rather than MERGE-WITH-NITS

PR3 (per the wp-as-formula arc) is exactly where lifters start generating wp_rule on the lang side from real source code. That's the failure mode this gap allows: a lifter producing a `wp_rule` that disagrees with the authored concept `wp_rule` would silently false-positive a morphism. Landing this PR with the gate missing means the next arc step inherits an unsound discharge primitive. Cheaper to add the gate now (one branch, one test) than to backtrack across PR3 receipts later.

Supra omnia, rectum.

— T Savo (review delegate)

TSavo:
Fixed in 348ece62 (force-pushed onto `feat/wp-as-formula-2-hub-op-rules`). Resolution per review issue:

| Issue | Resolution |
|---|---|
| **Soundness gap: unconditional wp_rule substitution** | Added equality gate in `try_structural_subsumption` immediately after `after_has_wp_rule` is set, before any fast-path or `_make_relaxed` call. Condition: `if after_wp_rule is not None and concept_wp_rule is not None and after_wp_rule != concept_wp_rule: return None`. Python `!=` on decoded dicts is structurally equivalent to JCS-canonical byte comparison (no floats in wp_rule; order-independent on keys). Gate is silent when either side lacks wp_rule -- only disagrees-with-both-present triggers refusal. |
| **Method-name overload** | Renamed the wp_rule-substitution fast-path discharge method from `"structural-wp-rule-abstraction"` to **`"structural-wp-rule-substitution"`**. Updated in: the fast-path return, both docstring blocks. Existing `"structural-wp-abstraction"` receipts are unaffected -- that name covers the prose-wp / key-rename path and was not overloaded. |
| **Regression tests** | Added `WpRuleGateTests` class in `test_mint_language_morphisms.py` with 4 tests: `test_wp_rule_disagreement_returns_none` (core gate), `test_wp_rule_agreement_still_discharges` (control; asserts method == `"structural-wp-rule-substitution"`), `test_only_lang_has_wp_rule_still_discharges`, `test_only_concept_has_wp_rule_gate_is_silent`. All 8 Python tests pass. |
| **Counts unchanged** | Two successive `./mint.sh` runs: **174 morphism rows** in cids.tsv, **340 rows** in transport-gaps.md. Byte-stable. All 7 current lang-vs-concept wp_rule pairs agree so gate does not fire. |
| **Cargo tests** | `cargo test -p libprovekit` 4/4, `cargo test -p provekit-cli --test transport_cli` 8/8. |

**Note on remaining unconditional path:** `_make_relaxed` still substitutes concept's wp_rule unconditionally when only the lang side carries wp_rule (concept makes no claim). That is sound -- concept's non-claim is re-injected as nothing, CID comparison decides. Not widened per task constraint.

TSavo:
## Round-3 deep re-review — MERGE-WITH-NITS

Adversarial verification of commit `348ece62` against round-2 CHANGES-REQUIRED.

### Equality gate — verified
- Placement: line 576, BEFORE `_make_relaxed` and BEFORE the fast-path early-return at line 600. Correct.
- Semantics: `after_wp_rule is not None and concept_wp_rule is not None and after_wp_rule != concept_wp_rule` → returns None.
- Equality: Python dict `!=` on AST-shaped wp_rule is value-equality, order-independent for dicts, order-sensitive for lists. Equivalent to JCS-byte-equality on this data. Not a latent bug.
- Single-sided cases (only-lang, only-concept, neither) all silently fall through; downstream behaviour is sound (see below).

### Method rename — verified
- `structural-wp-rule-substitution` used at one call site (line 668).
- No stale references to old name `structural-wp-rule-abstraction` in code, tests, JSON, MD, or TSV.
- `structural-wp-abstraction` (the unrelated notes-tolerant case in option/pair/result mints) untouched.
- `concept-library-completeness-probe.py` `DISCHARGE_SHORT` dict doesn't map the new name, but falls through to identity display — cosmetic only.

### Tests load-bearing — verified
Reverted `mint_language_morphisms.py` to pre-fix HEAD `82fc45da`:
- `test_wp_rule_disagreement_returns_none` → FAILS (returns `('structural-wp-rule-abstraction', False, False)` instead of None)
- `test_wp_rule_agreement_still_discharges` → FAILS (method is `structural-wp-rule-abstraction`, not `structural-wp-rule-substitution`)
- 2/4 pre-fix failures; the disagreement gate test is the load-bearing one.

### Byte-stability — verified
Two successive mint runs on the worktree:
- `cids.tsv`: byte-identical (kind/name/cid columns; absolute-path column is a working-directory artefact only)
- `transport-gaps.md`: byte-identical
- `grep -c "^morphism" cids.tsv` = 174
- `grep -c "^|" transport-gaps.md` = 340

### Cargo + Python tests — verified
- `python3 -m unittest test_mint_language_morphisms`: 8/8
- `cargo test -p libprovekit`: 4/4
- `cargo test -p provekit-cli --test transport_cli`: 8/8

### Soundness deep-dive

The "only-concept has wp_rule" case is the **common case** post-PR2, not a corner case. The four authored ops (seq/skip/conditional/source-unit) carry `wp_rule` only on the concept side; no lang lifter authors `wp_rule` yet (PR3 will). `_make_relaxed` re-injects concept's `wp_rule` into the relaxed lang copy and discharges via `structural-wp-abstraction`. Sound by the PR2/3 design intent: concept hub is the canonical source of truth for `wp_rule`; structural subsumption identifies "this lang op IS this shape"; the `wp_rule` inherits via the identification. The 27 morphisms commit `82fc45da` explicitly recovered are exactly this case.

The "only-lang has wp_rule" case: lang asserts `wp_rule` X but concept makes no claim; `_make_relaxed` strips lang's annotation, no concept `wp_rule` to inject; discharge via `structural-wp-rule-substitution`. Sound: concept makes no obligation, so dropping lang's bonus annotation can't violate one.

The "both agree" case: fast path discharges via `structural-wp-rule-substitution`. Sound.

The "both disagree" case: gate fires, returns None. Sound (refuses false positive).

The gate is forward-looking: it catches PR3-and-beyond conflicts where lifters start authoring `wp_rule` that disagrees with the hub. With current data (no lang lifter authoring wp_rule), the gate never fires in production runs, which is why byte-stable counts are preserved.

### Nits (non-blocking)

PR description is stale: describes the original PR2 work but doesn't mention the round-2 fix (concept `wp_rule` re-injection, +27 morphisms) or round-3 fix (equality gate + method rename). Commit messages cover the rounds; the PR body doesn't. Worth updating before merge so a reviewer reading the description sees the current shape.

### Verdict

**MERGE-WITH-NITS.** Soundness is improved monotonically: PR2 added the wp_rule machinery, round-2 fixed silent morphism dropping, round-3 added the equality gate that refuses false positives when lifters will eventually start authoring wp_rule. Supra omnia, rectum holds.

Recommend updating the PR description before merge (or noting it as follow-up).

PR #634, opened 2026-05-12
feat(abstraction-layer): ConceptAbstractionMemento + RealizationDesugaringMemento schemas + fixtures (PR 1/N)

TSavo:
## Summary

PR 1 of the abstraction-layer chain (issue #71). Implements the schema and fixture deliverables from spec `protocol/specs/2026-05-15-concept-hub-abstraction-layer.md` (merged PR #617). No hub catalog entries are added; `mint.sh` stays byte-clean.

### Deliverables

**(1) CDDL** -- `protocol/provekit-ir.cddl` extended with:
- `ConceptAbstractionMemento` + `AbstractionSlot`: the abstraction-tier hub node schema (spec §2.1)
- `RealizationDesugaringMemento`: a `DesugaringEquationMemento` in role `abstraction-realization` with `role`, `direction`, `target_lang`, `loss_record`, `discharge_receipt` additive fields (spec §2.2)
- `LossRecord` extended with `structural_divergence` as the fifth dimension (spec §2.4); successor-mint of #616 schema per LSP §4.4

Schema notes:
- `realizations: [* cid]` (zero-or-more) in PR1; PR2 tightens to `[+ cid]` via a successor schema mint with `refines = <PR1 CID>`
- `discharge_receipt: cid` optional in PR1; PR2 tightens to required
- `effects: []` is always serialized (never skipped) per CDDL spec

**(2) Rust types** -- `provekit-ir-types/src/lib.rs` manual-extension block:
- `LossRecord` as `BTreeMap<String, IrFormula>` (lexicographic key order = JCS canonical order)
- `AbstractionSlot`, `ConceptAbstractionMemento`, `RealizationPost`, `RealizationDesugaringMemento`
- All derive `Eq` for test assertions

**(3) `concept:dynamic-dispatch` fixture** (`ConceptAbstractionMemento`):
- Slots: `receiver`, `method_name`, `args` (variadic)
- Contract: Skolem `choice m. defined(m) ∧ wp_call(m, [receiver, ...args])` predicate (§2.3 wp_rule)
- `realizations: []` (empty for PR1)
- Pinned CID: `blake3-512:ca8ff4e631fe41885b37f897b42906e62e2b4079438ce6480885b5eb0f01207314fd38563591e82304aaa9c1b750e94a8baa24f5e9401cd2ed873da0cd0c90da`

**(4) `concept:double-dispatch` fixture** (`ConceptAbstractionMemento`):
- Slots: `receiver`, `secondary`, `method_name`, `args` (variadic)
- Contract: conjunction-of-guarded-clauses Skolem over (receiver, secondary) runtime-type pair
- `realizations: []`
- Pinned CID: `blake3-512:1cc86fe75917b11e41cb847ee5e5815d67e69f5d5c93f89c0fc54c273312060ae26c6170c8c2d9b22170fabbc627b5d614b3c39c486fcd9edc6e44433cff9909`

**(5) Three `RealizationDesugaringMemento` fixtures** -- double-dispatch x {c11, java, ruby}:

The three fixtures demonstrate the projection-distance law from spec §3.0:

| Target | `structural_divergence` | `domain_narrowing` | Pinned CID |
|--------|------------------------|-------------------|------------|
| `c11` (2D fn-ptr table) | heavy (true) | requires_static_2d_dispatch_table | `blake3-512:081b7f07...` |
| `java` (visitor pattern / invokeinterface) | mid (visitor_pattern_indirection) | absent | `blake3-512:8ea85ecd...` |
| `ruby` (case/respond_to chain) | near-zero (guarded_case_chain) | absent | `blake3-512:730a633f...` |

C is the unforgiving end: both `structural_divergence` AND `domain_narrowing` non-empty. Java is the middle: `structural_divergence` only (JVM dispatch is near-native). Ruby is the loose end: `structural_divergence` only, lighter (native `public_send`).

**(6) `structural_divergence` coordination**:
- Declared in CDDL and Rust
- Dimension name `effect_divergence` (not `resource_introduction`) matches the spec §2.4 text and the #66-impl-PR1 `LossRecord` implementation
- Successor-mint of #616 loss-record schema per LSP §4.4: absent dimension = formula ∅; existing mementos unaffected

### Tests

- `provekit-ir-types/tests/abstraction_layer_serde.rs`: 11 serde round-trip tests
- `provekit-claim-envelope/tests/abstraction_layer_cid_pin.rs`: 13 JCS CID-pinning + projection-distance law tests

All 24 new tests pass. Full workspace passes (no regressions).

### mint.sh clean

No new entries in `menagerie/concept-shapes/specs/` or `catalog/`. `cids.tsv` SHA256 unchanged: `2618b39e508112f4d351653eceb58feea121b3e78ecc5a450c824d67ec3e0e23`.

### Coordinates with

- **#66-impl-PR1** (`TransportGapMemento` / `PartialMorphismMemento` / `LossyMorphismMemento`): both PRs declare the same five loss-record dimensions (`value_divergence`, `ub_introduction`, `domain_narrowing`, `effect_divergence`, `structural_divergence`). This PR adds `structural_divergence`; #66-impl-PR1 holds the other four from #616. Neither PR modifies the other's types.
- **#617** (merged spec): this PR implements PR1 of §8.1 of that spec.

### What PR2 adds

- Tighten `realizations: [+ cid]` via a successor schema mint with `refines = <PR1 ConceptAbstractionMemento CID>`
- Tighten `discharge_receipt: cid` (required)
- Populate `realizations` with real `RealizationDesugaringMemento` CIDs from the catalog
- Mint real `MorphismDischargeReceipt` mementos for the three double-dispatch realizations

## Summary by CodeRabbit

* **New Features**
  * Abstraction layer added to IR and schema: concept abstractions, realization-desugaring equations, abstraction slots, and loss-records for richer contract/structural modeling.

* **Tests**
  * Extensive serialization and stability tests: canonical JSON/JCS round-trips, pinned-CID verification for fixtures, duplicate-key audit to ensure stable canonicalization, and loss-record shape/assertions across realizations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/634)

### DIFF
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1619,6 +1619,7 @@ dependencies = [
  "hex",
  "provekit-canonicalizer",
  "provekit-proof-envelope",
+ "serde",
  "serde_json",
  "thiserror",
 ]
--- a/implementations/rust/provekit-claim-envelope/Cargo.toml
+++ b/implementations/rust/provekit-claim-envelope/Cargo.toml
@@ -12,5 +12,6 @@ provekit-proof-envelope = { path = "../provekit-proof-envelope" }
 thiserror = { workspace = true }
 
 [dev-dependencies]
+serde = { workspace = true }
 serde_json = { workspace = true }
 hex = { workspace = true }
--- a/implementations/rust/provekit-claim-envelope/tests/abstraction_layer_cid_pin.rs
+++ b/implementations/rust/provekit-claim-envelope/tests/abstraction_layer_cid_pin.rs
@@ -16,14 +16,14 @@
 //     (B) concept:double-dispatch   -- ConceptAbstractionMemento
 //     (C) concept:double-dispatch->c11:2d-fn-ptr-table   -- RealizationDesugaringMemento
 //     (D) concept:double-dispatch->jvm:visitor-pattern   -- RealizationDesugaringMemento
-//     (E) concept:double-dispatch->ruby:case-respond_to  -- RealizationDesugaringMemento
+//     (E) concept:double-dispatch->ruby:case-type-tuple  -- RealizationDesugaringMemento
 //
 //   2. Round-trip parity: emit -> re-parse -> re-emit -> byte-identical JCS.
 //
 //   3. Projection-distance law held by the three RealizationDesugaringMementos:
-//      C (c11) carries both structural_divergence AND domain_narrowing (heavy).
-//      D (java) carries structural_divergence, less domain_narrowing (mid).
-//      E (ruby) carries only structural_divergence (near-zero).
+//      C (c11) carries structural_divergence + domain_narrowing + ub_introduction (heavy).
+//      D (java) carries structural_divergence + domain_narrowing (mid; visitor accept/visit).
+//      E (ruby) carries only structural_divergence (near-zero; case-match ≈ contract).
 //
 // To update pinned CIDs after a deliberate schema change:
 //   1. Remove the assertion and run the test to see the printed CID.
@@ -143,12 +143,26 @@ fn double_dispatch_jcs_round_trip() {
 // (C) RealizationDesugaringMemento: double-dispatch -> c11
 // ================================================================
 //
-// Heavy structural_divergence + domain_narrowing (C is the unforgiving end).
+// Heavy structural_divergence + domain_narrowing + ub_introduction (C is the unforgiving end).
+//
+// The rhs encodes the full 2D void*-table dispatch:
+//   ((fn_ptr)table[tag(receiver)][tag(secondary)])(receiver, secondary, args)
+// Steps: concept:member -> concept:index (dim 1) -> concept:index (dim 2)
+//        -> concept:cast (void* -> fn-ptr) -> concept:call
+//
+// Loss-record (3 distinct atomic claims, heaviest end of the bracket):
+//   structural_divergence: open_coded_vtable_replaces_single_op (2 index + 1 cast)
+//   domain_narrowing:      requires_static_2d_dispatch_table
+//   ub_introduction:       out_of_range_tag_is_ub
+//
+// JCS key order is alphabetical within each object (RFC 8785 §3.2.3).
+// TODO: re-sort the "Locked JCS key order" CDDL comments to match actual
+// canonicalizer output (alphabetical), not struct declaration order.
 
-const DD_C11_CANONICAL: &str = r#"{"direction":"left-to-right","effects":[],"fn_name":"concept:double-dispatch->c11:2d-fn-ptr-table","formal_sorts":["blake3-512:sort0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","blake3-512:sort0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","blake3-512:sort1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111","blake3-512:sort2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222"],"formals":["receiver","secondary","method_name","args"],"kind":"equation","loss_record":{"domain_narrowing":{"args":[{"name":"receiver","kind":"var"},{"name":"secondary","kind":"var"}],"kind":"atomic","name":"requires_static_2d_dispatch_table"},"structural_divergence":{"args":[],"kind":"atomic","name":"true"}},"post":{"lhs":{"args":[{"name":"receiver","kind":"var"},{"name":"secondary","kind":"var"},{"name":"method_name","kind":"var"},{"name":"args","kind":"var"}],"kind":"atomic","name":"concept:double-dispatch"},"rhs":{"args":[{"args":[{"args":[{"args":[{"kind":"ctor","name":"concept:member","args":[{"name":"receiver","kind":"var"},{"kind":"const","sort":{"kind":"primitive","name":"String"},"value":"dispatch_tbl"}]}],"kind":"ctor","name":"concept:index","args":[{"kind":"ctor","name":"concept:tag-of","args":[{"name":"receiver","kind":"var"}]}]}],"kind":"ctor","name":"concept:index","args":[{"kind":"ctor","name":"concept:tag-of","args":[{"name":"secondary","kind":"var"}]}]}],"kind":"atomic","name":"concept:call"}],"kind":"atomic","name":"concept:call"}},"pre":{"args":[{"name":"receiver","kind":"var"},{"name":"secondary","kind":"var"}],"kind":"atomic","name":"static_dispatch_table"},"role":"abstraction-realization","target_lang":"c11"}"#;
+const DD_C11_CANONICAL: &str = r#"{"direction":"left-to-right","effects":[],"fn_name":"concept:double-dispatch->c11:2d-fn-ptr-table","formal_sorts":["blake3-512:sort0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","blake3-512:sort0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","blake3-512:sort1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111","blake3-512:sort2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222"],"formals":["receiver","secondary","method_name","args"],"kind":"equation","loss_record":{"domain_narrowing":{"args":[{"kind":"var","name":"receiver"},{"kind":"var","name":"secondary"}],"kind":"atomic","name":"requires_static_2d_dispatch_table"},"structural_divergence":{"args":[{"args":[{"kind":"var","name":"receiver"}],"kind":"ctor","name":"concept:index"},{"args":[{"kind":"var","name":"secondary"}],"kind":"ctor","name":"concept:index"},{"args":[],"kind":"ctor","name":"concept:cast"}],"kind":"atomic","name":"open_coded_vtable_replaces_single_op"},"ub_introduction":{"args":[{"kind":"var","name":"receiver"},{"kind":"var","name":"secondary"}],"kind":"atomic","name":"out_of_range_tag_is_ub"}},"post":{"lhs":{"args":[{"kind":"var","name":"receiver"},{"kind":"var","name":"secondary"},{"kind":"var","name":"method_name"},{"kind":"var","name":"args"}],"kind":"atomic","name":"concept:double-dispatch"},"rhs":{"args":[{"args":[{"args":[{"args":[{"args":[{"kind":"var","name":"receiver"},{"kind":"const","sort":{"kind":"primitive","name":"String"},"value":"dispatch_tbl"}],"kind":"ctor","name":"concept:member"},{"args":[{"kind":"var","name":"receiver"}],"kind":"ctor","name":"concept:tag-of"}],"kind":"ctor","name":"concept:index"},{"args":[{"kind":"var","name":"secondary"}],"kind":"ctor","name":"concept:tag-of"}],"kind":"ctor","name":"concept:index"},{"kind":"const","sort":{"kind":"primitive","name":"String"},"value":"fn_ptr_2d"}],"kind":"ctor","name":"concept:cast"},{"kind":"var","name":"receiver"},{"kind":"var","name":"secondary"},{"kind":"var","name":"args"}],"kind":"atomic","name":"concept:call"}},"pre":{"args":[{"kind":"var","name":"receiver"},{"kind":"var","name":"secondary"}],"kind":"atomic","name":"static_dispatch_table"},"role":"abstraction-realization","target_lang":"c11"}"#;
 
 const DD_C11_CID: &str =
-    "blake3-512:081b7f07196d75c4a3796953113d73b3d8b4603ab3ffe1993015bb5cd09a33727122567876855875e62f4196d8c28e997e295bfc96f6b7a37462df4782bbb8a0";
+    "blake3-512:35932da2302a9ca08c4c1ccadc1ee04995b91a0ce005977a2804181a59c786fe90794bf0e2de561a871d8d40e1da6c525b1cf0a3b8517dd8dafc92227e6796f0";
 
 #[test]
 fn dd_c11_cid_stable() {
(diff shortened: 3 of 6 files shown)


TSavo:
## Verdict: **CHANGES-REQUIRED**

This is the schema-lock tier for the entire abstraction layer: every downstream `RealizationDesugaringMemento` (the 21-cell trinity body) canonicalizes through these CDDL types, and the five pinned CIDs are the wire-format anchors. Adversarial review found three load-bearing blockers in the pinned fixtures plus several supporting concerns. Two of the three blockers mean shipping incorrect substrate (Supra omnia, rectum); the third means the schema-coordination with PR #628 is on the wrong side of the spec.

---

### Blocker 1 (decisive): `DD_C11_CANONICAL` has duplicate `args` keys; the pinned CID `081b7f07…` is locked to a corrupted rhs

The hand-written canonical form in `implementations/rust/provekit-claim-envelope/tests/abstraction_layer_cid_pin.rs` for the dd→c11 fixture has **duplicate `args` keys** in two of the nested `concept:index` ctor objects:

```
$ python3 -c '...object_pairs_hook=lambda pairs: pairs...'
DUPLICATE KEYS at .post.rhs.args[0].args[0]:        ['args', 'kind', 'name', 'args']
DUPLICATE KEYS at .post.rhs.args[0].args[0].args[0]: ['args', 'kind', 'name', 'args']
```

`serde_json` silently dedups (last-key-wins) when parsing the canonical string before the JCS encoder ever sees it. After dedup the `post.rhs` collapses to:

```jsonc
{"kind":"atomic","name":"concept:call","args":[
  {"kind":"atomic","name":"concept:call","args":[
    {"kind":"ctor","name":"concept:index","args":[
      {"kind":"ctor","name":"concept:tag-of","args":[{"name":"secondary","kind":"var"}]}
    ]}
  ]}
]}
```

This is **not** Sir's specified 2D dispatch-table call: it has lost `receiver`, `method_name`, `args` (the call's variadic arg list), `dispatch_tbl`, the `concept:member` accessor, the `tag-of(receiver)` index, and one of the two `concept:index` dimensions. The `serde_json::Value::Object` round-trip is byte-stable AFTER dedup, which is why `dd_c11_jcs_round_trip` passes — but the CID is pinned to garbage, not to the realization the spec/fn_name claims (`concept:double-dispatch->c11:2d-fn-ptr-table`).

The companion serde fixture `DD_TO_C_JSON` in `abstraction_layer_serde.rs` is clean (no dup keys, structure matches Sir's spec). The two fixtures encode **different rhs structures**. The wire-format-lock test pins the broken one. This is precisely the failure mode the PR description warns about — "a wrong CDDL key order means all 21 trinity cells get re-CID'd later" — except worse, because it's a structurally corrupted term, not just a key-order mismatch.

Fix: re-write `DD_C11_CANONICAL` from the clean serde fixture by running it through the actual `encode_jcs` path (or `serde_json::to_value` followed by `encode_jcs`), don't hand-author. Re-mint the pinned CID. Audit all five canonical fixtures with `object_pairs_hook` before merge; the other four are clean (I checked) but this should be a permanent test guard.

---

### Blocker 2: `dd→java` rhs does not encode Sir's specified visitor pattern

Sir's 2026-05-12 worked example for Java was the interface-visitor: `receiver.accept(secondary)` where `accept` dispatches to `secondary.visit_<receiver_type>(receiver, args)` — a *two-call* indirection. The fixture's `fn_name` says `concept:double-dispatch->jvm:visitor-pattern` but the `rhs` is a single atom:

```json
{"kind":"atomic","name":"jvm:invokeinterface","args":[receiver, method_name, secondary, args]}
```

That's a one-call single-dispatch, not double-dispatch via visitor. The op-name `jvm:invokeinterface` is also a JVM bytecode op (one virtual call), not the visitor pattern. The operational claim in the `fn_name` does not match the term. Either the fn_name is wrong or the rhs is wrong; per Sir's direct input the rhs is wrong.

---

### Blocker 3: `dd→ruby` rhs does not encode Sir's specified `case`-match

Sir's worked example for Ruby was a `case` (or match-arms) over `(type(receiver), type(secondary))` writing the contract's guarded clauses directly — the loosest end of the bracket, where the realization is ≈ the contract and the implication structure is preserved. The fixture's `fn_name` says `case-respond_to` but the `rhs` is a single `ruby:public_send` call:

```json
{"kind":"atomic","name":"ruby:public_send","args":[receiver, method_name, secondary, args]}
```

`public_send` is dynamic single-dispatch on the receiver; it does not bring `secondary`'s runtime type into the resolution at all. This is structurally the same shape as the Java rhs (single op, four args), which directly violates the projection-distance characterization — Java is supposed to be "mid" with visitor indirection; Ruby is supposed to be "near-zero" with the contract written out. As-shipped, dd→java and dd→ruby are nearly identical terms with different op names; the "projection-distance law" the PR claims to demonstrate is not present in the fixtures.

---

### Blocker 4: `LossRecord` schema conflicts with PR #628; spec §2.4 sides with this PR

PR #628's `transport_gap.rs` declares:

```rust
pub struct LossRecord {
    pub value_divergence: Option<Value>,        // serde_json::Value (untyped)
    pub ub_introduction: Option<Value>,
    pub domain_narrowing: Option<Value>,
    pub effect_divergence: Option<Value>,
    pub structural_divergence: Option<Value>,    // successor-mint, same key name
    #[serde(flatten)] pub extra: BTreeMap<String, Value>,
}
```

PR #634 declares:

```rust
pub struct LossRecord(pub BTreeMap<String, IrFormula>);
```

Spec §2.4 explicitly says `loss-record` is `{loss-dimension => formula}` over an open set of dimension keys, with formulas being full `ir-formula` values. **PR #634's BTreeMap-of-IrFormula matches the spec; PR #628's named-struct-of-Value does not** (it conflates the formula type with raw JSON and pins the dimension key set in the struct shape, which the spec rejects on both counts). If both merge, the workspace has two competing `LossRecord` types with different formula types — a JSON `loss_record` with `{"value_divergence": <ir-formula-object>}` minted by #634 will refuse to deserialize as #628's `Option<Value>` only if the consumer is type-strict on `IrFormula`, but they will be unable to round-trip without information loss. The dimension key string `"structural_divergence"` is consistent across both, which is correct. Reconciliation should land #628 onto the spec-compliant shape, not the other way around. Worth flagging in the PR #634 thread so #628 is updated.

---

### Supporting concerns (not individually blocking, but cumulative)

- **The contract formulas are placeholders, not the spec's contract.** Spec §2.3 says the dynamic-dispatch contract is `let m = lookup(runtime_type(receiver), method_name) in defined(m) ∧ wp_call(m, [receiver, ...args])(Q)`. The fixture encodes `choice m: Int. defined(m) ∧ wp_call(m, receiver)` — no `lookup`, no `runtime_type`, `m`'s sort is `Int` (should be `Method`), the variadic `args` is dropped from the inner `wp_call`. As a logical claim this contract is a tautology (∃m. defined(m) ∧ wp_call(m, receiver)(Q)) and does not characterize dispatch. The double-dispatch contract has the same shape and similarly fails to bring `secondary` into a `lookup`. Spec §2.3 explicitly flags this is an open `wp`-formula vocabulary question and the resolution is "same formula language, new shared theory (the `dispatch-theory` SortMemento bundle with `Type`, `Name`, `Method`, `lookup`, `runtime_type`, `defined`)." A PR1 schema-tier placeholder is defensible, but the placeholder content WILL be locked by the pinned CID. Recommend either: (a) thread the actual contract symbols in PR1 since they're cheap (atomic predicate names), or (b) explicitly document in the fixture that the contract is a placeholder and the CID will be superseded by the catalog-mint PR.

- **The projection-distance law tests are presence/absence assertions only**, not heaviness assertions. They check that C has `structural_divergence` + `domain_narrowing`, Java has `structural_divergence` only, Ruby has `structural_divergence` only. They do not compare the *content* of `structural_divergence` across the three to show C > Java > Ruby. Java's is `{"name":"visitor_pattern_indirection"}` and Ruby's is `{"name":"guarded_case_chain"}` — both single atoms, equally "heavy" by any structural metric. The task description specifically asks "verify the assertion is non-trivial (not a tautology, actually compares the three records)." It does not.

- **The sort CIDs are placeholders** (`blake3-512:sort0000…`, `sort1111…`, `sort2222…`, `sort3333…`). Schema-tier OK to land, but every pinned CID inherits these placeholders and will be re-CID'd when real `SortMemento` CIDs replace them. Worth a TODO comment in the fixture noting this is intentional and the catalog-mint PR will supersede.

- **`AbstractionSlot` schema is missing a sort field per `formal_sorts` parallelism.** The spec keeps slot identity and slot sort in two parallel arrays (`slots`, `formal_sorts`), which is fine, but the `AbstractionSlot` struct only carries `name` + optional `variadic`. The CDDL `[+ AbstractionSlot]` + `formal_sorts: [+ sort-ref]` parallel structure is correct per spec §2.1; just noting.

- **PR1 relaxations of `realizations` (`[+]` → `[*]`) and `discharge_receipt` (required → optional)**: the CDDL block documents these and says "PR2 tightens to required per spec §2.2" via a successor-mint with `refines = <PR1 CID>`. Acceptable, but the PR1 CIDs **must** then be permanent (PR2 cannot edit them, only supersede). That's consistent with how `refines` chains work, just worth being explicit so the catalog mint doesn't try to edit PR1 fixtures.

- **The CDDL comments say "Locked JCS key order: kind, operator, tier, slots, …"** which is not alphabetical, so it's not actually what the canonicalizer emits (the canonicalizer sorts keys lexicographically, see `provekit-canonicalizer/src/jcs.rs:53`). The struct declaration order is `kind, operator, tier, slots, formal_sorts, result_sort, contract, contract_note, realizations, superseded_by, refines` — but `serde_json::to_string` on a struct emits in *struct declaration order*, NOT lex order; only `encode_jcs` re-sorts. So a `serde_json::to_string(&struct)` produces non-canonical JSON. As long as everyone routes through `encode_jcs` before hashing this is fine, but the CDDL "Locked JCS key order: kind, operator, tier, …" comment is misleading and should be either re-stated as alphabetical or relabeled "declaration order, not JCS canonical order."

- **Boy-scout (pre-existing, not introduced by this PR):** `menagerie/concept-shapes/cids.tsv` contains absolute paths from a previous worktree (`pk-shrink-hub-2/`); running `bash menagerie/concept-shapes/mint.sh` in this worktree rewrites every line. The PR claims "mint.sh byte-clean" which is false in any fresh checkout. Worth a follow-up to strip absolute paths or anchor them to repo-root.

---

### Hygiene that passed

- 4 files changed, scope matches PR description, no menagerie touches → no scope creep into catalog mints (correct, that's PR2).
- `cargo build --workspace` clean (5 unrelated warnings in provekit-lsp).
- `cargo test -p provekit-ir-types -p provekit-claim-envelope` → 24 tests pass (but Blocker 1 means one of them is pinning the wrong CID).
- `git diff --check` clean.
- `bash menagerie/cross-language-port/driver.sh` → exit 0.

---

### Required actions before re-review

1. Fix `DD_C11_CANONICAL` (re-author from clean serde fixture through `encode_jcs`); audit all five fixtures with `object_pairs_hook` and lock that audit as a test.
2. Rewrite `dd→java` rhs to encode the visitor pattern (two-call `accept`/`visit` indirection) or rename the fixture to match what's actually there.
3. Rewrite `dd→ruby` rhs to encode the `case`-match over `(type(receiver), type(secondary))` preserving implication structure, or rename.
4. Re-mint the 5 pinned CIDs after (1)-(3).
5. Coordinate with #628 on the `LossRecord` shape; spec §2.4 sides with this PR's BTreeMap<String, IrFormula>.
6. Strengthen the projection-distance assertion to compare the three records' content (e.g. depth or argument count of `structural_divergence` formulas across C / Java / Ruby).
7. Either thread real `lookup` / `runtime_type` / `defined` predicate names into the contracts or document the placeholder status in the fixture comments.

Once the five fixtures encode what they say they encode and the CIDs lock those, this PR is the right shape for the abstraction layer's wire-format anchor. The structure of the CDDL types is good; the `effects: []` always-serialize discipline is right; the successor-mint coordination for `structural_divergence` is the right pattern. It's the content of the fixtures, not the schema, that needs another pass.

---

T Savo

TSavo (commented):
## Adversarial re-review — verdict: **MERGE**

Re-reviewed at HEAD `28652b0e` after the two fix commits (`d0b0e800`, `28652b0e`). All three prior blockers close cleanly. No new blockers found.

### Blocker-by-blocker

**B1 — DD_C11 garbage-CID** — closed. `DD_C11_CANONICAL` now has no duplicate keys. JCS-canonical bytes hash to the new pin `35932da2…` (independently recomputed in Python: BLAKE3-512 over `json.dumps(parsed, sort_keys=True, separators=(',',':'))`). The rhs encodes the full 2D table dispatch: `concept:call(concept:cast(concept:index(concept:index(concept:member(receiver, "dispatch_tbl"), tag-of(receiver)), tag-of(secondary)), "fn_ptr_2d"), receiver, secondary, args)` — that's the 2D dispatch-table call shape, not the prior dedup-corrupted single-arg artifact.

**B2 — dd→java single-dispatch** — closed. `fn_name` is now `concept:double-dispatch->jvm:visitor-pattern`. rhs is `concept:seq` of two `concept:itab-method` ctors:
  - `(receiver, "accept", secondary)` — dispatches on receiver's type
  - `(secondary, "visit_receiver_type", receiver, args)` — dispatches on secondary's type

  No `jvm:invokeinterface` anywhere in the file. Two-call indirection confirmed structurally.

**B3 — dd→ruby single-dispatch** — closed. `fn_name` is now `concept:double-dispatch->ruby:case-type-tuple`. rhs is `concept:match` over `concept:pair(concept:type-of(receiver), concept:type-of(secondary))` with a `concept:match-arm` body and a `concept:raise("TypeError")` fallthrough. The implication structure of the abstraction's contract is preserved. No `ruby:public_send` in the file.

### Independent CID verification (all five fixtures)

Recomputed every pinned CID by parsing the canonical string, emitting JCS (sort_keys + (',',':') separators), and hashing BLAKE3-512:

| Fixture | Pinned | Result |
|---|---|---|
| DYNAMIC_DISPATCH | `ca8ff4e6…` | OK |
| DOUBLE_DISPATCH | `1cc86fe7…` | OK |
| DD_C11 | `35932da2…` | OK |
| DD_JAVA | `a024327b…` | OK |
| DD_RUBY | `5f6e90c1…` | OK |

Character-for-character match on all 128 hex chars of each pin.

### Dup-key audit guard — real, not bookkeeping

`scan_for_dup_keys` walks the raw serde_json token stream via `serde::de::Visitor`, tracking seen keys in a per-scope `Vec<String>` and erroring inside `visit_map` BEFORE `next_value` is called — i.e., BEFORE serde_json's BTreeMap last-key-wins dedup. Arrays are recursed via `visit_seq` so nested objects inside nested arrays are audited at every depth. Applied to all five fixtures in `all_five_fixtures_have_no_duplicate_keys`.

Sanity-tested empirically:
- Planted a duplicate top-level key `direction` in `DD_C11_CANONICAL` → audit fired: *"duplicate JSON object key: 'direction' at line 1 column 40"*.
- Planted a duplicate `kind` key inside a nested `{"kind":"primitive","name":"String"}` const-sort node (depth ~5) → audit fired: *"duplicate JSON object key: 'kind' at line 1 column 1721"*.
- Reverted → 15/15 tests pass clean.

The audit catches the exact class of bug that caused the original `081b7f07…` regression. Permanent guard achieved.

### Projection-distance law — preserved

The structural assertions still hold:
- C: 3 loss dimensions (`structural_divergence` + `domain_narrowing` + `ub_introduction`), sd has 3 args (2 index + 1 cast)
- Java: 2 loss dimensions (`structural_divergence` + `domain_narrowing`), sd has 2 args (2 itab-method)
- Ruby: 1 loss dimension (`structural_divergence` only), sd has 1 arg (1 raise)

Strict ordering tests pass. C-has-ub / Java-no-ub / Ruby-no-ub all enforced.

### Hygiene

- `cargo build --workspace`: clean (only pre-existing unrelated dead-code warnings in `provekit-lsp/src/config.rs`).
- `cargo test -p provekit-ir-types -p provekit-claim-envelope`: 15/15 in the cid-pin test binary, plus all sibling binaries green.
- `git diff --check`: clean.
- `bash menagerie/concept-shapes/mint.sh`: exit 0, 45 ops / 143 morphisms / 309 gaps.
- `bash menagerie/cross-language-port/driver.sh`: exit 0.

Mint + driver leave dirt in `cids.tsv` and `artifacts/.../transport-report.json` files, BUT this is **pre-existing** — committed paths reference a stale worktree (`pk-shrink-hub-2`) and any worktree-run rewrites them. Neither fix commit touches `menagerie/`. Out of scope for this PR.

### Scope

`git diff e05f8088..HEAD --stat` = **4 files**:
- `Cargo.lock` (+1)
- `provekit-claim-envelope/Cargo.toml` (+1, adds `hex` dev-dep)
- `provekit-claim-envelope/tests/abstraction_layer_cid_pin.rs` (+281, includes audit guard)
- `provekit-ir-types/tests/abstraction_layer_serde.rs` (+125)

No collateral damage. No concept hub op specs touched. No CDDL touched in the fix commits. `LossRecord` shape unchanged: `pub struct LossRecord(pub BTreeMap<String, IrFormula>)` — per spec §2.4 BTreeMap design (vs. struct-with-Options shape rejected in prior review of #628).

### Verdict

**MERGE.** Three blockers close cleanly. The fix is mechanically correct (CIDs verified by independent recomputation), structurally sound (visitor pattern + case-match shapes are the real thing), and adds a permanent guard against the class of bug that produced the regression. Supra omnia, rectum — the substrate now matches what the spec says it is.

PR #639, opened 2026-05-12
fix(lift-zig-source): correct string-literal escape decoding (Supra omnia, rectum)

TSavo:
## Bug

**Surfaced by:** PR #632 nit sweep (nit #1 from PR #596).

The Zig source lifter was refusing **all** string literals containing escape sequences with `unsupported-string-escape`. This meant no Zig source with `\n`, `\t`, `\\`, `\"`, `\xNN`, or `\u{NNNN}` could be lifted at all. It was a stopgap placeholder, not a fix. Supra-omnia-rectum violation: the IR term bytes could not correctly represent the source's runtime semantics for escaped strings.

## Root cause

`emitString` checked for any `\` in the raw source text and bailed immediately (lines 496-499 in the original). No CID was produced; the function was refused.

## Fix

**`emitString` — split by AST node tag:**
- `.string_literal`: call `std.zig.string_literal.parseAlloc(alloc, raw_with_quotes)` from Zig stdlib. It handles `\n`, `\r`, `\t`, `\\`, `\'`, `\"`, `\xNN`, `\u{NNNN}` per the Zig spec and returns decoded bytes. Invalid escape sequences produce a `invalid-string-literal` refusal (not a panic).
- `.multiline_string_literal`: walk the `multiline_string_literal_line` tokens (first..last from `nodeData.token_and_token`), strip the two-byte `\\` prefix from each token, join with `\n`. Multiline strings have **no** escape sequences per Zig spec — they are raw byte content. The previous `\` check was incorrectly refusing them too.

**`appendEscapedString` — pretty-printer fix:**
Extended to handle `\r` (`\\r`), `\t` (`\\t`), and a `\\xNN` fallback for all other non-printable/non-ASCII bytes. This is required for the `lift(compile(lift(src)))` round-trip to hold on strings containing any decoded byte other than `\\`, `"`, or `\n`.

## CID drift

**0 committed fixture CIDs moved.** No existing `zig-language-signature` catalog artifact or `cross-language-port` fixture exercised string literals with escape sequences — because they were being refused, none could have been committed. The regression tests are the new pins.

## Regression tests added (7 new, 1 updated)

| Test | Escape covered |
|------|---------------|
| `decodes \\n escape in string literal to newline byte in IR term` | `\n` |
| `decodes \\t and \\r escape sequences in string literals` | `\t` |
| `decodes \\x hex escape to raw byte` | `\xNN` (`\x41` == `A`) |
| `decodes \\u{NNNN} unicode escape to UTF-8 bytes` | `\u{263A}` == `☺` |
| `decodes \\\\ escaped backslash to single backslash byte` | `\\` |
| `lifts multiline string literal as raw decoded bytes` | multiline `\\` syntax |
| `round trip string literal with escape sequences preserves decoded bytes` | end-to-end `lift(compile(lift(src)))` |

The test `refuses string literal escape sequences...` was updated: it now asserts the decoded bytes appear in the IR (the PASS condition), not a refusal (the old FAIL condition it was locking in).

## Boy-scout observations (not fixed in this PR)

- `char_literal` is already refused with `unhandled-syntax` — correct, no change.
- Numeric literals with underscores (`1_000`) work correctly via `std.fmt.parseInt` base-0 — verified.
- `\x` hex format in `appendEscapedString` uses `\xNN` (two uppercase hex digits) which is valid Zig syntax and round-trips through the lifter.

## Test results

```
zig build test --summary all
Build Summary: 3/3 steps succeeded; 12/12 tests passed
```

```
cargo test -p provekit-cli --test mint_kit_integration -- zig
test zig_kit_contract_set_cid_is_pinned_to_self_contracts_canonical ... ok
1 passed; 0 failed
```

```
cargo test -p provekit-cli --test transport_cli
8 passed; 0 failed
```

(3 pre-existing failures in `mint_kit_integration` for `cpp` and `rust` kits are on `origin/main` before this branch — not introduced here.)

## Deep review needed

This is a substrate-touching change. The CID of every lifted Zig string literal now changes from "refused, no CID" to "correct decoded-bytes hash". Needs Opus review before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * String literals containing escape sequences (`\n`, `\t`, `\r`, Unicode escapes, backslashes) are now properly decoded
  * Control characters and non-printable bytes in strings are correctly encoded

* **New Features**
  * Added full support for multiline string literals

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/639)

### DIFF
--- a/implementations/zig/provekit-lift-zig-source/src/lift.zig
+++ b/implementations/zig/provekit-lift-zig-source/src/lift.zig
@@ -493,12 +493,44 @@ const Lifter = struct {
     }
 
     fn emitString(self: *Lifter, node: Node.Index) !provekit.Term {
-        const raw = self.tree.getNodeSource(node);
-        if (std.mem.indexOfScalar(u8, raw, '\\') != null) {
-            return self.refuse(node, "unsupported-string-escape", "string literal with escape sequences not yet decoded on lift");
+        switch (self.tree.nodeTag(node)) {
+            .string_literal => {
+                const raw = self.tree.getNodeSource(node);
+                // raw includes the surrounding double-quotes; parseAlloc decodes all Zig escapes.
+                const decoded = std.zig.string_literal.parseAlloc(self.alloc, raw) catch |err| switch (err) {
+                    error.InvalidLiteral => return self.refuse(node, "invalid-string-literal", "string literal contains an invalid escape sequence"),
+                    error.OutOfMemory => return error.OutOfMemory,
+                };
+                return provekit.Str(decoded);
+            },
+            .multiline_string_literal => {
+                // A multiline string is a sequence of `.multiline_string_literal_line` tokens.
+                // Each token's slice includes the `\\` prefix and a trailing newline.
+                // Decoded value: strip `\\ ` (or just `\\` when line is empty), join lines.
+                // No escape sequences exist inside multiline strings; they are raw bytes.
+                const first_tok = self.tree.nodeData(node).token_and_token[0];
+                const last_tok = self.tree.nodeData(node).token_and_token[1];
+                var out = std.ArrayList(u8).empty;
+                var tok = first_tok;
+                while (tok <= last_tok) : (tok += 1) {
+                    const line = self.tree.tokenSlice(tok);
+                    // Each line token starts with `\\` (two bytes). Strip them.
+                    // The token may or may not include a trailing newline depending on whether
+                    // it's the last line. To be safe, strip a leading `\\` then any space char,
+                    // then strip a trailing `\n` only between lines (not after the last).
+                    const content = if (line.len >= 2) line[2..] else line;
+                    // Strip trailing newline that the tokenizer includes.
+                    const trimmed = if (content.len > 0 and content[content.len - 1] == '\n')
+                        content[0 .. content.len - 1]
+                    else
+                        content;
+                    if (tok > first_tok) try out.append(self.alloc, '\n');
+                    try out.appendSlice(self.alloc, trimmed);
+                }
+                return provekit.Str(try out.toOwnedSlice(self.alloc));
+            },
+            else => unreachable,
         }
-        if (raw.len >= 2 and raw[0] == '"' and raw[raw.len - 1] == '"') return provekit.Str(raw[1 .. raw.len - 1]);
-        return provekit.Str(raw);
     }
 
     fn emitUnreachable(self: *Lifter, node: Node.Index) !provekit.Term {
@@ -970,7 +1002,15 @@ fn appendEscapedString(alloc: std.mem.Allocator, out: *std.ArrayList(u8), value:
         '\\' => try out.appendSlice(alloc, "\\\\"),
         '"' => try out.appendSlice(alloc, "\\\""),
         '\n' => try out.appendSlice(alloc, "\\n"),
-        else => try out.append(alloc, c),
+        '\r' => try out.appendSlice(alloc, "\\r"),
+        '\t' => try out.appendSlice(alloc, "\\t"),
+        0x20...0x21, 0x23...0x5B, 0x5D...0x7E => try out.append(alloc, c), // printable, not \ or "
+        else => {
+            // Emit as \xNN for any other byte (control chars, high bytes).
+            var buf: [4]u8 = undefined;
+            const s = std.fmt.bufPrint(&buf, "\\x{X:0>2}", .{c}) catch unreachable;
+            try out.appendSlice(alloc, s);
+        },
     };
 }
 
@@ -1093,7 +1133,7 @@ test "refuses unhandled switch without emitting unknown operation" {
     try std.testing.expectEqualStrings("switch.zig.f", out.refusals[0].function.?);
 }
 
-test "refuses string literal escape sequences instead of lifting literal backslashes" {
+test "decodes \\n escape in string literal to newline byte in IR term" {
     var arena = std.heap.ArenaAllocator.init(std.testing.allocator);
     defer arena.deinit();
     const alloc = arena.allocator();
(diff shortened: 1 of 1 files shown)


TSavo (commented):
## Verdict: **MERGE-WITH-NITS**

Substrate-touching review (lifter affects IR-term bytes → CIDs). Verified against Zig 0.16.0 stdlib at `/usr/local/Cellar/zig/0.16.0_1/lib/zig/std/zig/string_literal.zig`. Cargo and `zig build test` both green on a fresh worktree.

### Correctness (the load-bearing checks)

1. **`parseAlloc` delegation is correct.**
   - Input shape: `parseWrite` asserts `bytes[0]=='"' && bytes[len-1]=='"'`; passing `getNodeSource(node)` for a `.string_literal` tag satisfies that, since the token slice includes quotes.
   - Error set: `ParseError = {OutOfMemory, InvalidLiteral}` is closed. The catch is an exhaustive switch (Zig enforces). `InvalidLiteral` → structured `Refusal{kind: "invalid-string-literal"}`. ✓
   - Bytes-on-IR semantics: `\u{NNNN}` UTF-8-encodes via `std.unicode.utf8Encode`; `\xNN` writes a single raw byte (NOT UTF-8 validated). IR carries raw decoded bytes — the program's *meaning*, option (a) from the brief. ✓

2. **Multiline walk is correct in practice.**
   - Joining char is exactly `'\n'` (0x0A). ✓
   - Empty-line case (`\\` alone): `line[2..]` = empty slice → joined as empty line. ✓
   - `\r` is excluded from token slices (tokenizer terminates the line token at the byte position of `\r` or `\n`, with `result.loc.end = self.index` before the inner switch increments).

3. **Round-trip CID equality verified.** The round-trip test (`rt_escape.zig`) pins `canonicalTermBytes(first) == canonicalTermBytes(second)` — the strongest possible assertion. Even though pretty-printing emits `\xE2\x98\xBA` instead of `\u{263A}` for non-ASCII bytes, the SEMANTIC re-parse converges. Source is decoratively lossy; IR is bytes-stable. Aligned with Supra-omnia-rectum.

4. **Pretty-printer escape table is byte-complete.** Ranges `0x20-0x21 / 0x23-0x5B / 0x5D-0x7E` carve out exactly the printable-ASCII set minus `"` and `\`, with `\\ " \n \r \t` handled by named cases and everything else (0x00-0x1F, 0x7F-0xFF) falling through to `\xNN`. Every byte value 0x00-0xFF has defined output. ✓

5. **Memory lifetimes align.** `parseAlloc(self.alloc, raw)` and `out.toOwnedSlice(self.alloc)` (verified in `std/array_list.zig:654` — uses the explicit `gpa` argument) both allocate on the lifter arena; `provekit.Str` borrows the slice; both outlive the IR term consumption.

### Nits (follow-up, not blockers)

1. **Missing negative test.** No regression test exercises `error.InvalidLiteral` → `Refusal{kind: "invalid-string-literal"}`. The exhaustive-switch structural argument defangs this, but a one-line test (e.g. source containing `"\q"`) would close the loop empirically. Recommend follow-up.

2. **Dead defensive code in multiline strip.** Per Zig 0.16 tokenizer (`tokenizer.zig:762-775`), the `multiline_string_literal_line` token's `loc.end` is set at the `\n` byte, so the slice *never* ends in `\n`. The trailing-newline strip:
   ```zig
   const trimmed = if (content.len > 0 and content[content.len - 1] == '\n')
       content[0 .. content.len - 1]
   else
       content;
   ```
   is unreachable, and the comment ("tokenizer includes a trailing newline") is wrong. Similarly `if (line.len >= 2) line[2..] else line` is dead since `.multiline_string_literal_line` tokens always begin with `\\`. Either delete and let `line[2..]` stand bare with a one-line comment explaining the tokenizer guarantee, or keep as belt-and-suspenders but fix the comment.

3. **Multiline walk lacks tag assertion.** The loop `tok = first_tok; while (tok <= last_tok) tok += 1` works because the parser guarantees a contiguous run of `.multiline_string_literal_line` tokens, but `std.debug.assert(self.tree.tokenTag(tok) == .multiline_string_literal_line)` would make that guarantee load-bearing and catch any future grammar evolution.

4. **Multiline round-trip not directly pinned.** The existing escape round-trip test exercises the same `appendEscapedString` → `parseAlloc` pathway, so this is a coverage nit, not a correctness gap. A `multi_rt` companion test is the obvious addition.

### Boy-scout findings (out of scope for this PR — separate task)

- **`char_literal` still refused** (`lift.zig:436`). Same structural fix applies: `std.zig.string_literal.parseCharLiteral` exists in the same stdlib module. Tracking as a sibling task to this one.
- **Numeric underscores** confirmed handled via `std.fmt.parseInt(..., 0)` on existing path; no regression-test pin but pre-existing behavior, not a deficit introduced here.

### Substrate / CID drift

Zero. No existing fixture exercised escaped strings (they were all refused pre-fix). The 7 new tests are the first pins. The fix CANNOT break any existing pinned CID; it only creates new ones for inputs that previously could not produce any IR. Aligned with the claim in the commit message.

### Hygiene

- `zig build test -p provekit-lift-zig-source`: green.
- `cargo build -p libprovekit -p provekit-cli`: clean.
- Single-file change (181/-13 in `implementations/zig/provekit-lift-zig-source/src/lift.zig`); no menagerie or catalog touch; mint.sh untouched.

### Conclusion

The fix correctly reverses the PR #596 stopgap and replaces it with the canonical Zig stdlib decoder. Error handling is exhaustive. Round-trip CID stability is empirically pinned. The nits are cleanup, not correctness — none block merge.

**MERGE-WITH-NITS.** Land it; open follow-up tickets for the negative test, the dead-code cleanup, the `tokenTag` assert, and `char_literal` (separate scope).

— substrate review, T Savo

PR #688, opened 2026-05-12
fix(discharge): notes-field-tolerant subsumption (closes #686)

### LINKED ISSUE #686: discharge.py destructively wipes ruby:source-unit on every mint (pre-existing wp/notes mismatch)
Surfaced by #685 (assert re-mint) when the agent had to manually restore ruby:source-unit files after each `mint.sh` run.

**Pre-existing root cause:** `menagerie/ruby-language-signature/specs/op_source-unit.spec.json` carries `wp: "lossless Ruby source wrapper; project_effects descends to operational_term"` + an extra `notes` field; `menagerie/concept-shapes/specs/op_source-unit.spec.json` carries `wp_note: "lossless source wrapper; the source bytes are recoverable..."` with no `notes`. `structural_subsumption` in `mint_language_morphisms.py` doesn't bridge these — the morphism fails to discharge — `discharge.py`'s sweep removes `catalog/algorithms/morphism:ruby:source-unit:*` plus the receipt and the discharge artifact.

**Net effect:** every cell mint that runs the full chain destructively wipes the ruby:source-unit morphism. PRs that include a clean mint pollute their diff with these ruby deletions. The #673 original Haiku PR included them as scope creep precisely because the mint chain produced them automatically.

**Two paths:**
- (a) Fix ruby's `op_source-unit` spec to drop the `notes` field and bring the wp prose into agreement with the concept-hub form (substrate spec change).
- (b) Teach `structural_subsumption` to handle a `notes` field as soft-metadata that doesn't block subsumption (substrate discharge-logic change).

Option (b) is more principled — `notes` is documentation, not a discharge obligation. Either way, the wp-mismatch needs resolving so the mint chain stops wiping a legitimate morphism.

🤖 Filed 2026-05-12 per #685 finding.

TSavo:
## Summary

- **Root cause**: `ruby:op_source-unit.spec.json` carries a `notes` field in `post` that the concept hub spec does not. The existing `wp-text abstraction` path in `try_structural_subsumption` replaced `post.wp` but left `post.notes` intact, so the relaxed CID never matched the concept CID, discharge failed, and `discharge.py`'s sweep deleted the `morphism:ruby:source-unit` artifacts on every `mint.sh` run -- polluting every cell-mint PR diff with ruby deletions.

- **Fix (option b -- principled)**: Extended `_make_relaxed` so that when `relax_wp=True` it also strips documentation-only extra keys (`notes`, keyed in `_DOC_ONLY_KEYS`) that appear in the language `post` but not in the concept `post`. `notes` carries no semantic discharge obligation -- it is human-readable implementation detail in the same category as `wp`. The concept hub never emits `notes`; language specs may include it as an annotation.

- **Soundness preserved**: specs that differ in a structural field (`operator`, `arity_shape`, `formal_sorts`, effects direction) still do not discharge. The soundness gate test `test_structural_field_mismatch_does_not_discharge` verifies this. `wp` prose differing between lang and concept is still fully subsumed (existing behavior unchanged).

## Scope

Also benefits `ruby:call` and `python:call` which carry the same `notes` pattern -- they will discharge correctly if they hit the same gap in a future mint.

## Verification

- `morphism:ruby:source-unit:to:concept:source-unit` is present in `catalog/algorithms/`, `cids.tsv`, and `transport-gaps.md` Minted Coverage table after the fix.
- Receipt: `discharged: true`, `method: structural-wp-abstraction`, `wp_abstracted: true`.
- `mint.sh` run twice: `morphism_count 161` both times (byte-stable).
- `morphism_count` delta vs pre-fix main: +1 (ruby:source-unit restored, nothing else lost).

## Regression tests

5 unit tests in `menagerie/concept-shapes/tests/test_wp_notes_subsumption.py`, mocked against `canonical_cid_spec` so they run without the Rust binary:

| Test | What it checks |
| --- | --- |
| `test_notes_only_difference_discharges_via_wp_abstraction` | notes alone doesn't block when wp matches |
| `test_wp_and_notes_both_differ_discharges_via_wp_abstraction` | exact ruby:source-unit regression case |
| `test_structural_field_mismatch_does_not_discharge` | soundness gate -- real semantic diff must not discharge |
| `test_no_notes_no_regression_wp_only_still_works` | pre-existing wp-only abstraction unchanged |
| `test_identical_specs_returns_none` | byte-equal early-return guard |

All 5 pass. Existing `test_discharge.py` 3/3 still pass.

## Files changed

- `menagerie/concept-shapes/scripts/mint_language_morphisms.py` -- core fix in `try_structural_subsumption`
- `menagerie/concept-shapes/tests/test_wp_notes_subsumption.py` -- new regression tests (5 tests)
- `menagerie/concept-shapes/catalog/algorithms/morphism:ruby:source-unit:to:concept:source-unit.*` -- new CID (ruby morphism restored)
- `menagerie/concept-shapes/receipts/morphism_ruby_source_unit_to_source_unit.receipt.json` -- updated receipt
- `menagerie/concept-shapes/specs/morphism_ruby_source_unit_to_source_unit.spec.json` -- updated spec
- `menagerie/concept-shapes/cids.tsv`, `discharges/`, `catalog/index.json` -- updated artifacts

**Needs deep Opus review before merge per Supra omnia, rectum.**

## Release Notes

* **New Features**
  * Documentation notes field now supported in contract specifications.

* **Bug Fixes**
  * Structural subsumption now correctly treats documentation-only notes as non-functional differences, enabling discharge when specifications differ only in notes.

* **Tests**
  * Added comprehensive regression tests for notes-field handling in structural subsumption.


TSavo (commented):
**Verdict: MERGE-WITH-NITS**

Adversarial review against main confirms the fix is correct, narrowly scoped, and the soundness gate is real.

### Fix verified
- `_make_relaxed` change in `mint_language_morphisms.py` is exactly as described: `_DOC_ONLY_KEYS = {"wp", "notes"}`, strips `notes` only when `relax_wp=True` AND `doc_key not in concept_post`. No new `discharge_method` string; folds under existing `"structural-wp-abstraction"`.
- `notes_match` is correctly added to the byte-equal early-return guard and to the `needs` predicate that gates relaxation activation.

### Tests verified
- 5/5 tests in `tests/test_wp_notes_subsumption.py` pass. They cover: (a) notes-only diff discharges, (b) wp+notes diff (the exact ruby:source-unit failure mode) discharges, (c) **structural-operator mismatch with wp+notes both diff does NOT discharge** (the soundness gate), (d) wp-only pre-existing path still works, (e) byte-equal early-return.
- `scripts/test_discharge.py`: 3/3 pass.

### Empirical
- `morphism_count: 161` post-fix vs `160` pre-fix (cross-checked by running main's `mint_language_morphisms.py` against the same catalog state — exact delta = +1 ruby:source-unit). `morphism_ruby_source_unit_to_source_unit` is present in `cids.tsv`. Receipt records `method: "structural-wp-abstraction"`, `wp_abstracted: true`, `discharged: true`.
- Two consecutive mint runs from committed state are byte-identical (md5 of cids.tsv and index.json stable).

### Collateral
- PR commit touches exactly 10 files, all under `menagerie/concept-shapes/`. No collateral damage.
- `notes` field appears in 24 op specs across languages (inside `post`, never top-level outside x86-64 draft). Surveyed the contents: most are prose; a few (e.g. `op_compound_assign_*` "C evaluates lvalue once") encode semantics — but the structural fields (`operator`, `arity_shape`, `formal_sorts`, `effects`, `pre`, `post.kind`/`result`/`arity`) all still get checked byte-for-byte before `notes` is stripped. The wp-abstraction soundness argument carries through to notes unchanged: if `wp` (a discharge obligation in human-readable form) is bridgeable, then `notes` (sibling prose elaboration) is bridgeable a fortiori.
- Only consumers of `post.notes` in the codebase are the new logic itself and the discharged-substitution receipts. No surprise downstream.

### Nits (non-blocking)
1. **CONFLICTING with main**: sole conflict is in `cids.tsv` (mint-regeneration target). Trivially resolved by re-running `python3 scripts/mint_language_morphisms.py` post-merge.
2. **Receipt label fidelity**: receipt says `method: "structural-wp-abstraction"` and `wp_abstracted: true`, but here both `wp` AND `notes` were bridged. Consider extending receipts to record `notes_abstracted: true` (or similar) for audit clarity. Not a soundness gap — the discharge code path is the same — just a labelling refinement.
3. **Pre-existing, separate from this PR**: `cids.tsv` bakes absolute filesystem paths (`/Users/tsavo/provekit-worktrees/pk-fix-686/...`) into the committed artifact, so each minter's worktree leaves a different path footprint. PR #624 addressed this pattern for driver artifacts; `mint_language_morphisms.py` cids.tsv writer should follow the same run-dir-relative convention. Flag for follow-up, not this PR's scope.

The fix is correct, the soundness gate is real, and the framing (notes is documentation; doc-divergence shouldn't block algebraic subsumption) holds under scrutiny. Merge after re-mint resolves the cids.tsv conflict.

PR #697, opened 2026-05-12
feat(lift): naming round-trip — lifter extracts concept_hint from comments

TSavo:
## Summary

- Adds `concept_hint: Option<String>` to `ContractDecl` (provekit-ir-symbolic) and `FunctionContractMemento` (libprovekit) as metadata-only fields — excluded from `canonical_bytes`/CID derivation so shape identity is stable regardless of annotation changes.
- `provekit-lift-contracts` gains `lift_file_with_source(file, path, src: Option<&str>)`. When source text is supplied, it scans backwards from each function's span line, skipping `#[attr]` lines, for `// concept: <name>` annotations. Also accepts `/// concept: <name>` doc-comment form via `#[doc = "..."]` attributes.
- Name grammar enforced: `[a-zA-Z][a-zA-Z0-9\-:_]*`. Malformed names (spaces, leading digits, etc.) return `None`. `UNNAMED-CONCEPT-N` placeholders are stored verbatim.
- 5 new unit tests: human name extracted, UNNAMED placeholder preserved, absent returns None, malformed rejected, doc-comment form accepted.
- All `ContractDecl { ... }` and `FunctionContractMemento { ... }` struct literals across the workspace updated (23 files, boy-scout sweep).

## Architecture note

`concept_hint` is NOT in `build_value()`/JCS bytes. The field is carried in-memory alongside the shape CID. The downstream binding step (separate PR) will: cluster shape → lookup catalog → use hint to propose name or mint new entry. This PR is lifter-side extraction only.

## Test plan

- [ ] `cargo test -p provekit-lift-contracts` — 8 tests pass (5 new concept_hint tests)
- [ ] `cargo test -p provekit-ir-types` — 8 tests pass
- [ ] `cargo build --workspace` — clean build, no errors

## Release Notes

* **New Features**
  * Added support for concept hint metadata through code annotations (`// concept:` and `/// concept:` comments) to capture human-supplied concept names.
  * Added new `lift_file_with_source()` function to enable extraction of concept hints during contract lifting.

* **Chores**
  * Updated C11 language signature identifiers and contract specifications.
  * Added `span-locations` feature to dependency configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/697)

### DIFF
--- a/implementations/c/provekit-walk-c/src/c11_cursor_dispatch.generated.c
+++ b/implementations/c/provekit-walk-c/src/c11_cursor_dispatch.generated.c
@@ -6,7 +6,7 @@
 #include <string.h>
 
 const char *pk_c11_signature_cid(void) {
-    return "blake3-512:13e0215bde01ee47e8dbaed5ad3cc48eb4c6dd7e3ca0e25ef50e0b6cce42338850c7703b08a4836b851beea03e66e47ed05ecd63f599d4eec349b66970663c69";
+    return "blake3-512:b5ad11779fe86ac7600fd3c4efad91c668505c92e0838d66c581fe52de08057e29b7a1e21885cbb1da695019066d3b0939ccd64ba5ea3de4dd40559d6f3fc801";
 }
 
 const pk_c11_cursor_dispatch *pk_c11_cursor_dispatch_lookup(enum CXCursorKind kind) {
@@ -128,7 +128,7 @@ const pk_c11_cursor_dispatch *pk_c11_cursor_dispatch_lookup(enum CXCursorKind ki
             (enum CXCursorKind)9, /* CXCursor_VarDecl */
             "CXCursor_VarDecl",
             "c11:decl",
-            "blake3-512:c5bb5a34d98ac0f9f2d4a8cbf2de8229b0dd2f02662879d8141015aa7ee0c6fde41fa839ca72cab720669764343f322a8754d0f47e37ab31f826dfbb827fd161",
+            "blake3-512:3ff181ea4b9fa837e54f3f5248a2527b10d7410418a09e61e61a015cbb023d5bbacc1ecf59653b3437a25b1a12313fbc8abc573f290ea2a891bbc748745b5bb1",
             2,
             "{\"kind\": \"named\", \"slots\": [{\"name\": \"name\"}, {\"name\": \"initializer\"}]}",
             "declared name, initializer or unit",
@@ -1598,7 +1598,7 @@ const pk_c11_cursor_dispatch *pk_c11_cursor_dispatch_lookup(enum CXCursorKind ki
             (enum CXCursorKind)202, /* CXCursor_CompoundStmt */
             "CXCursor_CompoundStmt",
             "c11:seq",
-            "blake3-512:b8a40ec2291a515b73888cdb59fd87e8fb032042193e12b93cb105401383701528ba0cc7571a148c277795206b0662bcf99f53fdb96254976636e99047c7b333",
+            "blake3-512:6c476a5d6b7c4eaf832ec897b1b518364b049578764a37183a109a2ae86942e3699cc27e9ef8d6a059b095a8807a0054a7492ecb57c9c9895562a7b659f5dacb",
             2,
             "{\"arity\": 2, \"kind\": \"positional\"}",
             "fold children left with seq; empty body is skip",
@@ -1640,7 +1640,7 @@ const pk_c11_cursor_dispatch *pk_c11_cursor_dispatch_lookup(enum CXCursorKind ki
             (enum CXCursorKind)205, /* CXCursor_IfStmt */
             "CXCursor_IfStmt",
             "c11:if",
-            "blake3-512:29004b6aa5186996e03e7a7d0615aeefb1d147b1b2ce29aa476052c294578271c0f4b09d32c19cf5f458ced8cf0839c0f229d0bd7b131bd5c5fbbf44d8e25f35",
+            "blake3-512:ba0d198096aa099df8437c015e7dc52ccf7e97e17707ec6c6e69ae38fc85057b1b7a71e54897bdb1ac0da80a56707e4c00f46b39570644334526ec009f225636",
             3,
             "{\"kind\": \"named\", \"slots\": [{\"name\": \"cond\"}, {\"name\": \"then_branch\"}, {\"name\": \"else_branch\"}]}",
             "condition, then branch, else branch or skip",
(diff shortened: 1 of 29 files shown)


TSavo:

Reviewer cloned a fresh worktree of `feat/naming-roundtrip-lifter` (HEAD `0467cbb0`, branched off `ac82e193`), inspected the 23-file diff in full, ran the affected test suite, and probed edge cases with 17 adversarial fixtures + 1 byte-level CID-stability regression test.

## Verified (strong)

- **Shape-CID stability is real and byte-deterministic.** Wrote a regression test (`libprovekit/tests/cid_stable_concept_hint.rs`) that constructs `FunctionContractMemento` with `concept_hint = None` vs `Some("retry-with-jitter")` vs `Some("UNNAMED-CONCEPT-7")` vs `Some("totally:different::name-2")` and asserts identical `canonical_bytes` and `cid`. Passes. `build_value()` in `libprovekit/src/compose.rs` takes no `concept_hint` parameter; `build_memento_value()` does not pass it. The metadata-only invariant holds at the byte level.
- **Struct-literal sweep is complete.** `cargo check --workspace` is clean. The field has no `Default`, so any missed literal would error — confirming all 18 struct-literal sites were updated (contract.rs:76, marriage.rs:854+1068, llbc_calls.rs:333, cmd_compose.rs:446, asm-x86-64/lib.rs:785, ffi.rs:354, compose_smoke.rs, core_interface.rs, layer2.rs ×3, rust-tests/lib.rs:247, etc.).
- **Grammar bounds correct on the rejection side.** 9 malformed-name fixtures all rejected: leading digit, leading hyphen, unicode (Greek alpha), dot, paren, slash, leading underscore, space-in-name, empty-after-prefix. 5 well-formed fixtures all accepted: hyphen-in-middle, underscore-in-middle, colon-in-middle, 1000-char name, namespaced (`ns:concept-name`).
- **Path B (raw source scan) edge cases solid.** Skips `#[derive]`/`#[allow]` attribute lines between annotation and `fn` keyword. Traverses blank lines. Walks upward through unrelated `//` comments to find the `// concept:` line. Works on `impl` methods.
- **Affected-package test suites green.** `provekit-lift-contracts` 8/8 (5 new + 3 pre-existing), `libprovekit` 17/17 (+ 4 desugar + 4 ffi_smoke + 4 transport), `provekit-ir-symbolic` 30/30, `provekit-cli` 19/19 + 4/5 polyglot_smoke (the 1 failure is a pre-existing `provekit-linkerd` binary-missing flake, not introduced by this PR — `polyglot_smoke.rs` is not in the touched files).
- **provekit-walk pre-existing breakage is not from this PR.** Walk fails to compile on `ac82e193` already because `IrFormula::Substitute`/`Apply` variants from #620 lack match arms in `walk/src/marriage.rs` and `walk/src/llbc_calls.rs`. Pre-existing; the touched-file edits in walk are limited to struct-literal `concept_hint: None` additions and are syntactically clean.

## Blockers

**Path 1 (doc-attr) silently fails on the idiomatic placement of `/// concept:`.** The PR module docstring (lib.rs lines 30–36) and the docstring on `concept_hint_from_span` claim `/// concept: <name>` is a supported form via the `#[doc = "..."]` attribute. Empirically it is not — `concept_hint_from_span` iterates `attrs.iter().rev()` and `break`s at the first non-`doc` attribute. For the idiomatic placement:

```rust
/// concept: retry-with-jitter
#[requires(x > 0)]
fn retry(x: i64) -> i64 { x }
```

…attrs are `[doc, requires]` in source order. Reversed: `[requires, doc]`. The loop sees `requires` first, breaks immediately, and the `doc` attr is never inspected.

I verified empirically:
- `path_a_doc_attr_only` test: putting `/// concept:` BELOW `#[requires]` works (Path 1 fires because doc is then the last attr). Non-idiomatic placement.
- `doc_attr_works_without_source_text` test (PR's claim with `lift_file`, no `source_text`): FAILS — returns `None` instead of `Some("doc-only")`.

The PR's own `concept_hint_doc_comment_form_accepted` test masks this by passing `source_text`, so Path 2 catches it. The test does not exercise the claim it appears to validate.

**Compounding**: the only production caller, `provekit-build/src/lift_pass.rs:163`, uses `lift_file(&parsed, &path_str)` — no `source_text`. So in production today, both `// concept:` and `/// concept:` placed idiomatically (above `#[requires]`) yield `None`. The feature is wired but inert under idiomatic Rust ordering.

**One-line fix**: change `break` → `continue` in the reverse-iter loop at `provekit-lift-contracts/src/lib.rs:232`, OR drop the short-circuit entirely and scan every attr regardless of kind. Add a test that exercises Path 1 ALONE (no `source_text`) with idiomatic placement.

## Nits

- The `_function_cid` parameter in walk's `PinInvariantMementoView` is unrelated to this PR; ignore. (Boy-scout-scoping note: nothing changed here.)
- `provekit-walk` not being in any workspace + pre-existing broken match arms is a standing structural issue (separate PR territory).
- `lift_file_with_source` is wired but unused by production (`provekit-build` follow-up).

## Verdict

**CHANGES-REQUIRED.** Substrate PR. Supra omnia, rectum. The PR documents `/// concept:` as a supported form; that claim does not hold in production under idiomatic placement. One-line code fix + one Path-1-only test will discharge the blocker. The metadata-only/CID-stability claims are empirically true and the sweep is complete; merge after the fix.

— adversarial-reviewer / opus-4.7-1m

TSavo:
Fixed in a580a2e8.

**The one-line fix:** changed `break` to `continue` in `concept_hint_from_span` Path 1 (line 232). Idiomatic ordering `[doc, requires]` produces reverse iter `[requires, doc]`; the old `break` exited on `requires` before seeing the `doc` attr. The `continue` skips all non-doc attrs and finds the first matching doc attr.

**Regression test added:** `concept_hint_doc_above_requires_extracts_correctly` — mirrors the production caller shape (`lift_file`, no source_text). Verifies `/// concept: retry-with-jitter` above `#[requires(x > 0)]` extracts correctly. All 9 tests pass.

**Open question for Sir — lift_pass.rs:163:** production calls `lift_file` (Path A only); `lift_file_with_source` (Path B, raw source scan) is wired but inert. With this fix, Path A now works correctly under idiomatic ordering, so production is unblocked. Whether to switch `lift_pass.rs:163` to `lift_file_with_source` to also activate Path B is an architect call — flagging for follow-up, not touching in this PR.

TSavo:
## Adversarial Re-Review — VERDICT: MERGE

Re-verified the fix at commit `a580a2e8` on-disk in a fresh worktree.

### Blocker resolution — closed
- `implementations/rust/provekit-lift-contracts/src/lib.rs:232` now reads `continue;` (was `break;`). Reverse-iter loop correctly scans past non-doc attrs to find the `#[doc = ...]` carrying `concept: <name>`. Comment block at L213 + L229 updated to reflect the new semantics.

### Regression test — present and load-bearing
- `concept_hint_doc_above_requires_extracts_correctly` at L711. Uses `lift_file(&f, "test.rs")` (no `source_text`) — matches the production caller in `provekit-lift-pass/src/lib.rs:163`. Source has `/// concept: retry-with-jitter` directly above `#[requires(x > 0)]`, exactly the shape the blocker silently dropped.
- Asserts `out.decls[0].concept_hint == Some("retry-with-jitter")`.

### Empirical verification
- `cargo test -p provekit-lift-contracts` against `a580a2e8`: **9 passed, 0 failed**, including the new test.
- Temporarily reverted `continue` → `break` in the worktree, re-ran: new test **FAILS** with `left: None, right: Some("retry-with-jitter")`. Test genuinely guards the fix.
- Restored to `continue`. `cargo check --workspace` clean (only pre-existing dead-code warnings in `provekit-lsp`, unrelated).

### Follow-up (do NOT block on this — architect call)
Production `lift_pass.rs:163` calls `lift_file` (no `source_text`), so Path B (raw `// concept:` source-line scan) is currently inert in production. Two paths exist on a single code path, only one fires. Architect can decide whether to wire `source_text` through or excise Path B. The fix shipped here is correct and necessary regardless: Path A was the production path and it was broken; now it isn't.

**MERGE.**

PR #715, opened 2026-05-12
feat(ir-symbolic): predicate parse round-trip + smoke-test stub-1 unswap

TSavo:
## Summary

- Adds `provekit_ir_symbolic::parse_expr(s) -> Result<Rc<Formula>, ExprParseError>`: a recursive-descent parser for Rust-expression-shaped predicate strings. Produces structurally identical `Formula` trees to the kit's authoring API (`gt`, `and_`, `lte`, etc.). Operator mapping exactly matches the kit's Unicode names (`<=` → `"≤"`, `>=` → `"≥"`, `!=` → `"≠"`). Covers `||`/`&&`/`!` with correct precedence, dotted-path identifiers (`result.is_ok`), integer/boolean literals, parenthesised sub-expressions, and `ExprParseError { position, kind }` error reporting.
- Re-wires `smoke-test-e2e/driver/src/main.rs::formula_text_to_value` to use `parse_expr` + `serialize::formula_to_value` instead of the single-atom shim. Falls back with loud warning for unparseable strings.
- Re-wires `synthesize.rs::SmokeTestResolver` wp_rules to parse from human-readable strings via `parse_expr` + `convert::formula_to_ir`.
- Updates `report.md` §8: Stub 1 marked `[CLOSED]`. All three stubs are now closed.

## Test plan

- [ ] `cargo test -p provekit-ir-symbolic`: 49/49 pass (29 new parse_expr tests covering every operator, precedence, dotted paths, bool literals, 5 error cases)
- [ ] `cargo test -p smoke-test-e2e`: 1/1 pass
- [ ] `cargo check --workspace`: no errors introduced
- [ ] No new compiler warnings in modified crates

## Notes for reviewer

Substrate PR per standing rule — needs Opus deep review before merge.

Round-trip property tested: `jcs(formula_to_value(parse_expr(s)))` == `jcs(formula_to_value(parse_formula(json_parse(jcs(formula_to_value(parse_expr(s)))))))`. Byte-identical at the JCS level.


TSavo:
## Adversarial review — MERGE-WITH-NITS

Probes from Sir's review brief, all empirically verified against worktree at `bfead713`.

### Verified green

- **Probe 1 — precedence boundaries.** Wrote a scratch integration test (`tests/pr715_adversarial.rs`, not committed). All three of Sir's stated precedence rules pass:
  - `!x > 0` parses as `Not(GreaterThan(x, 0))` (matches the in-tree test `precedence_not_binds_tighter_than_and`).
  - `a && b || c` parses as `Or(And(a, b), c)`.
  - `a == b && c` parses as `And(Eq(a, b), c)`.
  Grammar (lines 14-26 of `parse_expr.rs`) is correctly stratified `or → and → not → cmp → atom`.

- **Probe 2 — round-trip byte-stability.** `assert_round_trip` in the inline tests is a meaningful check: it parses string `s`, JCS-encodes via `formula_to_value`, re-parses the JSON via `parse::parse_formula`, JCS-encodes again, asserts equality. Not a self-equality of the same value. 22 round-trip cases exercise it.

- **Probe 3 — identifier grammar.** Empirically: `a.b.c.d.e` works; `my_var` works; trailing `foo.` is `UnexpectedToken` (lexer's dotted-path arm requires another ident after the `.`); leading `.bar` is `UnexpectedToken`; `123abc` is an error.

- **Probe 4 — error positions.** Byte offsets, not char offsets (lexer maintains an explicit `byte_offsets` table). `(a && b` → `UnmatchedParen { position: 0 }` (at the open paren). `x @ 0` → `UnexpectedToken { position: 2, token: "@" }`.

- **Probe 5 — what's not supported.** Arithmetic `x + 1`, function calls `f(x)`, string literals `"hello"` all return errors as expected. Smoke-test driver hit the fallback exactly once on `out == if items_len == 0 { 0 } else { 1 }` (the `if/else` expression), loudly labelled — the only predicate in the fixture that needs the fallback.

- **Probe 6 — smoke-test re-wire.** `cargo run --manifest-path driver/Cargo.toml` runs clean: 9 sites, 6 concepts, verdicts identical to pre-PR (4 `exact`, 2 `loudly-bounded-lossy` annotation-lift/test-lift sites that have no wp rule, 3 `refuse` empty-contract sites). One eprintln fallback warning fires on the `if/else` predicate as documented.

- **Probe 7 — no shadowing.** `cargo check --workspace`: no errors. Existing `parse::parse_formula` still resolves and works; written a probe `p7_parse_formula_still_callable` to confirm both entry points are callable independently.

- **Probe 8 — smoke test green.** `cargo test -p provekit-ir-symbolic --lib`: 49/49 pass. `cargo test --manifest-path menagerie/smoke-test-e2e/Cargo.toml`: 1/1 pass (`commit_balance_change_post_is_non_negative`).

### Nits (non-blocking)

1. **`UnexpectedEof.position` is `last_token.pos + 1`, not `input.len()`.** For `x > 0 &&` (len 8), the reported EOF position is 7 (the byte after `&` at offset 6), not 8. Defensible — it points at "one past the start of the last token" — but `input.len()` would be more conventional for an EOF marker. Not a bug.

2. **Bare identifier in predicate position lifts as nullary atomic, not variable.** `parse_expr("foo")` returns `atomic_("foo", [])`, not a `Var` term. This is intentional (the grammar promises a Formula, not a Term) and load-bearing for `parse_expr("true")` / `parse_expr("false")` semantics — but worth documenting more explicitly in the doc comment for consumers who might expect `foo` to be a variable reference.

3. **Dead/unused warnings in `smoke-test-e2e-driver`** (`algebra.rs::render`, `FormulaShape::pretty`, `render_node`) are pre-existing and not introduced by this PR. Boy-scout opportunity for a follow-up.

### CI red — pre-existing, not caused by this PR

The "Cross-language conformance gate" check fails on `c11-cursorkind-check` complaining `menagerie/c11-language-signature/{README.md,component-cids.json}: stale`. This PR touches zero files in that directory; the staleness is upstream of the base branch (`smoke-test/end-to-end-vision-demo`). CodeRabbit returned SUCCESS; both `prove` jobs returned SUCCESS.

### Verdict: MERGE-WITH-NITS

Per Supra omnia, rectum: the parser produces structurally identical IR to the kit's authoring API (byte-identical JCS), the round-trip is real, the precedence is correctly stratified, the fallback is loud, and Stub 1 is genuinely closed in the report. Substrate quality.

Nits 1 and 2 are documentation improvements, not correctness defects. Nit 3 is housekeeping. CI red is base-branch baggage that needs a separate fix in `c11-language-signature/`.

Ship.

PR #719, opened 2026-05-12
feat(lift-rust-tests): emit EvidenceMemento alongside ContractDecl (PR-C of #716)

TSavo:
## Summary

- Extends `provekit-lift-rust-tests` (Rust's existing mirror of `provekit-lift-java-junit`) with a new `lift_file_with_evidence` API that mints one `EvidenceMemento` per lifted callsite assertion, alongside the existing `ContractDecl` output.
- All three coordinator-correction messages converge on this crate: it already walks `#[test]` functions, resolves callsite-keyed let-bindings, and emits per-callsite contracts. No new lifter needed; no `provekit-lift-contracts` extension (annotations != test-assertions, separate `source_kind`).
- `provekit-walk`'s MVP callsite tracing is narrower than what `lift-rust-tests` already does; no new dep there.

## New public API

```rust
// Pure (no filesystem I/O): caller supplies raw bytes they already have.
pub fn lift_file_with_evidence(
    file: &syn::File,
    source_path: &str,
    source_bytes: &[u8],
) -> AdapterOutput

// AdapterOutput gains:
pub evidences: Vec<EvidenceMemento>   // empty for the basic lift_file path
```

## EvidenceMemento fields

| Field | Value |
|---|---|
| `source_kind` | `SourceKind::TestAssertion` |
| `confidence_basis_points` | `10000` (static-derived) |
| `lifter_cid` | auto-promote sentinel (`blake3-512:000...000`) per spec §4.4 |
| `predicate` | `IrFormula` round-tripped via JCS from the lifted `ir_symbolic::Formula` |
| `source_locator.span` | 1-based col (proc_macro2 col is 0-indexed; +1 applied) |
| `source_locator.source_cid` | `BLAKE3-512(source_bytes)` |
| `extension_fields["test_function_name"]` | name of the `#[test]` fn |
| `extension_fields["target_callsite_symbol"]` | `<callee>@<file>:<line>:<col>` |
| `cid` | `BLAKE3-512(JCS(header sans cid key))` |

## CID derivation

Mirrors the `mint_contract` pattern in `provekit-claim-envelope`: builds a `canonicalizer::Value` tree for all fields (eliding `cid` by not inserting the key), calls `encode_jcs`, then `blake3_512_of`. No serde_json round-trip for the CID bytes.

## Test plan

- 36/36 tests green (`cargo test -p provekit-lift-rust-tests`)
- 11 new evidence tests:
  - [ ] `lift_with_evidence_emits_evidence_alongside_decls` -- both `decls` and `evidences` populated
  - [ ] `evidence_source_kind_is_test_assertion` -- correct `SourceKind`
  - [ ] `evidence_fields_are_correct` -- all header fields have the right values
  - [ ] `evidence_extension_fields_contain_test_function_name` -- `test_function_name` key
  - [ ] `evidence_extension_fields_contain_target_callsite_symbol` -- `target_callsite_symbol` key with correct prefix
  - [ ] `evidence_source_locator_has_1based_col` -- col >= 1 (catches the 0-indexed trap)
  - [ ] `evidence_source_cid_matches_source_bytes_hash` -- source_cid = BLAKE3-512(bytes)
  - [ ] `evidence_cid_is_deterministic` -- same source = same CID on two calls
  - [ ] `evidence_three_assertions_yields_three_evidences` -- one per callsite, all CIDs distinct
  - [ ] `lift_with_evidence_skips_literal_only_assertion` -- skip behavior preserved
  - [ ] `basic_lift_does_not_populate_evidences` -- backward compat: `lift_file` leaves `evidences` empty

### DIFF
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1947,6 +1947,7 @@ dependencies = [
  "provekit-ir-types",
  "provekit-lift",
  "provekit-verifier",
+ "provekit-walk",
  "quote",
  "serde_json",
  "syn",
@@ -2132,6 +2133,23 @@ dependencies = [
  "thiserror",
  "toml 0.8.23",
  "walkdir",
+]
+
+[[package]]
+name = "provekit-walk"
+version = "0.1.0"
+dependencies = [
+ "base64",
+ "libprovekit",
+ "proc-macro2",
+ "provekit-canonicalizer",
+ "provekit-claim-envelope",
+ "provekit-ir-types",
+ "provekit-proof-envelope",
+ "quote",
+ "serde_json",
+ "syn",
+ "thiserror",
 ]
 
 [[package]]
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1942,10 +1942,13 @@ name = "provekit-lift-rust-tests"
 version = "0.1.0"
 dependencies = [
  "proc-macro2",
+ "provekit-canonicalizer",
  "provekit-ir-symbolic",
+ "provekit-ir-types",
  "provekit-lift",
  "provekit-verifier",
  "quote",
+ "serde_json",
  "syn",
 ]
 
--- a/implementations/rust/provekit-ir-types/src/lib.rs
+++ b/implementations/rust/provekit-ir-types/src/lib.rs
@@ -1047,7 +1047,7 @@ pub struct ConceptSiteMemento {
 // round-trip tests only.
 // ============================================================
 
-/// One point inside a source span: 1-based line / col.
+/// One point inside a source span: 1-based line, 0-indexed col (UTF-8 bytes; spec §1.1).
 ///
 /// Locked JCS key order: col, line.
 ///
--- a/implementations/rust/provekit-lift-rust-tests/Cargo.toml
+++ b/implementations/rust/provekit-lift-rust-tests/Cargo.toml
@@ -10,6 +10,7 @@ description = "ProvekIt lift adapter: walk #[test] / #[tokio::test] functions an
 provekit-ir-symbolic = { path = "../provekit-ir-symbolic" }
 provekit-ir-types = { path = "../provekit-ir-types" }
 provekit-canonicalizer = { path = "../provekit-canonicalizer" }
+provekit-walk = { path = "../provekit-walk" }
 syn = { workspace = true }
 proc-macro2 = { workspace = true, features = ["span-locations"] }
 quote = { workspace = true }
--- a/implementations/rust/provekit-lift-rust-tests/Cargo.toml
+++ b/implementations/rust/provekit-lift-rust-tests/Cargo.toml
@@ -8,9 +8,12 @@ description = "ProvekIt lift adapter: walk #[test] / #[tokio::test] functions an
 
 [dependencies]
 provekit-ir-symbolic = { path = "../provekit-ir-symbolic" }
+provekit-ir-types = { path = "../provekit-ir-types" }
+provekit-canonicalizer = { path = "../provekit-canonicalizer" }
 syn = { workspace = true }
 proc-macro2 = { workspace = true, features = ["span-locations"] }
 quote = { workspace = true }
+serde_json = { workspace = true }
 
 [dev-dependencies]
 provekit-lift = { path = "../provekit-lift" }
--- a/implementations/rust/provekit-lift-rust-tests/src/lib.rs
+++ b/implementations/rust/provekit-lift-rust-tests/src/lib.rs
@@ -73,6 +73,8 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::rc::Rc;
 use std::sync::Arc;
 
+use std::collections::HashMap;
+
 use provekit_canonicalizer::{blake3_512_of, encode_jcs, Value as CValue};
 use provekit_ir_symbolic::{
     eq, gt, gte, lt, lte, make_var, ne, num, serialize::formula_to_value, str_const, ContractDecl,
@@ -81,6 +83,7 @@ use provekit_ir_symbolic::{
 use provekit_ir_types::{
     EvidenceMemento, IrFormula, SourceKind, SourceLocator, SourceLocatorPoint, SourceLocatorSpan,
 };
+use provekit_walk::emit::rust_function_term_json_cid;
 use syn::spanned::Spanned;
 
 /// The auto-promote sentinel lifter CID (128 hex zeros after the prefix).
@@ -148,12 +151,17 @@ pub(crate) struct LiftedCallsiteAssertion {
 }
 
 /// Extended version of `LiftedCallsiteAssertion` that also carries the
-/// callsite span -- needed for `SourceLocatorSpan` in `EvidenceMemento`.
+/// callsite span and raw callee name -- needed for `SourceLocatorSpan` and
+/// `test_target_function_cid` in `EvidenceMemento`.
 #[derive(Debug, Clone)]
 struct LiftedCallsiteAssertionWithSpan {
     name: String,
     formula: Rc<Formula>,
     callsite_span: proc_macro2::Span,
+    /// Raw callee name (e.g. `"deposit"` or `"Account::deposit"`). Used to
+    /// look up the target function in the same lift pass for Option A CID
+    /// computation (spec §10 `test_target_function_cid`).
+    callee: String,
 }
 
 #[derive(Debug, Clone)]
@@ -221,9 +229,36 @@ pub fn lift_file_with_evidence_and_skip(
     skip: &BTreeSet<String>,
 ) -> AdapterOutput {
     let source_cid = blake3_512_of(source_bytes);
+    // Build a map from bare function name to &syn::ItemFn for all non-test top-level
+    // functions in this file. Used by `visit_test_fn_with_evidence` to compute
+    // `test_target_function_cid` (spec §10) via Option A: same-lift-pass CID
+    // for free-fn callees whose definition is visible in this syn::File.
+    let fn_map: HashMap<String, &syn::ItemFn> = collect_non_test_fns(&file.items);
     let mut out = AdapterOutput::default();
-    walk_items_with_evidence(&file.items, source_path, &source_cid, skip, &mut out);
+    walk_items_with_evidence(&file.items, source_path, &source_cid, skip, &fn_map, &mut out);
     out
+}
+
+/// Collect all non-test `fn` items reachable from `items` (top-level and inside
+/// `mod` blocks) by bare name. Only free functions are included; method calls
+/// on `self` / receiver types cannot be resolved by name alone and fall through
+/// to the pending path.
+fn collect_non_test_fns<'a>(items: &'a [syn::Item]) -> HashMap<String, &'a syn::ItemFn> {
+    let mut map = HashMap::new();
+    for item in items {
+        match item {
+            syn::Item::Fn(f) if !has_test_attr(&f.attrs) => {
+                map.insert(f.sig.ident.to_string(), f);
+            }
+            syn::Item::Mod(m) => {
+                if let Some((_, inner)) = &m.content {
+                    map.extend(collect_non_test_fns(inner));
+                }
+            }
+            _ => {}
+        }
+    }
+    map
 }
 
 fn walk_items(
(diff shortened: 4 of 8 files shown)


TSavo:
## Adversarial review — REQUEST CHANGES

Branch `feat/716-PR-C-test-assertion-lifter` at `2f80f72`. Reviewed against `protocol/specs/2026-05-13-compound-contract-memento.md` (PR-A of #716). `cargo test -p provekit-lift-rust-tests` → 36/36 green. `cargo check --workspace` clean.

### What the PR gets right

- **Call-site attribution (load-bearing per Sir's directive): correct.** `lift_assertion_macro_at_callsites_with_span` calls `assertion_observed_exprs` on each assert side, then `collect_expr_callsites` walks `Expr::Call`/`MethodCall`/`Path` and resolves identifier paths through the `BTreeMap<String, BoundCall>` populated from preceding `let` statements. The contract name is `<callee>@<file>:<line>:<col>` where the locus is the producer call expression, not the test fn. The probe case `let mut a = Account::new(); a.deposit(100); assert_eq!(a.balance(), 100)` would attribute to `balance@...` (the assert side) and to `Account::new@...` / `deposit@...` IF the asserter side referenced the binding; literal-only sides correctly fall through and emit a `LiftWarning` per `skips_literal_assertion_without_callsite`. Architecture is sound. Note: `provekit-walk`'s call-site graph is NOT used; the in-crate let-binding tracker is sufficient and was the right architectural choice for v0.5.
- **CID determinism: correct.** `evidence_memento_cid` builds the canonicalizer `Value` header with the locked JCS key set, elides `cid`, calls `encode_jcs` + `blake3_512_of`. The `evidence_cid_is_deterministic` test confirms idempotence. JCS bytewise key order on `schemaVersion`/`source_kind`/`source_locator` resolves correctly (`schemaVersion` < `source_kind` at position 1: `c` 0x63 < `o` 0x6F).
- **source_cid: correct.** `BLAKE3-512(source_bytes)` directly via `provekit_canonicalizer::blake3_512_of`, no normalization or BOM stripping. `evidence_source_cid_matches_source_bytes_hash` pins it.
- **Scope discipline: clean.** Three files, single crate. No reach into `discharge.py`, `op_*.spec.json`, or anything outside `provekit-lift-rust-tests` + the dependencies it pulled in (`provekit-ir-types`, `provekit-canonicalizer`, `serde_json`).
- **Auto-promote sentinel lifter_cid: matches §4.4.** 128 hex zeros, accepted by pass-1 validation per spec.
- **`lift_file_with_evidence_and_skip` is real, not dead code.** It's the `skip` + evidence combination, invoked from `lift_file_with_evidence` with empty skip.

### Spec violations (blockers)

**1. `col` is 1-indexed in the emitted evidence; spec §1.1 mandates 0-indexed.**

Spec, line 143 of the compound-contract memento doc: *"`col` counts UTF-8 BYTES within the line, 0-indexed. Line numbers are 1-indexed."* The PR (lib.rs §414, §418) does:

```rust
col: (span_start.column + 1) as u32,
```

The comment claims this is "1-based" per the `SourceLocatorPoint` doc-comment, but the spec — the normative source — says 0-indexed for `col`. The IR types module's doc-comment at line 1050 (`"1-based line / col"`) is the document that needs correcting, not the spec. Cross-language CIDs (Java/Python lifters reading the same spec) WILL diverge from Rust if Rust ships `+1`. The in-tree precedent `provekit-walk/src/locus.rs` uses `start.column` raw (no +1) — this PR is the outlier within the codebase as well.

The test `evidence_source_locator_has_1based_col` pins the WRONG behavior; it should assert `col >= 0` and the lifter should drop the `+1`.

Fix: remove the two `+1` increments on column. The IR types' doc-comment should be amended to say "1-based line, 0-indexed col (UTF-8 bytes)" — ideally in this PR, since the discrepancy is what tripped the implementation. Line stays as-is (proc_macro2's `line` is 1-indexed, matching the spec).

**2. `extension_fields` is missing the spec-required `test_target_function_cid`; ships unspecified keys instead.**

Spec §10 inventory table, `test-assertion` row: **Required `extension_fields` = `{ "test_target_function_cid": <cid> }`**. Spec §4.2 step 2 explicitly says the compound-mint pipeline gathers evidences *"whose `extension_fields.test_target_function_cid` (or equivalent per-kind back-link, §10) is `function_term_cid`"*. Without this key, the evidences emitted here are orphaned from compound-mint and from the load-bearing example in spec §10 (rows 3 and 4 of the safe_div worked example explicitly cite `test_target_function_cid`).

The PR instead emits `test_function_name` and `target_callsite_symbol` — neither is in the spec inventory. The `target_callsite_symbol` value (`<callee>@<file>:<line>:<col>`) is a STRING SYMBOL, not a CID; it cannot stand in for a content-addressed back-link.

Acknowledged design tension: the lifter sees test source, not production source, so the FunctionContractMemento CID for the target may not be locally derivable. Reasonable resolutions:
- (a) Compute `test_target_function_cid` via a deterministic synthesis from the callsite symbol (a stub CID that PR-F's compound mint will reconcile when the real FCM is in the pool); or
- (b) Emit it as `null`/omit, document the constraint, and have PR-F's compound mint backfill — but the spec marks it Required, so emitting MUST happen at lift time.

If the answer is "we genuinely cannot derive it in PR-C, spec §10 row must be amended," that amendment belongs in this PR's spec-side companion or as an open spec issue, NOT as a silent omission with a substituted key. As-shipped, downstream consumers reading §10 will look for the field and not find it.

### Other concerns (non-blocking, but flag)

- **Spec §1.1 §3.1 cid_input includes `cid` field at the top of the alphabetical list**: spec text says "cid is elided." Code correctly elides. Confirmed.
- **`SourceLocatorPoint` doc-comment at `provekit-ir-types/src/lib.rs:1050` says "1-based line / col"** — this contradicts the compound-memento spec. Whichever way (1) resolves, this doc-comment needs to align with the spec.
- **`serde_json_to_cvalue` floats fallback to string**: defensible (extension_fields shouldn't contain floats in v0), but if a future extension_fields contributor stores a float, the silent string conversion will produce a different CID than a Java lifter using a numeric JCS encoding. Worth a TODO/assert.
- **`formula_to_ir_formula` panics on round-trip failure.** Documented as "should always succeed by construction." That's fine for v0 but a `.expect()` on JCS bytes is a hard panic in production; a `LiftWarning` path would be more graceful.

### Verdict

Architecturally on target on the load-bearing point (probe 1). Two normative spec violations on (a) col indexing and (b) required `extension_fields` key. Both will roll the CID across language implementations, which is the exact failure mode the "Supra omnia, rectum" / cross-language federation rule exists to prevent.

REQUEST CHANGES on:
1. Drop `+1` on `col`; align test `evidence_source_locator_has_1based_col` to `col >= 0`; amend the `SourceLocatorPoint` doc-comment.
2. Emit `test_target_function_cid` in `extension_fields`; OR amend spec §10 in this PR's companion to legitimize the substitute keys; OR open a tracked follow-up before merge with explicit spec coverage.

Merge unblocks once the bytes match the spec.

TSavo:
Both blockers from the Opus deep review resolved in commit `0ceb1614`.

| # | Blocker | Resolution |
|---|---------|-----------|
| B1 | `col` must be 0-indexed (spec §1.1) | Removed `+1` at lib.rs:414/418. `proc_macro2`'s raw `start.column` / `end.column` used directly (already 0-indexed). Updated `SourceLocatorPoint` doc-comment in `provekit-ir-types/src/lib.rs` from "1-based line / col" to "1-based line, 0-indexed col (UTF-8 bytes; spec §1.1)". Replaced the wrong `evidence_source_locator_has_1based_col` test (was asserting `col >= 1`, pinning the bug) with `evidence_source_locator_col_is_0indexed_spec_b1` which pins the exact expected value (`col=11` for `assert_eq!(f(0), 0)`) and would catch any future `+1` re-introduction. Aligns with `provekit-walk/src/locus.rs` which already used raw `start.column`. |
| B2 | `test_target_function_cid` missing from `extension_fields` | Added `provekit-walk` as a dep. In `lift_file_with_evidence_and_skip`, builds `HashMap<name, &syn::ItemFn>` for all non-test fns in the same `syn::File`. Added `callee: String` to `LiftedCallsiteAssertionWithSpan`. **Option A** (same-file free fn): calls `rust_function_term_json_cid(item_fn, source_path)` for a real `blake3-512:` CID. **Option B** (cross-crate or method call): emits `"pending:<symbol>"` -- clearly-non-CID marker, does NOT pretend the binding is satisfied. `test_function_name` and `target_callsite_symbol` retained for debugging. Added two regression tests: `evidence_test_target_function_cid_present_and_valid_spec_b2` (Option A: verifies CID is `blake3-512:` prefix + 128 hex, matches independently-computed function-term CID) and `evidence_test_target_function_cid_pending_when_cross_crate_spec_b2` (Option B: verifies `pending:` prefix, no false CID). |

**Boy-scout:** Fixed four pre-existing `non-exhaustive patterns` compile errors in `provekit-walk` (`dropper/predicate.rs`, `dropper/predicates/not_null.rs`, `lift.rs`, `marriage.rs`) caused by `IrFormula::Substitute` + `Apply` variants added in PR-A of this branch. These were blocking the `provekit-walk` dep from building under `provekit-lift-rust-tests`.

`cargo test -p provekit-lift-rust-tests`: **39 passed, 0 failed** (38 unit + 1 integration).
`cargo check --workspace`: clean.

TSavo:
## Adversarial re-review @ 0ceb1614 — MERGE-WITH-NITS

**Verdict:** MERGE-WITH-NITS. B1 and B2 are both empirically closed. The `pending:<symbol>` form is well-shaped (loudly-bounded-lossy per Supra omnia, rectum), but ships without a spec backstop; needs a paper-trail follow-up before downstream PRs rely on it.

### B1 — col is 0-indexed (spec §1.1 normative): CLOSED

- `+1` removed at `provekit-lift-rust-tests/src/lib.rs:467` and `:471`. Verified.
- Doc-comment at `provekit-ir-types/src/lib.rs:1050` now reads "1-based line, 0-indexed col (UTF-8 bytes; spec §1.1)". Verified.
- Old `evidence_source_locator_has_1based_col` REMOVED (was pinning the bug). Verified absent in test list.
- New `evidence_source_locator_col_is_0indexed_spec_b1` pins exact `col == 11` for `"assert_eq!("` (11 bytes). Verified.
- Regression-catch check: pre-fix code emits `column + 1 = 12`; test asserts `== 11`; would fail pre-fix, passes post-fix. Confirmed catches regressions.

### B2 — test_target_function_cid populated (spec §10 REQUIRED): CLOSED

- `provekit-walk` added as dep in `provekit-lift-rust-tests/Cargo.toml:13`. No cycle (walk does not depend on lift-rust-tests).
- `LiftedCallsiteAssertionWithSpan.callee: String` field added at lib.rs:163.
- `collect_non_test_fns()` builds `HashMap<String, &syn::ItemFn>` for same-file non-test fns. Recurses into `mod` blocks.
- Same-file callee path: calls `provekit_walk::emit::rust_function_term_json_cid` and emits real `blake3-512:` CID.
- Cross-crate / method-receiver path: emits `"pending:<symbol>"`.
- `evidence_test_target_function_cid_present_and_valid_spec_b2` asserts Option A produces a valid `blake3-512:` CID + matches the independently-computed function-term CID. Verified.
- `evidence_test_target_function_cid_pending_when_cross_crate_spec_b2` asserts Option B emits `pending:` and explicitly NOT `blake3-512:`. Verified.

### B3 — `pending:<symbol>` v0 disposition: SHAPE-OK, SPEC-GAP

Spec §10 column `Required extension_fields` literally reads `{ "test_target_function_cid": <cid> }` for `test-assertion`. The one existing sentinel in the spec (§4.4 auto-promote `lifter_cid`, 128 hex zeros) is shape-COMPATIBLE with a CID. `pending:<symbol>` is shape-INCOMPATIBLE by design.

This is actually the correct shape per Sir's own architecture memos:
- "loudly-bounded-lossy transformation is a first-class legitimate outcome, the loss-characterization IS its contract"
- Distinguishing `pending:` from `blake3-512:` makes consumers able to detect-and-defer in-band, unlike the original `function_symbol` violation that smuggled a string as a CID

But the spec does not yet sanction it. No gap-memento filed. Acceptable v0 IFF the paper trail is created in a follow-up. Per Sir's standing integrator-merge delegation, this is a merge-with-required-followup, not a blocker.

### Boy-scout match arms: SOUND

`provekit-walk/{dropper/predicate.rs, dropper/predicates/not_null.rs, lift.rs, marriage.rs}` add arms for `IrFormula::Substitute` and `IrFormula::Apply`. All return `None`/`false` (conservative), mirroring how the SAME functions handle `Forall`/`Exists`/`Choice`. No `unreachable!()` panics. Structurally exhaustive. Sound.

### Tests + workspace

- `cargo test -p provekit-lift-rust-tests --release`: **39 passed, 0 failed** (38 unit + 1 round_trip). Matches claim.
- `cargo check --workspace`: clean. Only pre-existing dead-code warnings (provekit-lsp config fields, ir-symbolic unused helpers, etc.). No errors. No new warnings from this PR.

### Required follow-ups (non-blocking under integrator delegation)

1. **Spec or gap-memento for `pending:`.** Either: amend `2026-05-13-compound-contract-memento.md` §10's `test-assertion` row to allow `pending:<symbol>` as a non-CID v0 placeholder; OR file a known-limitation memento under `protocol/specs/` that PR-F (compound-aware discharge) explicitly resolves. The form is well-shaped; the paper trail is missing.

2. **PR body stale.** Description still says "1-based col (proc_macro2 col is 0-indexed; +1 applied)" — contradicts the B1 fix. Update before merge.

3. **Minor (consider for PR-F):** `extension_fields.target_callsite_symbol` (`<callee>@<file>:<line>:<col>`) and `pending:<callee>` overlap on the callee identity. If PR-F's resolver consumes `pending:` to look up the function across crates, consider whether the locator-bearing `target_callsite_symbol` makes the bare `pending:<callee>` redundant. Not a fix needed here — flag for PR-F's design.

Cleared for merge.

PR #717, opened 2026-05-12
feat(domain-claim): normalize substrate surface to k(I)=t (PR-A of N)

TSavo:
## Summary

Sir's 2026-05-12 architectural directive: **"The code needs to literally read k(I)=t at a high level of abstraction."**

The substrate's organizing equation is `k(I) = t`: apply key `k` (an operation, content-addressed) to input `I` (an artifact, content-addressed), and you get truth-claim `t` (a content-addressed concept, target-language source, shape, or other canonical truth). The trichotomy decides whether the equation holds exactly, holds with bounded loss, or refuses.

Today the verifier has per-memento-type knowledge. It dispatches differently on `ConceptSiteMemento`, on `RealizationDesugaringMemento`, on `FunctionContractMemento`, on `MorphismDischargeReceipt`. As new memento types land (`CompoundContractMemento` in PR #716, etc.), the verifier needs new dispatch arms. The substrate's primary equation is aspirational at the consumer surface.

This PR introduces the canonical wire-form `DomainClaim` type. Every memento decomposes into a `DomainClaim` via an `Into<DomainClaim>` impl. After PR-B, the verifier consumes only `DomainClaim`s; per-memento-type knowledge becomes thin From-impls. The verifier's surface is literally `k(I) = t` with the verdict attached, in code as well as in prose.

This is **PR-A**: spec + Rust types + serde round-trip tests + the one source-side `From` impl whose source memento lives in `provekit-ir-types`. No verifier refactor yet; PR-B does that.

## What this PR ships

- **Spec** `protocol/specs/2026-05-13-domain-claim-normalization.md` (~330 lines):
  - §1 CDDL schema (DomainClaim, VerdictKind, VerdictBody, DomainClaimProvenance)
  - §1.2 verdict consistency invariants (the §5.2 ConceptSite invariants lifted to the wire form)
  - §2 normative mapping table (per-memento-type `Into<DomainClaim>` semantics)
  - §3 CID derivation (BLAKE3-512 over JCS with `signature` field elided; signer-independent addressing)
  - §3.2 multi-input tupling
  - §4 verifier consumption preview
  - §5 backward-compat deprecation shim
  - §6 PR roadmap (PR-A..PR-D) including **conditional rows** for memento types not on main (#716 CompoundContractMemento/EvidenceMemento, MorphismDischargeReceipt-pre-Rust-type)
  - §7 cross-language compatibility (the wire form is language-agnostic; Java/Python/TS verifiers consume the same JSON)
[...]

PR #720, opened 2026-05-12
feat(domain-claim): TryFrom<&FunctionContractMemento> for DomainClaim (PR-B of #717)

TSavo:
## Summary

PR-B of the DomainClaim normalization chain (#717 / PR-A landed at ee3fa75a).

Bare `FunctionContractMemento`s have no inline discharge. They must be wrapped in a `ConceptSiteMemento` (the binding) or a `CompoundContractMemento` (PR #716) before reaching the verifier surface (spec §2.3, §6.1). This impl lives in `libprovekit` rather than `provekit-ir-types` because `FunctionContractMemento` is defined in `libprovekit::compose`; placing it in `provekit-ir-types` would invert the dependency direction.

The `TryFrom` impl is intentionally always-`Err` so the type system encodes the invariant at the call site: callers are forced to handle the error and either reject the bare contract or promote it through a binding.

## Changes

- **`provekit-ir-types/src/lib.rs`**: add `UnboundContract` variant to `DomainClaimConversionError` with a spec-citing Display message
- **`libprovekit/src/compose.rs`**: `impl TryFrom<&FunctionContractMemento> for provekit_ir_types::DomainClaim` -- always returns `Err(UnboundContract)`
- **`libprovekit/src/compose.rs`**: 4 unit tests covering error variant identity, determinism across two calls on the same FCM, Display message quality, and pattern-match soundness

## Test plan

- [ ] `cargo test -p libprovekit domain_claim_fcm` -- 4 tests, all green
- [ ] `cargo check --workspace` -- clean (warnings are pre-existing dead code in unrelated crates)
- [ ] Spec compliance: §2.3 says bare contracts are NOT consumable directly; this impl enforces that at the type level
- [ ] Needs Opus deep review before merge (standing rule)

## Notes

The task description's field mapping (`lifter_cid`, `source_cid`, `concept_cid`) does not match the actual `FunctionContractMemento` struct fields. The normative source is spec §2.3 + §6.1, which are unambiguous: bare FCMs always return `Err(UnboundContract)`. The always-Err shape is the correct and complete implementation.

### DIFF
--- a/implementations/rust/libprovekit/src/compose.rs
+++ b/implementations/rust/libprovekit/src/compose.rs
@@ -1082,3 +1082,95 @@ pub fn jcs_bytes_of_value(v: &Value) -> Vec<u8> {
 pub use crate::wp::{
     free_vars_formula, free_vars_term, substitute_in_formula, substitute_in_term,
 };
+
+// ============================================================
+// DomainClaim projection for FunctionContractMemento (PR-B of #717)
+//
+// Bare `FunctionContractMemento`s have no inline discharge. They MUST
+// be wrapped in a `ConceptSiteMemento` (the binding) or a
+// `CompoundContractMemento` (PR #716) before reaching the verifier
+// surface. This impl is intentionally always-Err so the type system
+// encodes the invariant: the call-site is forced to handle the error
+// and either reject the bare contract or promote it through a binding
+// (spec §2.3, §6.1).
+// ============================================================
+
+impl TryFrom<&FunctionContractMemento> for provekit_ir_types::DomainClaim {
+    type Error = provekit_ir_types::DomainClaimConversionError;
+
+    fn try_from(_m: &FunctionContractMemento) -> Result<Self, Self::Error> {
+        Err(provekit_ir_types::DomainClaimConversionError::UnboundContract)
+    }
+}
+
+#[cfg(test)]
+mod domain_claim_fcm_tests {
+    use super::*;
+    use provekit_ir_types::{DomainClaimConversionError, IrFormula};
+
+    /// Build a minimal but valid `FunctionContractMemento` for testing.
+    /// `pre` and `post` use `And { operands: [] }` which is the canonical
+    /// encoding of the trivial `true` formula.
+    fn minimal_fcm() -> FunctionContractMemento {
+        FunctionContractMemento {
+            fn_name: "test_fn".to_string(),
+            formals: vec![],
+            formal_sorts: vec![],
+            formal_regions: vec![],
+            return_sort: provekit_ir_types::Sort::Primitive {
+                name: "bool".to_string(),
+            },
+            return_region: None,
+            pre: IrFormula::And { operands: vec![] },
+            post: IrFormula::And { operands: vec![] },
+            body_cid: None,
+            effects: EffectSet::default(),
+            locus: Locus {
+                file: Some("test.rs".to_string()),
+                line: 1,
+                col: 0,
+            },
+            canonical_bytes: vec![],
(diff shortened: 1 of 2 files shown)


TSavo:
**Verdict: MERGE-WITH-NITS** (Opus deep review)

## Spec compliance

The always-`Err(UnboundContract)` reading is the only viable one. Three converging arguments:

1. **§1.2 trichotomy is closed.** `VerdictKind` ∈ {exact, loudly-bounded-lossy, refuse}. §2.3's "special 'unverified' sentinel that the wire form rejects" is exactly what this is: it does not exist on the wire, so no valid `DomainClaim` is constructible from a bare FCM.
2. **`FunctionContractMemento` has no `kit_cid`-bearing field.** §2.3's hand-wave ("carried in `locus` or `auto_minted_mementos` provenance") doesn't match the struct on `main`: `locus` is `{file, line, col}`, `auto_minted_mementos: Vec<AliasingMemento>`. There is no path to extract `k` from a bare FCM today.
3. **PR-A precedent.** `DomainClaimConversionError` already shipped in PR-A with `InvalidVerdictString` + `StandaloneRealization`, refining §6.1's literal "`Into`" to `TryFrom`. Adding `UnboundContract` continues that pattern; the `TryFrom`-vs-`Into` discrepancy is not new with this PR.

## Architectural rightness (load-bearing)

The non-obvious-but-correct call is **always-Err vs Ok(verdict=Refuse, refusal_reason="undischarged")**. Conflating them would smuggle a semantic confusion onto the wire: `Refuse` means the operator tried and could not show `k(I)=t`; "unbound" means the object isn't ready to make any verdict at all. A downstream verifier reading `Refuse` would think the operation was attempted. The type-level `Err` makes the precondition violation impossible to lose. Supra omnia, rectum.

## Verification

- `cargo test -p libprovekit domain_claim_fcm` → 4/4 green (`bare_fcm_returns_unbound_contract_error`, `bare_fcm_error_is_deterministic`, `unbound_contract_display_is_informative`, `unbound_contract_error_variant_matches`).
- `cargo check --workspace` → clean (only pre-existing `provekit-lsp` dead-code warnings).
- Display impl cites spec §2.3 and names both binding types (`ConceptSiteMemento` / `CompoundContractMemento`). Caller guidance is clear.
- Tests cover variant identity, determinism, display quality, pattern-match soundness — the four claimed cases land.

## Forward compat with #716

Clean. When `CompoundContractMemento` lands, its conversion (per §6.2) operates on the compound, not on the bare FCM. The always-Err FCM impl correctly forces callers off the bare-FCM path and onto compound or `ConceptSiteMemento` — exactly the redirect §2.3 prescribes.

## Nits (non-blocking)

1. **Follow-up issue worth filing:** §2.3's "carried in `locus` or `auto_minted_mementos` provenance" doesn't match the FCM struct. Either the spec should be amended to acknowledge that bare FCMs are structurally unable to carry `kit_cid` (which strengthens the always-Err mapping), or a future PR adds a `lifter_cid: String` field to `FunctionContractMemento` (which would still need a discharge to project — so the always-Err for bare-FCM-without-binding still stands).
2. **Test coverage tightener:** `unbound_contract_display_is_informative` could additionally assert the Display string contains `"§2.3"` (it does, in the impl). Minor; leave for boy-scouting later.
3. **Naming:** PR title says "PR-B of #717"; spec §6.1's PR-B also includes the `verify(&DomainClaim)` API + `From<&core::types::DomainClaim>` work. Worth a brief comment in the PR body that this is PR-B/1 of N, with the verifier refactor + core-types conversion following.

## Disposition

Approve. Merge.

PR #721, opened 2026-05-12
feat(lift-contracts): type-signature evidence lifter (PR-D of #716)

TSavo:
## Summary

- Extends `provekit-lift-contracts` with a new `lift_file_with_sig_evidence` entry point that walks every function in a `syn::File` (not just annotated ones) and emits one `EvidenceMemento` per type-derived contract site.
- Fills the gap identified in #716: type signatures encode partial contracts that are statically complete for their domain and should flow into the compound-contract substrate.
- Does NOT touch `provekit-lift-rust-tests` (PR-C's home) or any other crate. Existing `lift_file` / `lift_file_with_source` paths are unchanged; `AdapterOutput.evidences` is empty on those paths.

## Covered type shapes

| Return / param type | `type_shape` | Predicate | Extra extension_fields |
|---|---|---|---|
| `-> ()` (implicit or explicit) | `unit` | `true` | — |
| `-> Option<T>` | `option` | `true` | `inner_type` (if concrete) |
| `-> Result<T,E>` | `result` | `true` | `ok_type`, `err_type` (if concrete) |
| `-> Vec<T>` | `vec` | `true` | `element_type` (if concrete) |
| `-> &T` / `-> &mut T` | `borrow` / `borrow-mut` | `true` | `inner_type`, `lifetime` |
| `&self` | `receiver-shared` | `true` | — |
| `&mut self` | `receiver-exclusive` | `true` | — |
| `self` (owned) | `receiver-owned` | `true` | — |
| `x: NonZeroU32` (all NonZero* family) | `non-zero` | `ne(x, 0)` | `inner_type` |
| `x: i32` / `u64` / `bool` / etc. | `bounded-primitive` | `true` | `inner_type` |
| `x: &T` / `x: &mut T` | `borrow` / `borrow-mut` | `true` | `inner_type`, `lifetime` |
| bare generic `T` (single uppercase or `Fut`/`Item`/etc.) | — | skip | — |

Conservative semantics: when the type does not yield a derivable contract, no evidence is emitted.

## EvidenceMemento fields

| Field | Value |
|---|---|
| `source_kind` | `SourceKind::TypeSignature` ("type-signature" on wire) |
| `confidence_basis_points` | `10000` (static-derived) |
| `lifter_cid` | `blake3-512:000...000` auto-promote sentinel (§4.4) |
| `predicate` | `atomic_("true", [])` for shape-only, or `ne(x, 0)` for NonZero params |
| `source_locator.span` | function-name token span (1-based col) |
| `source_locator.source_cid` | `BLAKE3-512(source_bytes)` |
| `extension_fields["function_symbol"]` | `"name@path"` plain symbol identifier (no realized function-term CID yet; mirrors PR-C's `target_callsite_symbol` pattern) |
| `extension_fields["signature_position"]` | `"return"` or `"param:N"` |
| `extension_fields["type_shape"]` | per-kind shape string |
| `cid` | `BLAKE3-512(JCS(header sans cid key))` |

## Design notes

- **`function_symbol` not `function_term_cid`**: there is no realized function-term CID at lift time. The field carries a deterministic `name@path` identifier following PR-C's `target_callsite_symbol` pattern. A future pass can replace it with a real CID once the function term is minted.
- **Trivial predicates for shape-only cases**: `atomic_("true", [])` is intentional for types like `Option`, `Vec`, `&T`. The IR has no native `is_some`/`is_ok` shape; the contract is expressed entirely through `type_shape` + `inner_type` in `extension_fields`. This is a loudly-bounded-lossy outcome per Supra omnia, rectum — the loss is characterized, not hidden.
- **CID helpers duplicated from PR-C**: `formula_to_ir_formula`, `serde_json_to_cvalue`, `evidence_memento_cid` are copied rather than sharing a crate. Known followup to factor into a shared helper crate.

## Test plan

- [x] `sig_option_return_produces_option_shape_evidence` — Option return → type_shape "option" + inner_type
- [x] `sig_result_return_produces_alt_shape_evidence` — Result return → type_shape "result" + ok_type + err_type
- [x] `sig_vec_return_produces_finite_list_evidence` — Vec return → type_shape "vec" + element_type
- [x] `sig_unit_return_produces_trivial_result_evidence` — explicit `-> ()` → type_shape "unit"
- [x] `sig_self_mut_self_produce_ownership_evidences` — `&self` → receiver-shared, `&mut self` → receiver-exclusive
- [x] `sig_non_zero_param_produces_refined_domain_evidence` — NonZeroU32 → non-zero, source_kind TypeSignature, confidence 10000
- [x] `sig_mixed_fn_produces_all_evidence_kinds` — fn with receiver + NonZero + i32 + Option return → all four shapes present
- [x] `sig_bare_generic_only_emits_no_evidence` — `fn identity(x: T) -> T` → empty evidences (conservative)
- [x] `sig_primitive_param_and_implicit_unit_return` — `fn sink(x: i32)` → bounded-primitive + unit
- [x] `sig_evidence_cid_is_deterministic` — same source bytes → same CIDs on two calls
- [x] `basic_lift_file_does_not_populate_evidences` — backward compat: `lift_file` leaves evidences empty
- [x] `sig_evidence_carries_function_symbol_not_cid` — every evidence has `function_symbol` = `"name@path"`, `function_term_cid` never appears

21/21 tests green. `cargo check --workspace` clean.

## Summary by CodeRabbit

* **New Features**
  * Added signature evidence lifting to analyze and classify function parameter and return types.
  * Support for type shape classification including references, `Option`, `Result`, and `Vec` types.
  * Evidence generation with deterministic content identification for contracts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/721)

### DIFF
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1836,8 +1836,11 @@ name = "provekit-lift-contracts"
 version = "0.1.0"
 dependencies = [
  "proc-macro2",
+ "provekit-canonicalizer",
  "provekit-ir-symbolic",
+ "provekit-ir-types",
  "quote",
+ "serde_json",
  "syn",
 ]
 
--- a/implementations/rust/provekit-lift-contracts/Cargo.toml
+++ b/implementations/rust/provekit-lift-contracts/Cargo.toml
@@ -8,6 +8,9 @@ description = "ProvekIt lift adapter: walk #[contracts::requires]/#[contracts::e
 
 [dependencies]
 provekit-ir-symbolic = { path = "../provekit-ir-symbolic" }
+provekit-ir-types = { path = "../provekit-ir-types" }
+provekit-canonicalizer = { path = "../provekit-canonicalizer" }
 syn = { workspace = true }
 proc-macro2 = { workspace = true, features = ["span-locations"] }
 quote = { workspace = true }
+serde_json = { workspace = true }
--- a/implementations/rust/provekit-lift-contracts/src/lib.rs
+++ b/implementations/rust/provekit-lift-contracts/src/lib.rs
@@ -42,12 +42,26 @@
 // `canonical_bytes` / CID derivation.  Changing or removing the
 // annotation never rewrites the shape identity.
 
+use std::collections::BTreeMap;
 use std::rc::Rc;
+use std::sync::Arc;
 
+use provekit_canonicalizer::{blake3_512_of, encode_jcs, Value as CValue};
 use provekit_ir_symbolic::{
-    and_, atomic_, eq, gt, gte, lt, lte, make_var, ne, num, str_const, ContractDecl, Formula, Int,
-    Sort, Term,
+    and_, atomic_, eq, gt, gte, lt, lte, make_var, ne, num, serialize::formula_to_value,
+    str_const, ContractDecl, Formula, Int, Sort, Term,
 };
+use provekit_ir_types::{
+    EvidenceMemento, IrFormula, SourceKind, SourceLocator, SourceLocatorPoint, SourceLocatorSpan,
+};
+
+/// The auto-promote sentinel lifter CID (128 hex zeros after the prefix).
+/// Used until PR-F wires the real lifter CID (compound spec §4.4).
+pub const AUTO_PROMOTE_LIFTER_CID: &str = concat!(
+    "blake3-512:",
+    "0000000000000000000000000000000000000000000000000000000000000000",
+    "0000000000000000000000000000000000000000000000000000000000000000",
+);
 
 #[derive(Debug, Clone)]
 pub struct LiftWarning {
@@ -59,6 +73,10 @@ pub struct LiftWarning {
 #[derive(Debug, Default)]
 pub struct AdapterOutput {
     pub decls: Vec<ContractDecl>,
+    /// One `EvidenceMemento` per lifted type-signature site.
+    /// Populated only when `lift_file_with_sig_evidence` is called; empty
+    /// when using the basic `lift_file` / `lift_file_with_source` API.
+    pub evidences: Vec<EvidenceMemento>,
     pub warnings: Vec<LiftWarning>,
     pub seen: usize,
     pub lifted: usize,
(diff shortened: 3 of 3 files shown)


TSavo:
## Adversarial review — CHANGES REQUIRED

Three blocking spec violations. Cargo green + 21/21 pin current (wrong) behavior; the tests are not sufficient.

### Blocker 1: missing `return_type` extension field (spec §10)

Spec `protocol/specs/2026-05-13-compound-contract-memento.md` §10 mandates, for `source_kind = "type-signature"`:

> Required `extension_fields`: `{ "return_type": <type_string> }`

The PR emits `type_shape`, `inner_type`, `ok_type`, `err_type`, `element_type`, `function_symbol`, `signature_position` — but no top-level `return_type: <type_string>` key on return-position evidences. That field is normatively required by §10. The other keys are fine as extension-open metadata, but `return_type` is non-negotiable.

Note this is the spec's own framing, not a parallel to PR-C #719. PR-D's `function_symbol` is allowed as an extension key (§10 last paragraph: "Each label is extension-open"); the missing field is `return_type`. Don't fix the wrong thing.

### Blocker 2: `col` is 1-indexed; spec §1.1 mandates 0-indexed

`lib.rs:197,201`:

```rust
col: (fn_span_start.column + 1) as u32,
...
col: (fn_span_end.column + 1) as u32,
```

Spec §1.1: "**Normative: column counting.** `col` counts UTF-8 BYTES within the line, **0-indexed**."

Empirical check on proc_macro2 1.x with `span-locations`: for source `"fn x() {}"`, `fn` token has `span().start().column == 0`. proc_macro2 is already 0-indexed at the API surface. The `+1` transformation actively breaks compliance for both `start` and `end`. Drop both `+ 1`s.

§4.5 reinforces: "`col` is 0-indexed UTF-8 byte count from the most recent LF (or start of file) to `byte_offset`." Two implementations must produce identical `{line, col}` from identical bytes; PR-D's offset-by-one will desync CIDs against any compliant lifter.

### Blocker 3: trivial `true` predicate for Option/Result contradicts spec §1.1.1 example

Spec §1.1.1 table row for `predicate` field gives an explicit normative example:

> "For an `Option<T>` return-type evidence, this is a predicate of the form `result.is_some() \/ result.is_none()`."

PR body says trivial `atomic_("true", [])` is "intentional ... loudly-bounded-lossy per Supra omnia, rectum." Reject. Supra omnia, rectum permits *loudly-bounded-lossy* outcomes when the loss is characterized and the contract can't be expressed; it doesn't permit downgrading an expressible predicate to `true` because the IR vocabulary needs a small extension. `or_(is_some(result), is_none(result))` is constructible from existing `IrFormula` primitives (or needs at most a one-line addition of `is_some` / `is_none` atomic predicates).

The `unit` case (`-> ()` ⇒ `true`) and `bounded-primitive` case are genuinely trivial — fine. The `Option` / `Result` / `Vec` / `borrow` cases are not; they encode real shape information that should land in the predicate, not just in extension_fields tags. Otherwise downstream dischargers see `true /\ true /\ true` and the type evidence contributes nothing to `composed_pre/post`.

Concrete asks:
- `Option<T>` ⇒ predicate `is_some(result) \/ is_none(result)` (per §1.1.1)
- `Result<T,E>` ⇒ predicate `is_ok(result) \/ is_err(result)`
- `NonZero*` parameter ⇒ already correct: `ne(x, 0)`
- `-> ()` and bare bounded primitive ⇒ `true` is genuinely trivial; keep

### Non-blockers / passes

- `confidence_basis_points = 10000` for type-signature: matches §10 prior. ✓
- `lifter_cid = AUTO_PROMOTE_LIFTER_CID` (128 hex zeros): matches §4.4 sentinel. ✓
- `source_kind = TypeSignature`: ✓
- Backward compat: `lift_file` path leaves `evidences` empty. ✓
- Bare generic conservative skip: ✓
- `cargo check --workspace`: clean. ✓
- 21/21 tests pass — but several tests assert the wrong invariants (1-indexed col, trivial predicates, missing `return_type`). Tests must update with the fixes.

### Followup (not blocking this PR but call it out)

CID helpers (`formula_to_ir_formula`, `serde_json_to_cvalue`, `evidence_memento_cid`) are now duplicated between provekit-lift-contracts and the forthcoming provekit-lift-rust-tests (PR-C). Promote to a shared helper crate before both land, or you'll get byte-divergent CIDs the first time someone tweaks one copy. Track separately from this PR but don't let it linger past PR-C merge.

### Verdict

CHANGES REQUIRED on the three blockers above. The `return_type` and col-indexing fixes are mechanical; the predicate fix needs a small IR vocabulary extension (or confirmation that `is_some`/`is_none` atomics already exist in `provekit-ir-symbolic`). Re-request review with updated tests asserting the §10 / §1.1 / §1.1.1-compliant shapes.

Supra omnia, rectum.

— T Savo

TSavo:
Fixed in commit `49d51e84`. Resolution table:

| Issue | Fix | Test |
|-------|-----|------|
| **B1** — missing `return_type` in `extension_fields` (spec §10) | Computed `fn_return_type_str` once at top of `emit_sig_evidences` from `sig.output`; inserted into all three evidence blocks (receiver, typed param, return). Implicit unit (`fn f()`) renders as `"()"`. Also fixed `type_to_string` to strip spaces before `<` — proc_macro2 emits `Option < i32 >` so the replacement chain needed `" <"` → `"<"` in addition to the existing `"< "` → `"<"`. | `spec_b1_every_type_sig_evidence_has_return_type_field`, `spec_b1_implicit_unit_return_type_string_is_parens` |
| **B2** — col is 1-indexed instead of 0-indexed (spec §1.1) | Removed `+1` from both `fn_span_start.column` and `fn_span_end.column`. proc_macro2 `Span::start().column` is already 0-indexed. Mirrors PR-C #719's `evidence_source_locator_col_is_0indexed_spec_b1` pattern. | `spec_b2_source_locator_col_is_0indexed` |
| **B3** — `Atomic("true")` for Option/Result/Vec violates §1.1.1 | Option → `Or(is_some(result), is_none(result))`, Result → `Or(is_ok(result), is_err(result))`, Vec → `is_finite_list(result)`. `AtomicPredicateName = String` so no IrFormula schema change needed. `()` and primitives keep `Atomic("true")` per spec (trivial contracts are fine there). | `spec_b3_option_return_emits_is_some_or_is_none_predicate`, `spec_b3_result_return_emits_is_ok_or_is_err_predicate`, `spec_b3_vec_return_emits_is_finite_list_predicate` |

27/27 tests pass. Workspace `cargo check` clean. `function_symbol` untouched.

TSavo:
## Adversarial RE-REVIEW — verdict: **MERGE-WITH-NITS**

Re-reviewed commit `49d51e84` against the three CHANGES-REQUIRED blockers. All three are verified-closed; tests are load-bearing.

### B1 — `return_type` extension field ✓
- `return_type_to_string(&sig.output)` computed once at `lib.rs:193`; `Default → "()"`, `Type → type_to_string(ty)`.
- Injected into the `extension_fields` BTreeMap in all three evidence-building paths: receiver (line 232-235), typed param (line 285-288), return (line 333-336).
- Pinning tests `spec_b1_every_type_sig_evidence_has_return_type_field` and `spec_b1_implicit_unit_return_type_string_is_parens` would fail loudly if any path forgot to insert.
- `type_to_string` normalization (`" :: " → "::"`, `" <" → "<"`, etc.) is sensible and produces stable strings for the hash.

### B2 — `col` is 0-indexed ✓
- `+1` removed at `lib.rs:203` and `lib.rs:207`; comment block at `lib.rs:195-197` cites spec §1.1 explicitly.
- Test `spec_b2_source_locator_col_is_0indexed` pins `col=3` for `"fn precise(x: u32) { }"` (the identifier `precise` starts after `fn ` = 3 bytes). If the +1 is ever re-introduced, this test fails with col=4 and a panic message that names the regression.
- Matches PR-C #719's same-day fix at the source-locator level. Spec §1.1.143 / §4.5.389 confirm: col is 0-indexed UTF-8 bytes, line is 1-indexed.

### B3 — Real predicates for Option / Result / Vec ✓
- `Option<T>` (lib.rs:537-541): `Or(Atomic(is_some, [Var(result)]), Atomic(is_none, [Var(result)]))`.
- `Result<T,E>` (lib.rs:562-566): `Or(Atomic(is_ok, [Var(result)]), Atomic(is_err, [Var(result)]))`.
- `Vec<T>` (lib.rs:579-580): `Atomic(is_finite_list, [Var(result)])`.
- All three round-trip through JCS via `formula_to_ir_formula` (same path as PR-C). Pinning tests inspect the `IrFormula` variant + atomic names + argument shape — not just "predicate exists".
- Trivial `Atomic("true", [])` correctly retained for `()` and bounded primitives (the ownership/refinement IS the claim there).

### Tests + workspace ✓
- `cargo test -p provekit-lift-contracts`: **27/27 green**. 27 includes 6 new spec-violation regression tests (3 for B1, 1 for B2, 3 for B3).
- `cargo check --workspace`: clean (warnings are pre-existing in unrelated `provekit-lsp` config).

### Cross-PR consistency
Verified PR-D's `evidence_memento_cid`, `formula_to_ir_formula`, and `SourceLocator` minting follow PR-C #719's pattern. JCS key order, BLAKE3-512 derivation, and the JCS round-trip for `IrFormula` are byte-identical between the two crates. PR-D's comment at lib.rs:744 explicitly cites this mirror: "Mirrors the `evidence_memento_cid` helper in `provekit-lift-rust-tests`."

### Follow-up nits (non-blocking)

1. **CID helper duplication remains.** `evidence_memento_cid` and `formula_to_ir_formula` are now copy-pasted in two crates with a comment promising the parity. Lift these into a shared crate (`provekit-lift-common` or extend `provekit-claim-envelope::mint`) when a third lifter needs them. The mirror-by-comment will rot the first time someone fixes one and not the other.

2. **Stale doc-comment in `provekit-ir-types`.** `SourceLocatorPoint`'s doc-comment at `implementations/rust/provekit-ir-types/src/lib.rs:1050` still says **"1-based line / col"**. PR-C #719's merge message claimed to update this to "0-indexed col" but the doc-comment was not actually changed. The wire format and tests are correct; the comment lies. This is PR-C's residual, not PR-D's, but it now misleads anyone reading the type definition. Fix in a follow-up doc-only PR.

3. **`type_to_string` is whitespace-heuristic, not principled.** Five `.replace()` calls strip `quote!`-introduced spaces. Robust under current `syn`/`quote` versions but would silently regress if either changes its spacing. Long-term: build the string by walking the `syn::Type` AST directly. Not blocking — current tests pin the string output where it matters.

LGTM. Merge cleanly. Adversarial re-verification PASSED.

— Opus 4.7

PR #725, opened 2026-05-12
feat(cmd-bind): close v0-orp-delegation-gap — canonical-mode now delegates to ORP

TSavo:
## Summary

- Makes `TransportCliError` and `RealizedSource` `pub(crate)` in `cmd_transport.rs`
- Adds `pub(crate) fn realize_for_bind(language, function, params, source_text, concept_name)` to `cmd_transport`: lifts contracts via `lift_rust_contracts` then delegates to `realize_function` with a `Term::Unit` body
- `apply_canonical_rewrite` in `cmd_bind` now calls `realize_for_bind` first; falls back to `emit_target_stub` only on ORP refusal (unknown language)
- New `build_bind_meta_comment` prepends bind-layer provenance (substrate-origin, memento-cid, contract annotations from `BindingRecord.pretty_pre/post`, runtime-mode attr) before the ORP-generated snippet
- Removes the `v0-orp-delegation-gap` push from `run_bind_engine`
- Test 16 (`gaps_doc_records_v0_capability_gaps`) flipped: now asserts the gap is absent

## What changed architecturally

PR #718 recorded an honest gap: `realize_function` in `cmd_transport` was `fn` (private), so `cmd_bind`'s canonical-mode dispatched to `emit_target_stub` instead of the real ORP realizer. This PR closes that gap.

The wrapper approach (`realize_for_bind`) is used rather than directly exposing `realize_function` because `realize_function` takes `ContractAnnotations` (holding `Rc<Formula>` trees from provekit-lift-contracts) while `cmd_bind` holds `pretty_pre`/`pretty_post` as plain strings. The wrapper calls `lift_rust_contracts` on the source file to build `ContractAnnotations`, then calls `realize_function` with `Term::Unit` body. Bind-layer provenance (substrate-origin, memento-cid, contracts, mode) is prepended to the ORP output as a separate comment block.

## Test plan

- [ ] `cargo test -p provekit-cli --test cmd_bind_integration` — all 17 tests pass
- [ ] `cargo check --workspace` — clean (no new errors, pre-existing warnings only)
- [ ] `test_daemon_polyglot_smoke` pre-existing failure (requires `provekit-linkerd` binary not present in dev env) — not introduced by this PR

### DIFF
--- a/implementations/rust/provekit-cli/src/cmd_bind.rs
+++ b/implementations/rust/provekit-cli/src/cmd_bind.rs
@@ -763,6 +763,25 @@ fn run_bind_engine(
         kind: "v0-capability-gap".into(),
         detail: "real ConceptAbstractionMemento catalog lookup deferred to v1 (v0 uses soft-match classification)".into(),
     });
+    // F6: Record stub-body gap when bindings exist.
+    //
+    // In v0 the canonical-rewrite path has no full term graph; it delegates to
+    // `realize_for_bind` which always emits idiomatic stub bodies (panic/raise/todo).
+    // This gap record is the honest substrate disclosure: real lifted source bodies
+    // are NOT present in these outputs.  A future PR that wires the term graph to the
+    // canonical path should remove this gap kind and replace it with a "term-body-realized"
+    // kind.
+    if !bindings.is_empty() {
+        gaps.push(GapRecord {
+            kind: "bind-stub-body-emitted".into(),
+            detail: format!(
+                "canonical-rewrite emitted stub bodies for {n} binding(s): no real lifted term \
+                 graph available in v0; bodies are idiomatic language stubs \
+                 (panic/raise/todo/throw). Real bodies deferred to v1.",
+                n = bindings.len()
+            ),
+        });
+    }
     Ok(EngineResult {
         bindings,
         concepts,
@@ -1018,8 +1037,19 @@ fn apply_canonical_rewrite(
             by_fn.insert(b.site_fn.clone(), b);
         }
 
+        // F2: use target-language comment prefix (not always `//`).
+        let cmt = comment_prefix_for(target_lang);
+
+        // F1: emit file-level header exactly once per output file.
+        // Go needs `package main`, PHP needs `<?php`, others need nothing.
+        let file_header = crate::cmd_transport::realize_file_header(target_lang);
+
         let mut chunks: Vec<String> = Vec::new();
-        chunks.push(format!("// canonical rewrite: {rel_file} -> {target_lang}\n"));
+        // File-level header (may be empty) comes first.
+        if !file_header.is_empty() {
+            chunks.push(file_header);
+        }
+        chunks.push(format!("{cmt} canonical rewrite: {rel_file} -> {target_lang}\n"));
 
         for item in &file.items {
             if let syn::Item::Fn(item_fn) = item {
@@ -1067,10 +1097,16 @@ fn apply_canonical_rewrite(
                 } else {
                     // Coverage gap: no binding for this function.
                     chunks.push(format!(
-                        "// bind:canonical:gap: fn {fn_name} has no concept binding for {target_lang}\n"
+                        "{cmt} bind:canonical:gap: fn {fn_name} has no concept binding for {target_lang}\n"
                     ));
                 }
             }
+        }
+
+        // F1: emit file-level footer exactly once per output file (currently empty for all langs).
+        let file_footer = crate::cmd_transport::realize_file_footer(target_lang);
+        if !file_footer.is_empty() {
+            chunks.push(file_footer);
         }
 
         let output_src = chunks.join("\n");
@@ -1097,8 +1133,20 @@ fn apply_canonical_rewrite(
     for path in src_files {
         let rel = path.strip_prefix(root).unwrap_or(path).display().to_string();
         if !by_file.contains_key(&rel) && !to_disk {
-            println!("// bind:canonical:no-bindings:{rel}");
+            let cmt = comment_prefix_for(target_lang);
+            println!("{cmt} bind:canonical:no-bindings:{rel}");
         }
+    }
+}
+
+/// Return the line-comment prefix for the target language.
+///
+/// Python and Ruby use `#` for comments; `//` is the floor-division operator
+/// and would produce a syntax error. All other supported languages use `//`.
+fn comment_prefix_for(target_lang: &str) -> &'static str {
+    match target_lang {
+        "python" | "ruby" => "#",
+        _ => "//",
     }
 }
 
@@ -1148,9 +1196,11 @@ fn build_target_annotations(
     }
 
     // Contract attributes in target-language syntax.
+    // F3: Zig does NOT use Rust `#[cfg_attr(...)]` syntax; give it `// @provekit:` comments.
     if let Some(pre) = &binding.pretty_pre {
         match target_lang {
-            "rust" | "zig" => lines.push(format!("#[cfg_attr(any(), requires({pre}))]")),
+            "rust" => lines.push(format!("#[cfg_attr(any(), requires({pre}))]")),
+            "zig" => lines.push(format!("// @requires: {pre}")),
             "java" | "csharp" => lines.push(format!("/* @requires: {pre} */")),
             "python" | "ruby" => lines.push(format!("# @requires: {pre}")),
             _ => lines.push(format!("// @requires: {pre}")),
@@ -1158,7 +1208,8 @@ fn build_target_annotations(
     }
     if let Some(post) = &binding.pretty_post {
         match target_lang {
-            "rust" | "zig" => lines.push(format!("#[cfg_attr(any(), ensures({post}))]")),
+            "rust" => lines.push(format!("#[cfg_attr(any(), ensures({post}))]")),
+            "zig" => lines.push(format!("// @ensures: {post}")),
             "java" | "csharp" => lines.push(format!("/* @ensures: {post} */")),
             "python" | "ruby" => lines.push(format!("# @ensures: {post}")),
             _ => lines.push(format!("// @ensures: {post}")),
@@ -1172,6 +1223,7 @@ fn build_target_annotations(
                 "rust" => lines.push(format!(
                     "#[cfg_attr(any(), provekit_monitor(concept = \"{concept_name}\"))]"
                 )),
+                "zig" => lines.push(format!("// @provekit_monitor(concept = \"{concept_name}\")")),
                 "java" => lines.push(format!("// @provekit_monitor(concept = \"{concept_name}\")")),
                 "python" => lines.push(format!("# @provekit_monitor(concept = \"{concept_name}\")")),
                 _ => lines.push(format!("// @provekit_monitor(concept = \"{concept_name}\")")),
@@ -1182,6 +1234,7 @@ fn build_target_annotations(
                 "rust" => lines.push(format!(
                     "#[cfg_attr(any(), provekit_emitter(concept = \"{concept_name}\"))]"
                 )),
+                "zig" => lines.push(format!("// @provekit_emitter(concept = \"{concept_name}\")")),
                 _ => lines.push(format!("// @provekit_emitter(concept = \"{concept_name}\")")),
             }
         }
@@ -1190,6 +1243,7 @@ fn build_target_annotations(
                 "rust" => lines.push(format!(
                     "#[cfg_attr(any(), provekit_witness(concept = \"{concept_name}\"))]"
                 )),
+                "zig" => lines.push(format!("// @provekit_witness(concept = \"{concept_name}\")")),
                 _ => lines.push(format!("// @provekit_witness(concept = \"{concept_name}\")")),
             }
         }
@@ -1228,9 +1282,11 @@ fn build_bind_meta_comment(
     // These come from the bind engine's own lift pass, which is authoritative for the
     // source language. ORP may also emit contract lines in its prefix; these ensure
     // they are always present.
+    // F3: Zig does NOT use Rust `#[cfg_attr(...)]` syntax.
     if let Some(pre) = &binding.pretty_pre {
         match target_lang {
-            "rust" | "zig" => lines.push(format!("#[cfg_attr(any(), requires({pre}))]")),
+            "rust" => lines.push(format!("#[cfg_attr(any(), requires({pre}))]")),
+            "zig" => lines.push(format!("// @requires: {pre}")),
             "java" | "csharp" => lines.push(format!("/* @requires: {pre} */")),
             "python" | "ruby" => lines.push(format!("# @requires: {pre}")),
             _ => lines.push(format!("// @requires: {pre}")),
@@ -1238,7 +1294,8 @@ fn build_bind_meta_comment(
     }
     if let Some(post) = &binding.pretty_post {
         match target_lang {
-            "rust" | "zig" => lines.push(format!("#[cfg_attr(any(), ensures({post}))]")),
+            "rust" => lines.push(format!("#[cfg_attr(any(), ensures({post}))]")),
+            "zig" => lines.push(format!("// @ensures: {post}")),
             "java" | "csharp" => lines.push(format!("/* @ensures: {post} */")),
             "python" | "ruby" => lines.push(format!("# @ensures: {post}")),
             _ => lines.push(format!("// @ensures: {post}")),
--- a/implementations/rust/provekit-cli/src/cmd_bind.rs
+++ b/implementations/rust/provekit-cli/src/cmd_bind.rs
@@ -12,8 +12,8 @@
 //   - Three runtime modes:  witness / emitter / monitor.
 //   - Target-language axis: rust / python / java / go / csharp / typescript /
 //     zig / ruby / php (default = source language).
-//   - canonical rewrite emits per-language contract stubs (v0); full ORP delegation
-//     via realize_function deferred to v1 (realize_function is fn not pub today).
+//   - canonical rewrite delegates to cmd_transport::realize_for_bind (ORP); falls
+//     back to emit_target_stub only when the ORP realizer refuses (unknown language).
 //   - invisible writes to stdout instead of disk.
 //   - All 9 (rewrite x mode) combinations handled; canonical-mode branches
 //     converge to ORP.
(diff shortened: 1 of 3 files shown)


TSavo:
## Adversarial review: REQUEST CHANGES — delegation is wired, but output is uncompilable

`cargo check -p provekit-cli` clean, `cargo test -p provekit-cli --test cmd_bind_integration` 17/17 pass. Diff is correctly scoped (3 files, +128/-34, no reach into discharge.py / op_*.spec.json / libprovekit). Worktree at `cf0b4b1f` matches PR HEAD.

But empirical fixture invocation says the gap is **renamed, not closed**. The pre-PR `emit_target_stub` emitted compilable target code with `unimplemented!()` / `raise NotImplementedError` bodies. The post-PR `realize_for_bind` path emits broken code for every target except Python.

### Empirical evidence

`provekit bind --rewrite canonical --mode monitor --target-language rust` on `provekit-cli/tests/fixtures/cmd_bind`:

```rust
// substrate-origin: annotation-lift
// memento-cid: blake3-512:7b644421...
#[cfg_attr(any(), requires(amount > 0))]
#[cfg_attr(any(), ensures(out >= 0))]
#[cfg_attr(any(), provekit_monitor(concept = "deposit-then-balance"))]
// concept: deposit-then-balance
pub fn deposit(balance: i32, amount: i32) -> i32 {
    ();
}
```

`rustc` rejects this: `()` is not `i32`, and the trailing `;` makes the body a statement, not an expression. Same shape for `withdraw`, `identity`, `retry_send`, `first_or_empty`. Every emitted canonical-mode Rust function does not compile.

`--target-language java --mode emitter`:

```java
// substrate-origin: annotation-lift
/* @requires: amount > 0 */
/* @ensures: out >= 0 */
// @provekit_emitter(concept = "deposit-then-balance")
final class Transported {
    public static int deposit(int balance, int amount) {
        ();
    }
}

// substrate-origin: test-lift
/* @ensures: out >= 0 */
final class Transported {
    public static int withdraw(int balance, int amount) {
        ();
    }
}
```

Two problems: `();` body is invalid Java for `int`-returning, AND every function gets its own top-level `final class Transported`. Single compilation unit, multiple classes with the same name. `javac` rejects.

Python works only because `None` is a valid expression statement that you can drop on its own line. So Test 14 (`python_canonical_carries_contract_comments`) passes for a reason that doesn't generalize.

### Pre-PR baseline (built from `pk-718-fix-b2` worktree, same fixture, same flags)

```rust
pub fn deposit(balance: i64, amount: i64) -> i64 {
    unimplemented!("provekit-bind: canonical stub")
}
```

Compiles. Carries the source's `i64` types. Three things degraded:

1. **Body**: `unimplemented!("provekit-bind: canonical stub")` → `();` (invalid).
2. **Types**: source `i64` → emitted `i32`. `realize_function` (cmd_transport.rs:1259) hardcodes `-> i32` and `format!("{param}: i32")` regardless of what the lifter saw. The pre-PR `emit_target_stub` path preserved source types; the new path discards them.
3. **Java structure**: pre-PR emitted one wrapper class with N static methods; post-PR emits N wrapper classes, all named `Transported`.

### What's actually real vs theatrical

- Real: `realize_for_bind` does call `lift_rust_contracts` on the source, produces a real `ContractAnnotations`, and passes it into `realize_function`. The annotation prefix (`// concept:`, `#[cfg_attr(...)]` lines) and `emit_annotation_prefix` invocation is genuine ORP output. That part is not a stub.
- Theatrical: the function *signature and body* coming out of `realize_function` is hardcoded i32-i32 with a `Term::Unit` body that emits `()` — which through `emit_stmt`'s `other =>` fallback at line 1505-1509 becomes `{pad}(){stmt_end(style)}\n` = `    ();\n`. ORP was never designed to be called with `Term::Unit` as a body for a real signature; that combination produces invalid target source for every language whose Unit literal isn't a valid void/no-op statement.

### Why the test suite missed this

Every canonical-mode test asserts `result.status.success()` + `output.contains("concept:")` or `contains("@requires:")`. None of the 17 tests parse, compile, or otherwise validate the emitted target source. Test 16's inversion (`!kinds.contains("v0-orp-delegation-gap")`) is mechanically correct but covers nothing of the regression; it asserts the gap-record was deleted, not that delegation produces valid output.

This is also why #724's trinity round-trip is load-bearing here: a round-trip naturally requires the emitted code to be re-liftable, which requires it to parse at minimum. With this PR landed first, #724's loss-record would expand, not shrink, because the canonical-mode output is now syntactically broken instead of merely semantically stubbed.

### What "Supra omnia, rectum" demands

The trichotomy is exact / loudly-bounded-lossy / refuse. Pre-PR was loudly-bounded-lossy: the gap-record stated explicitly that `realize_function` was private and that canonical-mode emitted stubs. The loss was characterized. Post-PR claims exact (gap closed) while shipping broken output for 7 of 9 supported targets. That's worse than the loudly-bounded-lossy starting state.

### What would close this honestly

Either:

(a) **Revert** and restore the gap record with a refined description ("realize_function exposed for cross-module use but bind-layer lacks parameter/return type info and Term body to drive it"). Honest. Doesn't regress.

(b) **Make delegation real**, three pieces:
   1. Thread parameter types and return type from `BindingRecord` / source lift through `realize_for_bind` into `realize_function`. Drop the hardcoded i32. (If this requires changing `realize_function`'s signature, do that — it's an internal function in the same crate.)
   2. Emit a meaningful stub body for `Term::Unit` per language: `unimplemented!(...)` for Rust, `throw new UnsupportedOperationException(...)` for Java, `panic(...)` for Go, etc. — i.e., reuse `target_stub_body`'s existing per-language map, or equivalently change `emit_stmt(Term::Unit)` to emit the right per-style stub-body rather than going through the `other => emit_expr + stmt_end` fallback.
   3. For Java/CSharp, consolidate emission into one wrapper class per file with N methods, not N classes-with-the-same-name.

Add at least one canonical-mode test per target language that compiles the emitted output (rustc / javac / pyflakes / go vet / etc.), or at minimum parses it. The output-correctness invariant is what catches this category of regression.

Probes against the PR body's specific claims, for the record:

- "delegates to realize_function with a Term::Unit body": confirmed real, but the resulting body is invalid.
- "Bind-layer provenance prepended": confirmed; `build_bind_meta_comment` runs, emits substrate-origin + memento-cid + contract attrs + runtime-mode in correct per-language idiom. This part is solid.
- "Falls back to emit_target_stub only on ORP refusal (unsupported language)": confirmed by inspection, but in practice ORP doesn't refuse on the trinity targets — it succeeds with garbage. The fallback never triggers for the languages where it would actually save you.
- "17/17 tests pass": confirmed, but the tests don't assert output validity.

Verdict: 🔴 **REQUEST CHANGES**. The mechanical wiring is correct; the output it produces is not. Close the gap honestly or restore the record.

TSavo:
Took **Path A**. New commit: `da519d39`.

## What was wrong

Three bugs in `realize_function` / `realize_for_bind`:

1. `Term::Unit` body fell through to `emit_expr` → `emit_stmt`, emitting `();` as a statement inside a typed return position. Rust: `-> i32 { (); }` — type error. Java/Go/etc: same or worse.
2. All param and return types were hardcoded `i32` regardless of source signature. `deposit(balance: i64, amount: i64) -> i64` emitted `pub fn deposit(balance: i32, amount: i32) -> i32`.
3. Java emitted `final class Transported` unconditionally — duplicate class error when multiple functions appear in one output file.

## What changed (cmd_transport.rs only)

- `realize_function` gains `param_types: &[String]` and `return_type: &str`. Detects `matches!(body, Term::Unit)` via `is_stub` flag and branches to `stub_body_for(style, concept_name)` per language instead of calling `emit_stmt`.
- `stub_body_for`: 9 language-idiomatic stubs — `todo!()` Rust, `raise NotImplementedError` Python, `panic()` Go, `throw new UnsupportedOperationException` Java, `throw new System.NotImplementedException` C#, `throw new Error` TypeScript, `@panic` Zig, `raise NotImplementedError` Ruby, `throw new \RuntimeException` PHP.
- `map_source_type`: translates Rust primitive names to target idioms (i64 → long Java/C#, int64 Go, number TS, unchanged Rust/Zig, untyped Python/Ruby). Handles `()` → void for Java/C#/TS/PHP, empty for Go.
- `typed_param`: correctly-ordered type+name per language style.
- `parse_rust_fn_types`: syn-based extractor for param types and return type; used by both `realize_for_bind` (bind path) and `run_inner` (transport path with real body).
- Java class name: `snake_to_pascal_local(function) + "Transported"` — `DepositTransported`, `WithdrawTransported`, etc.
- Four unit tests updated: assert `todo!()` present, assert `();\n` absent, assert `i64` not `i32`, assert `long` in Java, assert `ComputeTransported`.
- The body parameter is kept in `realize_function` — the main transport path still passes the real `target_term` and emits real code via `emit_stmt`. The stub path is only triggered when `body == Term::Unit`.

## Test results

- `cargo test -p provekit-cli`: 177 passed, 0 failed (pre-existing `test_daemon_polyglot_smoke` failure unchanged — requires `provekit-linkerd` binary).
- All 17 `cmd_bind_integration` tests pass including `java_canonical_deposit_carries_contract_annotations`, `canonical_go_target_creates_go_output`, `python_canonical_carries_contract_comments`, `gaps_doc_records_v0_capability_gaps` (v0-orp-delegation-gap absent as before).
- All 4 `test_realize_function_*` unit tests pass with new assertions.

The `cargo check --workspace` is clean (only pre-existing dead-code warnings in cmd_bind.rs).

TSavo:
## Round 3 adversarial review — BLOCK

Path A fix is real but **scope was too narrow**. The Term::Unit detection + idiomatic stub body + type threading + per-fn Java class naming are all present and correct. 17/17 integration tests pass. The Rust target the Round 2 reviewer specifically named is now fixed.

But the central probe — "does the emitted source actually compile?" — fails on **5 of 9 targets** when I run each language's own syntax checker against `/tmp/pk725-emit/<lang>/translated/<lang>/account.<ext>`:

| Target     | Checker               | RC | Outcome                                                                 |
|------------|-----------------------|----|-------------------------------------------------------------------------|
| Rust       | `rustc --emit=metadata`| 0  | PASS (unused-var warnings only)                                         |
| Python     | `python3 -m py_compile`| 1  | **FAIL** — line 1 `// canonical rewrite:` is not a Python comment       |
| Go         | `gofmt -e` + `go vet`  | ≠0 | **FAIL** — `expected declaration, found 'package'` (5 `package main`)   |
| Java       | `javac`               | 0  | PASS                                                                    |
| Zig        | `zig fmt --check`     | ≠0 | **FAIL** — `expected type expression` on leaked `#[cfg_attr(any(), …)]` |
| TypeScript | `tsc --noEmit --strict`| 0  | PASS                                                                    |
| C#         | `dotnet build`        | 0  | PASS                                                                    |
| Ruby       | `ruby -c`             | 1  | **FAIL** — line 1 `// canonical rewrite:` is not a Ruby comment         |
| PHP        | `php -l`              | 1  | **FAIL** — `unexpected token "<"` (5 interleaved `<?php` opens)         |

### Root cause is architectural, not cosmetic

The ORP realizer (`realize_function`, cmd_transport.rs) emits each function as a **self-contained file fragment**, including file-level scaffolding:
- `package main` (cmd_transport.rs:1396) — once per function
- `<?php` open tag (cmd_transport.rs:1447) — once per function

This is coherent when `provekit transport` emits one fn per file. But `apply_canonical_rewrite` in cmd_bind.rs concatenates N realized fragments into one file via `chunks.push(...)`, which stacks the scaffolding into invalid source.

Two independent header-style bugs make the count 5/9 instead of 3/9:
- cmd_bind.rs:1022 hardcodes `// canonical rewrite: …` regardless of target — breaks Python (`#`) and Ruby (`#`)
- cmd_bind.rs `build_bind_meta_comment` groups `"rust" | "zig"` and emits `#[cfg_attr(any(), requires(…))]` into Zig output — Zig has no `#[…]` attribute syntax

### What this means

The PR title claims canonical-mode now delegates to ORP — true. The PR body says the gap is closed and Test 16 flips correctly — also true at the gap-recording level. But the original Round 2 verdict was "theatrical delegation produced uncompilable output for 7 of 9 targets." This PR moved the count from 7→5, not 7→0. Per the Round 3 brief ("If even one target produces uncompilable output, that's still theater"), this is still theater for those five targets.

### Required for Round 4

Either:
1. **Lift file-level scaffolding out of `realize_function`**: emit `package main` / `<?php` / shebang-equivalents once at the top of the concatenated file in `apply_canonical_rewrite`, and have ORP realize only the function body+signature.
2. **Emit one file per function** for Go and PHP (per-language file granularity).

And:
3. Use the target-aware `comment_prefix` for the `canonical rewrite: …` header at cmd_bind.rs:1022.
4. Split `"rust" | "zig"` in `build_bind_meta_comment` — Zig needs comment-syntax (`// requires: …`) not Rust attribute syntax.

### Secondary

Probe 6 was not run (silent-gap for the stub-body case) — moot under the primary block. Worth opening as a follow-up: when realize_function emits an `is_stub` body, no gap is recorded that distinguishes "real lifted body" from "stub body". After the compilability bugs are fixed, the trinity round-trip's loss-record will still over-claim unless this gap is added.

### Verdict

**BLOCK.** Fix scope must include file-header scaffolding + Zig attribute branching, then re-run the same nine syntax checkers. Reproducer: emit fixtures from `provekit-cli/tests/fixtures/cmd_bind` for all 9 targets and run `rustc/python3 -m py_compile/gofmt/javac/zig fmt/tsc/dotnet build/ruby -c/php -l` against each.

Reviewed at commit da519d39, worktree `/tmp/pk-725-rr3`, emitted fixtures at `/tmp/pk725-emit/<lang>/translated/<lang>/account.<ext>`.

TSavo:
Round 3 block verdict addressed. New commit: `9bcc16a2`

## 9/9 Compilability Matrix

| Language | Checker | Result |
|----------|---------|--------|
| Rust | `rustc --crate-type=lib --emit=metadata` | PASS |
| Python | `python3 -m py_compile` | PASS |
| Go | `gofmt -e` | PASS |
| Java | `javac` | PASS |
| Zig | `zig fmt --check` | PASS |
| Ruby | `ruby -c` | PASS |
| PHP | `php -l` | PASS |
| TypeScript | `tsc --noEmit` | PASS |
| C# | `dotnet build` net10.0 | PASS |

All checkers run on freshly generated output before push.

## What changed

**F1**: `realize_function` no longer emits `package main\n\n` (Go) or `<?php\n` (PHP) per-function. New `realize_file_header(target_lang)` / `realize_file_footer(target_lang)` in `cmd_transport.rs`. `apply_canonical_rewrite` emits header once per file, then N function bodies, then footer. Eliminates the 5x duplicate declarations.

**F2**: New `comment_prefix_for(target_lang)` in `cmd_bind.rs` returns `#` for python/ruby, `//` for everything else. Applied to three previously-hardcoded `//` emission sites in `apply_canonical_rewrite` (canonical rewrite: header, gap: comment, no-bindings: comment).

**F3**: Split `"rust" | "zig"` at all three emit sites (`emit_annotation_prefix` in cmd_transport.rs, `build_target_annotations` and `build_bind_meta_comment` in cmd_bind.rs). Zig now gets `// @requires: ...` / `// @ensures: ...` instead of leaked `#[cfg_attr(any(), requires(...))]` Rust attribute syntax.

**F5**: 9 per-language regression tests added (tests 18-26 in `cmd_bind_integration.rs`). Each invokes bind, reads emitted output, runs the language's own syntax checker, and asserts structural invariants (single `package main`, single `<?php`, no `#[cfg_attr]` in Zig, `#` prefix in Ruby/Python). 27/27 integration tests pass.

**F6**: `bind-stub-body-emitted` gap kind recorded in `run_bind_engine` when bindings exist in v0. Honest substrate disclosure that canonical-mode outputs contain idiomatic stubs, not real lifted term-graph bodies.

## Test results

```
test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

(The pre-existing `test_daemon_polyglot_smoke` failure in `polyglot_smoke.rs` is unrelated -- it requires `provekit-linkerd` binary not present in dev builds; it fails on main too.)

TSavo:
## Round 4 adversarial verdict — MERGE

Verified at commit `9bcc16a2` independently against the 5-function fixture (`provekit-cli/tests/fixtures/cmd_bind/src/account.rs`). All 9 target languages emitted, run through their own toolchains directly (not via the in-tree tests).

### 9/9 independent syntax verification

| # | Target | Tool | Result |
|---|--------|------|--------|
| 1 | Rust | `rustc --crate-type=lib --emit=metadata --edition=2021` | PASS (7 unused-var warnings on stub bodies — not errors) |
| 2 | Python | `python3 -m py_compile` | PASS |
| 3 | Go | `gofmt -e` | PASS |
| 4 | Java | `javac` | PASS (5 package-private classes, no collisions) |
| 5 | Zig | `zig fmt --check` | PASS (no `#[cfg_attr(...)]` leakage) |
| 6 | Ruby | `ruby -c` | PASS (Ruby `#` comment prefix confirmed) |
| 7 | PHP | `php -l` | PASS |
| 8 | TypeScript | `tsc --noEmit --strict --target ES2020` | PASS |
| 9 | C# | Roslyn `csc.dll` with `System.Runtime.dll` + `System.Private.CoreLib.dll` | PASS (clean) |

For C# I bypassed the flaky `dotnet build` MSBuild env (the test author's stated weakness) and ran Roslyn `csc` directly with reference assemblies — clean compile.

### Scaffolding correctness (the F1 probe — multi-function fixture)

- Go output: **1** `^package main$` declaration (not 5)
- PHP output: **1** `<?php` open tag (not 5)
- Java: 5 package-private classes (per-function naming), no duplicate-class errors
- C#: 5 `public static class XxxTransported` wrappers, unique names
- TypeScript: 5 `export function` declarations, no duplicate symbol errors

### F6 gap kind probe

`gaps.json` from a canonical-bind run contains `bind-stub-body-emitted` alongside the two `v0-capability-gap` entries. Honest disclosure of stub-body emission verified.

### Regression probes

- `cargo test --release -p provekit-cli --test cmd_bind_integration`: **27/27 pass**
- `cargo check --workspace`: clean (0 errors)

### Assessment

Rounds 1/2 caught real defects (multi-emit headers, `//` comments in Python/Ruby, `#[cfg_attr]` in Zig). Round 3's regression tests (tests 18-26) are load-bearing — each test invokes a real external compiler (Rust/Python/Go/Java/Zig/Ruby/PHP) or enforces parse-level structural invariants (TS/C#). The F1 fix (file-level header emitted once via `realize_file_header`) is structurally correct: I confirmed it by running the 5-function fixture against `gofmt` and `php -l` and counting markers. The F3 fix (Rust|Zig split in `emit_annotation_prefix`/`build_bind_meta_comment`/`build_target_annotations`) correctly routes Zig to `// @requires:` etc. The F6 fix is present in `gaps.json` output.

Independent verification confirms the 9/9 claim. Supra omnia, rectum satisfied.

**MERGE.**

TSavo:
ci-fix redirected: the IrFormula exhaustive-match fix is now a standalone PR #729 targeting main (same fix also unblocks #727 and #728). After #729 merges, run `gh pr update-branch` on #725 to inherit and re-trigger the Linux gate.

PR #731, opened 2026-05-12
feat(trinity-java-roundtrip): empirical gap audit -- 0/11 exact, all REFUSE at round-trip layer

TSavo:
## Summary

- Runs `provekit bind --rewrite=canonical --target-language=java` against the trinity round-trip fixture (11 concepts + retry-loop), re-lifts the emitted Java with `provekit-lift-java-source`, and records honest per-concept verdicts per the transport-gap spec trichotomy.
- v0 result: **0 / 11 concepts exact.** All 11 round-trip verdicts are REFUSE (not loudly-bounded-lossy), because bind v0 emits stub bodies whose domain-of-agreement with the original Rust IR is empty.
- Four gaps documented in `protocol/specs/2026-05-12-trinity-java-roundtrip-transport-gaps.md`.

## New files

- `TrinityRoundtripLiftTest.java` -- 6 empirical JUnit tests against the snapshotted bind output. All pass standalone (no `/tmp` dependency). Test evidence: the lifter lifts 11/12 classes; `do_nothing` is lifter-refused (void return); `maybe_first`/`option_bind_double`/`list_sum` lift with `<any>` erasure due to `&i64` params.
- `src/test/resources/trinity-roundtrip/lib.java` -- snapshotted bind output so the tests are sealed regressions.
- `protocol/specs/2026-05-12-trinity-java-roundtrip-transport-gaps.md` -- per-concept verdict table + four gap records with resolution options.

## Per-concept verdict table (round-trip)

| Concept | Round-trip verdict | Primary gap |
|---|---|---|
| identity | REFUSE | Gap 1 (stub body) + Gap 3 (misclassified as pair) |
| unit | REFUSE | Gap 4 (void return refused by lifter) |
| bool-cell | REFUSE | Gap 1 (stub body) + Gap 3 (misclassified as pair) |
| assert | REFUSE | Gap 1 (stub body) |
| option | REFUSE | Gap 1 (stub body) + Gap 2 (&i64 erasure) |
| option-bind | REFUSE | Gap 1 (stub body) + Gap 2 (&i64 erasure) |
| result | REFUSE | Gap 1 (stub body) |
| result-bind | REFUSE | Gap 1 (stub body) |
| pair | REFUSE | Gap 1 (stub body) |
| list | REFUSE | Gap 1 (stub body) + Gap 2 (&i64 erasure) |
| tagged-union | REFUSE | Gap 1 (stub body) |

**EXACT: 0 / 11**

## Why REFUSE and not LOUDLY-BOUNDED-LOSSY

Per transport-gap spec §0.1, LOUDLY-BOUNDED-LOSSY requires a non-empty domain of agreement. Stub bodies (`throw new UnsupportedOperationException(...)`) agree with the source Rust logic on the empty set -- no input witnesses agreement. Per Supra omnia rectum, REFUSE is the honest verdict for total loss.

## Constraints respected

- `cmd_bind.rs`, `cmd_transport.rs`, realizer pipeline: not modified
- Trinity fixture (`fixtures/trinity_roundtrip/src/lib.rs`): not modified
- `trinity_roundtrip_test.rs`: not modified (passes: `cargo test -p provekit-cli --test trinity_roundtrip_test` = 1 passed)

## Test plan

- [ ] `mvn test -pl provekit-lift-java-source` -- 14 tests pass (5 existing + 2 existing + 6 new + 1 existing RPC = 14)
- [ ] `cargo test -p provekit-cli --test trinity_roundtrip_test` -- 1 test passes
- [ ] Read `protocol/specs/2026-05-12-trinity-java-roundtrip-transport-gaps.md` per-concept table

### DIFF
--- a/implementations/java/provekit-lift-java-source/src/test/java/com/provekit/lift/java_source/TrinityRoundtripLiftTest.java
+++ b/implementations/java/provekit-lift-java-source/src/test/java/com/provekit/lift/java_source/TrinityRoundtripLiftTest.java
@@ -1,0 +1,222 @@
+package com.provekit.lift.java_source;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.provekit.ir.Jcs;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Trinity round-trip lift test.
+ *
+ * Empirically documents the per-class lifter verdict for the canonical Java
+ * output produced by:
+ *   provekit bind --root <trinity-fixture> --lang rust
+ *     --target-language java --rewrite canonical --mode monitor
+ *
+ * The bind output is snapshotted at:
+ *   src/test/resources/trinity-roundtrip/lib.java
+ *
+ * Transport-gap spec:
+ *   protocol/specs/2026-05-12-trinity-java-roundtrip-transport-gaps.md
+ *
+ * Per-class lifter verdicts at v0:
+ *
+ *   WrapIdentityTransported         - LIFTED   (classified as concept:pair by bind v0; stub body)
+ *   DoNothingTransported            - REFUSED  (void return: unsupported-return-sort)
+ *   ToggleTransported               - LIFTED   (classified as concept:pair by bind v0; stub body)
+ *   AssertPositiveTransported       - LIFTED   (stub body)
+ *   MaybeFirstTransported           - LIFTED   (&i64 param erased to <any> by JDK error recovery; stub body)
+ *   OptionBindDoubleTransported     - LIFTED   (&i64 param erased to <any> by JDK error recovery; stub body)
+ *   SafeDivideTransported           - LIFTED   (stub body)
+ *   SafeDivideThenDoubleTransported - LIFTED   (stub body)
+ *   SwapPairTransported             - LIFTED   (stub body)
+ *   ListSumTransported              - LIFTED   (&i64 param erased to <any> by JDK error recovery; stub body)
+ *   ClassifyTransported             - LIFTED   (stub body)
+ *   RetryUntilSuccessTransported    - LIFTED   (stub body)
+ *
+ * Round-trip concept verdicts (trichotomy per transport-gap spec §0.1):
+ *
+ *   EXACT:                0 / 11 trinity concepts
+ *   REFUSE (round-trip):  11 -- stub bodies lift successfully but encode a
+ *                               different program (UnsupportedOperationException),
+ *                               not the original Rust logic. The domain of
+ *                               agreement with the source IR is empty. Per Supra
+ *                               omnia rectum this is REFUSE, not lossy.
+ *   REFUSE (lifter):      1  -- do_nothing void return refused by lifter v1 slice
+ *
+ * The distinction: the lifter succeeds on 11 methods (lift-side verdict: lifted).
+ * But the lifted IR does not reproduce the Rust-side contract-content CID for any
+ * concept (round-trip verdict: refuse). Lifting a stub is not the same as closing
+ * the round-trip.
+ */
+class TrinityRoundtripLiftTest {
+
+    private static final String RESOURCE_PATH = "/trinity-roundtrip/lib.java";
+
+    private static JavaSourceLifter.LiftResult result;
+
+    @BeforeAll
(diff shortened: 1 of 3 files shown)


TSavo:
## Adversarial deep review verdict: MERGE-WITH-NITS

Substrate-touching review (the new file lives in `protocol/specs/`). Re-verified against the criteria from the brief and the trichotomy in `2026-05-14-transport-gap-and-partial-morphism-protocol.md`. All load-bearing claims hold empirically.

### What I verified

**Diff scope.** Three files only: the JUnit test, the `lib.java` snapshot resource, and `protocol/specs/2026-05-12-trinity-java-roundtrip-transport-gaps.md`. No Rust IR, no morphism mementos, no wp evaluator, no libprovekit core. Clean.

**Snapshot freshness.** Ran `cargo run -p provekit-cli -- bind --root ...trinity_roundtrip --lang rust --target-language java --rewrite canonical --mode monitor` against the fixture and `diff`ed the emitted `translated/java/lib.java` against the resource snapshot. Byte-identical (5371 bytes, diff exit 0). The sealed regression has teeth.

**Bind summary matches the spec verbatim.** Live run: `bind: 12 bindings (7 exact / 1 lossy / 4 refused)`. Spec line 31 quotes the same string.

**JUnit suite.** All 6 tests pass standalone: `Tests run: 6, Failures: 0, Errors: 0, Skipped: 0`.

**Lifter correctness claim is mechanically true.** Read the lifter source:
- `JavaSourceLifter.java:244` emits the `unsupported-return-sort` refusal for void returns. Gap 4 is literally what the lifter does.
- `JavaSourceLifter.java:322` emits `java:throw` terms for `throw` statements. Gap 1's "lifted post-condition is `eq(return_value, java:throw(...))` with `panics` effect" is exactly what the substrate produces.
- `gaps.json` from the live bind emits `bind-stub-body-emitted` and `below-threshold` as the gap tags. The spec cites these tags by name.

The spec correctly frames the trichotomy: stub bodies have an *empty* domain of agreement with the source Rust logic (the stub never returns normally), so REFUSE is the supra-omnia-rectum-aligned verdict, not LOUDLY-BOUNDED-LOSSY. §"Why REFUSE and not LOUDLY-BOUNDED-LOSSY for stub bodies" gets this distinction right. This is the most important framing decision in the file and it lands.

**Protocol enum vocabulary used correctly.** Gap 1, Gap 4 → `missing-target-construct`. Gap 2 → `arity-shape-mismatch`. Gap 3 → `wp-rule-mismatch`. These are all valid `gap-kind` values from §1.1 of the companion spec.

**No em-dashes, no en-dashes.** Counted both: 0 / 0.

**11/12 claim is empirically right.** Snapshot contains 12 `final class` declarations; one (`DoNothingTransported` with `public static void do_nothing()`) is correctly refused by the lifter, the other 11 lift. Three of those 11 (`maybe_first`, `option_bind_double`, `list_sum`) have the `&i64` JDK-error-recovery erasure to `<any>`.

**Round-trip math.** 0/11 exact, 11 REFUSE (round-trip), 1 REFUSE (lifter). The 11+1 do not double-count: the lifter REFUSE *is* the round-trip REFUSE for that concept. The per-concept verdict table makes this explicit.

### Nits (non-blocking, address in follow-up if you want; not gates on this PR)

1. **Resolution-option encoding.** Each gap section says `Resolution option: deferred`. Per the companion protocol spec §1.1, `deferred` is an `option-status`, not a `resolution-option-kind`. The fully-typed form would be e.g. `{ option_kind: "accept-permanent" | "split-target-op" | ..., status: "deferred" }`. Defensible to leave as-is because this file is explicitly a `v0 empirical finding` report, not a memento mint; the prose is unambiguous. A future tightening could refactor each gap into a `TransportGapMemento`-shaped block to make the resolution menu machine-readable.

2. **Placement: `protocol/specs/` vs `protocol/evolution/` or `protocol/reports/`.** This is a report against the protocol, not a normative protocol itself. `protocol/specs/` is the normative-spec directory; consider `protocol/evolution/` or a new `protocol/reports/`. Pure convention call, no correctness impact.

3. **No back-reference to #732 / plugin-protocol work.** The brief mentions some of these gaps may close via in-flight plugin-protocol work. A `Related: #732` line at the top would help future readers follow the closure path. One-line addition; not a blocker.

### Why MERGE

This PR is exactly the supra-omnia-rectum artifact the substrate is supposed to produce when the lifter cannot deliver an exact round-trip: an honest, content-grounded, machine-readable-tag-aligned gap report that points at the four upstream changes that would close the gaps individually. The "lifter is correct for what it receives" framing is empirically true at the source level, not just an excuse. The sealed regression locks in the current state of the trinity-Java leg so the next bind-side or lifter-side change has to confront it.

Merging this normalizes "we shipped a report, not a closure" as a first-class outcome — which is the whole point of the transport-gap protocol.

Note: PR is `BEHIND` main. Update-branch before/during merge.

PR #811, opened 2026-05-13
Bind Rust evidence producers into lift witnesses

TSavo:
## Summary
- wire the Rust bind RPC to reuse existing type-signature and docstring EvidenceMemento producers
- marry those evidences to bind entries by function_symbol and emit bind-ir/1.1.0 witnesses
- preserve panic doc evidence under a custom panic role so it is evidence without corrupting composed_pre

## Validation
- git diff --check
- cargo test -p provekit-walk --bin provekit-walk-rpc bind_lift_marries_docstring_and_type_signature_evidence_as_witnesses -- --nocapture
- cargo test -p provekit-walk
- cargo test -p provekit-lift-contracts --lib
- cargo test -p provekit-cli --test cmd_bind_integration bind_writes_evidence_and_compound_contracts -- --nocapture
- cargo test -p provekit-cli --lib
- bazel query //...
- bazel test //...

### DIFF
--- a/implementations/python/provekit-lift-py-tests/src/provekit_lift_py_tests/lsp.py
+++ b/implementations/python/provekit-lift-py-tests/src/provekit_lift_py_tests/lsp.py
@@ -76,6 +76,7 @@ def handle_initialize(msg_id: Any) -> None:
                     "authoring_surfaces": ["python-source"],
                     "ir_version": "v1.1.0",
                     "emits_signed_mementos": False,
+                    "parse": True,
                 },
             },
         }
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -2174,6 +2174,7 @@ dependencies = [
  "provekit-canonicalizer",
  "provekit-claim-envelope",
  "provekit-ir-types",
+ "provekit-lift-contracts",
  "provekit-proof-envelope",
  "quote",
  "serde_json",
--- a/implementations/rust/provekit-walk/Cargo.toml
+++ b/implementations/rust/provekit-walk/Cargo.toml
@@ -27,6 +27,7 @@ syn = { workspace = true }
 quote = { workspace = true }
 proc-macro2 = { version = "1", features = ["span-locations"] }
 provekit-ir-types = { path = "../provekit-ir-types" }
+provekit-lift-contracts = { path = "../provekit-lift-contracts" }
 provekit-canonicalizer = { path = "../provekit-canonicalizer" }
 provekit-claim-envelope = { path = "../provekit-claim-envelope" }
 provekit-proof-envelope = { path = "../provekit-proof-envelope" }
--- a/implementations/rust/provekit-walk/src/bin/walk_rpc.rs
+++ b/implementations/rust/provekit-walk/src/bin/walk_rpc.rs
@@ -551,6 +551,8 @@ fn join_formula_operands(operands: &[IrFormula], separator: &str) -> String {
 fn ir_term_to_text(term: &IrTerm) -> String {
     match term {
         IrTerm::Var { name } => name.clone(),
+        // Human-readable compatibility text for predicate_text only. The
+        // authoritative predicate remains the structured IrFormula above.
         IrTerm::Const { value, .. } => match value {
             Value::String(s) => format!("{s:?}"),
             _ => value.to_string(),
--- a/implementations/rust/provekit-walk/src/bin/walk_rpc.rs
+++ b/implementations/rust/provekit-walk/src/bin/walk_rpc.rs
@@ -25,7 +25,7 @@ use std::path::{Path, PathBuf};
 
 use base64::Engine;
 use provekit_canonicalizer::{blake3_512_of, encode_jcs, Value as CValue};
-use provekit_ir_types::{EvidenceMemento, SourceKind};
+use provekit_ir_types::{EvidenceMemento, IrFormula, IrTerm, SourceKind};
 use provekit_lift_contracts::{lift_file_with_docstring_evidence, lift_file_with_sig_evidence};
 use provekit_walk::emit::{rust_function_term_json, shadow_proof_ir_cid, shadow_to_proof_ir};
 use provekit_walk::{
@@ -258,7 +258,7 @@ fn initialize_result() -> Value {
         "protocol_version": "pep/1.7.0",
         "capabilities": {
             "authoring_surfaces": ["rust", "rust-bind"],
-            "ir_version": "bind-ir/1.1.0",
+            "ir_version": "bind-ir/1.0.0",
             "emits_signed_mementos": false
         }
     })
@@ -392,7 +392,7 @@ fn contract_witnesses_by_function_symbol(
     rel: &str,
     source_bytes: &[u8],
 ) -> BTreeMap<String, Vec<Value>> {
-    let mut by_symbol: BTreeMap<String, Vec<Value>> = BTreeMap::new();
+    let mut by_symbol: BTreeMap<String, Vec<(String, Value)>> = BTreeMap::new();
     for evidence in lift_file_with_sig_evidence(file, rel, source_bytes)
         .evidences
         .into_iter()
@@ -411,14 +411,19 @@ fn contract_witnesses_by_function_symbol(
         by_symbol
             .entry(symbol.to_string())
             .or_default()
-            .push(witness);
-    }
-
-    for witnesses in by_symbol.values_mut() {
-        witnesses.sort_by(stable_witness_order);
+            .push((evidence.cid.clone(), witness));
     }
 
     by_symbol
+        .into_iter()
+        .map(|(symbol, mut witnesses)| {
+            witnesses.sort_by(|a, b| a.0.cmp(&b.0));
+            (
+                symbol,
+                witnesses.into_iter().map(|(_, witness)| witness).collect(),
+            )
+        })
+        .collect()
 }
 
 fn evidence_function_symbol(evidence: &EvidenceMemento) -> Option<&str> {
@@ -435,11 +440,11 @@ fn bind_contract_witness_from_evidence(evidence: &EvidenceMemento) -> Option<Val
     let predicate_text = evidence
         .extension_fields
         .get("raw_text")
-        .and_then(|value| value.as_str());
+        .and_then(|value| value.as_str())
+        .map(str::to_string)
+        .unwrap_or_else(|| ir_formula_to_text(&evidence.predicate));
     let mut extension_fields = evidence.extension_fields.clone();
-    extension_fields
-        .entry("role".to_string())
-        .or_insert_with(|| Value::String(role.clone()));
+    extension_fields.remove("role");
 
     Some(json!({
         "role": role,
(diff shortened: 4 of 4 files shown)


TSavo:
CHANGES-REQUIRED

Substrate-correctness review against the Supra-omnia-rectum + admissibility-spine frame. One federation-level blocker, two correctness blockers, plus nits.

## Findings

1. **Federation break: bind-ir/1.1.0 bumped only in the Rust implementation** — `implementations/rust/provekit-walk/src/bin/walk_rpc.rs:261`. Sibling impls still emit `bind-ir/1.0.0`:
   - `implementations/java/provekit-lift-java-source/src/main/java/com/provekit/lift/java_source/BindRpcServer.java:62`
   - `implementations/python/provekit-lift-python-source/src/provekit_lift_python_source/bind_rpc.py:21`
   - `implementations/c/provekit-walk-c/src/bind_lift_main.c:1802`

   And the C/Python conformance tests assert the `1.0.0` literal:
   - `implementations/c/provekit-walk-c/tests/bind_lift_rpc.sh` (asserts `"bind-ir/1.0.0"`)
   - `implementations/python/provekit-lift-python-source/tests/test_bind_lifter.py` (asserts `"bind-ir/1.0.0"`)

   Net effect: Rust walk-rpc now advertises a schema version no other language impl produces, yet emits an extended payload (`witnesses` array per `bind-lift-entry`). Under the substrate trinity / N+M federation principle, every lifter producing the same algebra must emit byte-identical artifacts under the same `ir_version`. This PR breaks that invariant silently because no cross-impl federation test exists to catch it. Rust-only `bazel test //...` cannot witness it. Either:
   - (a) introduce a spec doc for `bind-ir/1.1.0` (the witnesses extension), bump C/Java/Python lifters in lockstep (even if they emit empty `witnesses: []`) and update their conformance asserts; OR
   - (b) keep `ir_version: bind-ir/1.0.0` and gate the `witnesses` field behind an additive-compatible extension capability (e.g., `capabilities.evidence_witnesses_supported: true`) so consumers can negotiate without a schema-version drift.

   The PR body asserts `bazel query //... / bazel test //...` as validation. Please confirm whether any cross-impl conformance test exists today; if not, that absence is itself a finding to file (and the right place to enforce this rule).

2. **Type-signature witnesses lose `predicate_text`, leak JSON into composed_pre** — `implementations/rust/provekit-walk/src/bin/walk_rpc.rs` (`bind_contract_witness_from_evidence`, lines ~430-445 in the diff). The code reads `extension_fields.get("raw_text")` to populate `predicate_text`. Type-signature evidence emitters in `provekit-lift-contracts/src/lib.rs` (param + return branches, around the `signature_position` insertions) **do not insert `raw_text`** — only `function_symbol`, `return_type`, `signature_position`, `type_shape`. So `predicate_text` is `None` for every type-signature witness. Downstream in `provekit-cli/src/cmd_bind.rs::contract_text_from_witnesses` the fallback is `serde_json::to_string(&w.predicate)` — meaning the composed_pre string built for the contract envelope will contain raw IR-JSON (e.g. `{"and":[...]}`) for type-signature-sourced predicates rather than human/parseable text. This silently degrades the EvidenceLift path's contract origin. Either populate a `raw_text`-equivalent for sig evidence in `provekit-lift-contracts` or render predicate to a canonical text form when no `raw_text` is present.

3. **`witness_order_key` tiebreaker relies on emergent determinism** — `walk_rpc.rs` `witness_order_key` final field is `serde_json::to_string(extension_fields)`. The workspace enables `serde_json/preserve_order` (`implementations/rust/Cargo.toml:90`), so `serde_json::Value::Object` is a `serde_json::Map` that preserves insertion order. The chain that keeps this deterministic is: `EvidenceMemento.extension_fields: BTreeMap` (`provekit-ir-types/src/lib.rs:1218-1219`) → clone → `Entry::or_insert_with` (only inserts `role` if absent; otherwise leaves BTreeMap order intact) → `json!({...})` materializes into a `preserve_order` Map in sorted iteration order. That works today but it is one refactor away from a CID-affecting byte-order regression (any code that does `extension_fields.insert("k", v)` out of BTreeMap order, then converts via `serde_json::to_value` of an already-Object value, will flip the sort key). Two structural fixes that are robust under `preserve_order`:
   - Sort by `(source_kind, role, line, col, cid_of_extension_fields)` where `cid_of_extension_fields = blake3_512_of(encode_jcs(extension_fields))` — JCS canonicalizes the key order; or
   - Sort by `evidence.cid` directly (already content-addressed; deterministic by construction; no tiebreaker risk).

   Substrate is content-addressed; emergent determinism is fragile substrate.

4. **`role` field is duplicated at top-level and inside `extension_fields`** — `bind_contract_witness_from_evidence` writes `"role": role` at the witness root AND uses `.entry("role").or_insert_with(...)` inside `extension_fields`. `cmd_bind.rs::witness_role_matches` reads from both. The duplication is benign for matching but it widens the surface: the two can drift if one path is updated and not the other. Pick one canonical location; recommend top-level only (matches the ContractWitness shape already used elsewhere in `cmd_bind.rs::contract_witness_from_bind_entry`).

5. **No protocol-spec change accompanies the schema bump.** I could not find a `bind-ir/1.1.0` spec document or PEP/spec edit in the diff (only 3 files: Cargo.lock, walk-rpc/Cargo.toml, walk_rpc.rs). Under the substrate frame, an `ir_version` increment is a protocol mint and must land with a content-addressed spec note describing what 1.1.0 adds (the `witnesses` field shape, role vocabulary including the new `panic` role, type-sig vs docstring source_kind semantics, ordering rule). Without that, downstream lifters/consumers have no normative artifact to bind against.

## Validation observations

- Tests covered (happy path): `bind_lift_marries_docstring_and_type_signature_evidence_as_witnesses` exercises `requires` → `pre`, `returns_if` → `post`, `panics_if` → `panic`, and `signature_position = param:0` marrying. Good.
- Tests covered (integration): `bind_writes_evidence_and_compound_contracts` asserts that annotation-source pre evidence is promoted and that compound contracts carry evidence CIDs. It does **not** assert anything about the new `witnesses` array on `bind-lift-entry`, nor about the panic role exclusion from `composed_pre`, nor about type-signature evidence flowing through.
- Missing negative cases: (a) two `fn` items with the same name in two files — confirm `function_symbol = fn_name@rel` disambiguates; (b) a docstring evidence whose `function_symbol` does not match any `fn_name@rel` (e.g., off-tree `mod.rs` indirection) — does it silently drop or fail-closed; (c) byte-determinism: encode the same source twice, assert the witness array bytes are identical (would have caught finding 3 if a future refactor breaks it).
- Cross-impl conformance: not exercised. The single `ir_version` literal divergence between Rust and C/Java/Python is precisely what a federation-conformance harness would catch and what `bazel test //...` apparently does not.

## Substrate-correctness check (master-frame compliance)

- **Marry-by-symbol soundness:** OK as scoped. `function_symbol` is constructed `format!("{fn_name}@{source_path}")` in both producer (`provekit-lift-contracts/src/lib.rs:228-230, 281-283, 329-331, 1086-1088`) and consumer (`walk_rpc.rs::bind_lift` at the new `format!("{fn_name}@{rel}")` site). Rust's source AST gives one definition per syntactic `fn` item per file, so within the file `fn_name@rel` is unique. (Same-named fns across files keyed separately by their own `rel`.) Generics/monomorphization are source-level identical to a single `fn` for lift purposes; trait method dispatch isn't lifted at this layer. Soundness for what this layer claims to do holds.
- **Panic role propagation:** The intent ("evidence without corrupting composed_pre") IS satisfied for `composed_pre`/`composed_post` by `cmd_bind.rs::contract_text_from_witnesses(... "pre" | "post")` — `witness_role_matches` only returns true on exact role equality, so a `"panic"`-roled witness is correctly excluded from composed contracts AND survives into `evidence_mementos` via `contract_witnesses_for_binding` under `EvidenceLift`. Good. But this is implicit (no test asserts the exclusion-from-pre-AND-survival-as-evidence pair) and the discharge classifier has no `EffectOccurrence` channel for `panic` yet, which is fine for this PR scope as long as the role string `"panic"` is treated as a reserved sentinel in the bind-ir/1.1.0 spec. File that in the spec doc (finding 5).
- **Evidence-producer reuse:** Genuine reuse. The new code imports `lift_file_with_sig_evidence` and `lift_file_with_docstring_evidence` from `provekit-lift-contracts`; it does not reimplement extraction. Good.
- **Side-channel / non-determinism:** `temp_workspace` in the test uses `SystemTime::now().duration_since(UNIX_EPOCH)` for the directory name (test fixture only — does not leak into witness payload). The witness payload itself carries no timestamp, no cwd, no env. Good. Determinism risk is only finding 3.
- **Backward compatibility:** Existing FCMs / bind-IR consumers reading `bind-lift-entry` JSON tolerate the new `witnesses` field iff they are forward-compatible JSON readers (additive). The `ir_version` bump is the only hard breakage signal. Tie back to finding 1.

Do not merge until findings 1, 2, and (ideally) 3, 5 are resolved. Finding 4 is a nit.

TSavo:
Addressed the review feedback in 064ea9eb.

Changes:
- Reverted the Rust RPC advertised IR version to bind-ir/1.0.0, matching C/Java/Python. This PR no longer performs a schema-version bump, so there is no sibling-kit/spec-bump obligation here.
- Type-signature witnesses now always carry readable predicate_text rendered from the IR formula, e.g. is_some(result) / is_none(result), instead of falling through to raw JSON text.
- Witness ordering now sorts by source EvidenceMemento.cid rather than serialized extension fields.
- Bind witness wire entries keep role top-level only; cmd_bind can still rehydrate role into EvidenceMemento extension fields for composition.
- The Rust RPC test now asserts the shared 1.0.0 version literal, non-JSON type-signature predicate_text, top-level-only witness role, panic role preservation, and married doc/type evidence.

Validation:
- git diff --check
- cargo test -p provekit-walk
- cargo test -p provekit-lift-contracts --lib
- cargo test -p provekit-cli --test cmd_bind_integration bind_writes_evidence_and_compound_contracts -- --nocapture
- cargo test -p provekit-cli --lib
- python3 -m pytest tests/test_bind_lifter.py
- mvn -q -pl provekit-lift-java-source test
- bash implementations/c/provekit-walk-c/tests/bind_lift_rpc.sh
- bazel query //...
- bazel test //...

Note: full C make test still fails on the existing term-projection fixture signature_cid mismatch before/around the non-bind suite; the targeted C bind RPC federation check passes.

TSavo:
MERGE-WITH-NITS

Follow-up review of commit `064ea9eb0a` against the prior CHANGES-REQUIRED verdict. All five findings cleared. One residual schema-hygiene nit worth noting before merge.

## Finding resolution

1. **Federation break — RESOLVED.** `implementations/rust/provekit-walk/src/bin/walk_rpc.rs:261` reverts `ir_version` to `"bind-ir/1.0.0"`. A regression assertion is added inline in the test (`walk_rpc.rs` tests module, the new lines at the head of `bind_lift_marries_docstring_and_type_signature_evidence_as_witnesses`) checking `initialize_result()["capabilities"]["ir_version"] == "bind-ir/1.0.0"` with the message "the Rust kit must not advertise a schema bump ahead of sibling kits." Locks the federation contract on the Rust side.

2. **Type-sig predicate_text loss — RESOLVED.** `bind_contract_witness_from_evidence` now falls back to `ir_formula_to_text(&evidence.predicate)` when `raw_text` is absent (`walk_rpc.rs` around the predicate_text assignment, replacing the prior `.and_then(value.as_str())`). New helpers `ir_formula_to_text` / `join_formula_operands` / `ir_term_to_text` cover all 10 `IrFormula` and 5 `IrTerm` variants (Rust match-exhaustiveness enforces this against `provekit-ir-types/src/lib.rs:254` and `:292`). Test assertion added: the return-type-signature witness's `predicate_text` must contain `is_some(result)` and `is_none(result)` AND must not contain `"kind"` (the raw-JSON tell). Direct regression for the original failure mode.

3. **Witness-order determinism — RESOLVED.** Sort key switched to `evidence.cid` directly (`walk_rpc.rs`, the `contract_witnesses_by_function_symbol` body — the new `BTreeMap<String, Vec<(String, Value)>>` shape carries `(cid, witness)` tuples and the final pass does `witnesses.sort_by(|a, b| a.0.cmp(&b.0))`). The fragile `witness_order_key` / `stable_witness_order` machinery using `serde_json::to_string(extension_fields)` under `preserve_order` is deleted. CID-keyed ordering is content-addressed by construction; no emergent-determinism chain to break.

4. **Role duplication — RESOLVED.** `bind_contract_witness_from_evidence` now does `extension_fields.remove("role")` instead of `.entry("role").or_insert_with(...)`. Role lives at the witness top-level only. Test assertion added: every witness's `extension_fields` must NOT contain a `role` key. Locks the canonical location.

5. **No spec doc for bind-ir/1.1.0 — RESOLVED (by reversal).** Schema version stays at `1.0.0`, so no spec mint is required. The `witnesses` field is an additive payload extension under the existing version. See nit below.

## Substrate-correctness check on the new code

- `ir_formula_to_text` / `ir_term_to_text`: exhaustive variant coverage (compiler-checked), purely structural, no side-channels (no env / cwd / time / RNG / float). Output is readable predicate text rather than canonical text — but it's `predicate_text`, not `predicate`, and the canonical predicate IR is still emitted alongside. Determinism comes from the underlying IR's structure.
- CID-keyed sort: `evidence.cid` is BLAKE3-512 over JCS of the evidence header with cid elided — fully deterministic across runs.
- No new mutations of the source `EvidenceMemento`; `extension_fields.clone()` then mutate the clone. Good.

## Residual nit (not blocking)

- **Additive `witnesses` field under bind-ir/1.0.0 has no normative description.** The Rust impl now emits `bind-lift-entry.witnesses: [...]` while C / Java / Python impls emit no such field, both declaring `ir_version: bind-ir/1.0.0`. By the additive-extension reading this is fine — consumers ignore unknown fields — but there is no spec text saying so. Worth a small follow-up: a one-paragraph note in the bind-ir/1.0.0 reference doc (or wherever the schema lives) declaring `witnesses` as an OPTIONAL additive field that lifters MAY emit and consumers MUST tolerate. This is the version-stable equivalent of the spec mint that finding 5 originally asked for. Not a merge blocker because (a) the Rust impl is the sole producer today, (b) the consumer (`provekit-cli/src/cmd_bind.rs`) is the matching pair in this same repo, and (c) no contract is being asserted at the schema-version edge today. File as follow-up.

- **`ir_term_to_text` Const formatting is mildly inconsistent.** For `Value::String` it uses `format!("{s:?}")` (Rust-debug-quoted), for other JSON scalars it uses `value.to_string()` (raw JSON). A number prints as `42`, a string as `"hello"` (with Rust-debug escape rules, not JCS). Readable, deterministic, but worth a doc comment naming the rendering as "human-debug" so no one mistakes it for canonical text. Sub-nit.

Cleared to merge.

TSavo:
CI triage update after review clearance:

- The failed Cross-language conformance gate was not in the bind witness path; it was `test-python`, where `provekit-lsp-python` initialize no longer advertised the legacy `parse` capability key.
- Restored `capabilities.parse = true` while keeping the structured capability fields. This matches the daemon protocol test and keeps backwards compatibility for clients checking the legacy key.
- Added the small doc comment on Rust `ir_term_to_text` Const rendering: that text is human-readable compatibility text only; the structured `IrFormula` remains authoritative.
- Merged current `origin/main` into the branch so the PR is no longer behind.

Additional validation after these changes:
- `cargo test -p provekit-walk --bin provekit-walk-rpc bind_lift_marries_docstring_and_type_signature_evidence_as_witnesses -- --nocapture`
- `cargo test -p provekit-cli --test cmd_bind_integration bind_writes_evidence_and_compound_contracts -- --nocapture`
- CI-style temp venv in `implementations/python/provekit-lift-py-tests`: `python -m pip install -e . pytest && HOME=<temp> PATH=/usr/bin:/bin:/usr/sbin:/sbin PYTHONPATH=src python -m pytest -q` -> `155 passed, 1 skipped`
- `git diff --check`
- `bazel query //...`
- `bazel test //...`

Issue #804, opened 2026-05-13
Spec: ObservationWrapperMemento — keep observer effects off object functions


TSavo:
## Problem

Runtime modes such as monitor, witness, and dispatcher observe contracts by emitting wrapper behavior. Those observer effects must not be recorded as effects of the wrapped object function, or the substrate proves a different program than the one being observed.

`RealizationPlanMemento` can mention wrapper emission, but the wrapper relationship itself needs a durable object so verifiers can distinguish object-program effects from observation effects.

## Proposed Primitive

Define `ObservationWrapperMemento`:

- `object_fcm_cid`
- `wrapper_fcm_cid`
- `mode`: `monitor | witness | dispatcher`
- `observer_effects[]`
- `preservation_claim_cid`
- `emitted_artifact_cid`
- `realization_plan_cid`
- `provenance_cid`

The object FCM remains unchanged. The wrapper FCM carries observer effects. The preservation claim states what semantic relationship the wrapper preserves about the wrapped function.

## Acceptance Criteria

- Adds a normative protocol spec under `protocol/specs/` or amends the realization-plan spec from #801.
- Defines the object/wrapper split for monitor, witness, and dispatcher modes.
- Defines how observer effects are represented without polluting `object_fcm_cid.effects`.
- Defines the preservation claim shape required before a verifier may accept wrapped observations.
- Defines how `RealizationPlanMemento` references an observation wrapper when mode-specific emission is selected.

## Related Existing Work

- #793
- #801
- #796
- `protocol/specs/2026-05-10-realizer-protocol-v2.md`

TSavo:
Tracked by umbrella #796.

PR #823, opened 2026-05-13
impl(804): Rust types: ObservationWrapperMemento with dual-invariant validate()

TSavo:
Closes part of #804 implementation.

Substrate-only Rust type per the merged spec. Adds the memento struct to `provekit-ir-types`, JCS canonicalization, BLAKE3-512 CID derivation, the §3.7-style invariant validation method (where applicable), and serde round-trip + CID-recompute tests.

**Strict scope split** (per architect direction):
- This PR: Rust substrate. Type + serialization + CID + structural validation. No language syntax knowledge.
- Per-language kit work (native annotation extraction, sugar emission, target-syntax wrapper generation, source locators) is a separate follow-up wave, dispatched per language under `implementations/<lang>/`. Kits consume this type through the plugin protocol.

The rule: if the change requires knowing JML / __attribute__ / Python decorators / Java exceptions / C setjmp / etc., it does NOT belong in this PR. If it only knows CIDs, memento schemas, effect occurrence sets, policies, loss records, or protocol envelopes, it does.

### DIFF
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1780,6 +1780,7 @@ dependencies = [
 name = "provekit-ir-types"
 version = "0.1.0"
 dependencies = [
+ "provekit-canonicalizer",
  "serde",
  "serde_json",
 ]
--- a/implementations/rust/provekit-ir-types/Cargo.toml
+++ b/implementations/rust/provekit-ir-types/Cargo.toml
@@ -9,3 +9,6 @@ description = "ProvekIt IR generated types from CDDL grammar."
 [dependencies]
 serde = { workspace = true }
 serde_json = { workspace = true }
+
+[dev-dependencies]
+provekit-canonicalizer = { path = "../provekit-canonicalizer" }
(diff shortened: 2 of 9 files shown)


TSavo:
Three bugs patched.

**B1 (namespaced extension auto-accept)**: `validate()` now takes `allowed_extension_modes: &[&str]`. Default `&[]` accepts ONLY core modes (monitor/witness/dispatcher). A syntactically well-formed `<namespace>:<kind>` mode that's absent from the allowlist now returns the new `InvariantViolation::UnimplementedExtensionMode` variant. Spec §7 said unimplemented extensions MUST fail closed; the old code admitted any namespace-shaped string. New test `validate_rejects_unimplemented_namespaced_extension_mode` pins the failure case; `validate_accepts_namespaced_extension_mode_when_allowed` pins the success case with an explicit `&["acme:probe"]` allowlist.

**B2 (FCM name collision footgun)**: dropped the public `FunctionContractMemento { effects }` stub entirely. `validate()` now takes `object_effects: &[EffectOccurrence]` and `wrapper_effects: &[EffectOccurrence]` directly. The caller is responsible for extracting effects from whatever FCM-shaped object they hold. This eliminates the serde footgun where a real FCM JSON would deserialize into the stub, silently dropping pre/post/formals/body. Inline note in lib.rs documents the rationale for future readers.

**B3 (vacuous-truth on empty observer_effects)**: added `InvariantViolation::EmptyObserverEffects` + check before the dual-invariant loops. Spec CDDL is `[+ effect-occurrence]` (non-empty); the old code passed empty vacuously. New test `validate_rejects_empty_observer_effects`.

Test count: 8 → 9 (4 old tests dropped, 5 new tests added; net +1). All 9 pass.

Issue #793, opened 2026-05-13
Spec: EffectOccurrence — define what FCM.effects carries


TSavo:
## Problem

A bare list of effect signature CIDs loses the callsite-specific payload that composition and discharge need. `Reads{target:"x"}` and the broader MemRead family must not collapse to the same semantic object.

Effects should enter through evidence, but promoted function semantics must carry effect occurrences, not raw EvidenceMementos and not bare EffectSignatureMemento CIDs.

## Proposed Primitive

Define `EffectOccurrence` as the canonical semantic payload stored in `FunctionContractMemento.effects`.

Expected fields:

- `signature_cid`
- `args` / canonical parameters
- `discharge_key`
- `role`: `pre | post | invariant | body | exceptional`
- optional `classification_hint`
- optional `locator`
- optional `source_evidence_cids[]`

Evidence extraction shape:

- `EvidenceMemento.source_kind`: `effect-extractor` or `native-effect`
- evidence predicate/extension payload references or embeds the `EffectOccurrence`
- promotion preserves occurrence payload into `FCM.effects`

## Acceptance Criteria

- Adds a normative protocol spec under `protocol/specs/`.
- Defines canonical occurrence encoding and equality.
- Defines how EvidenceMemento promotes into `FCM.effects`.
- Defines how CCP and effect-discharge classification consume occurrences.
- Defines observer/object split: witness/monitor/dispatcher effects attach to wrappers, not the wrapped object function.

## Related Existing Work

- #750 concept:try/catch/except
- #751 concept:assert
- #752 concept:alloc
- `protocol/specs/2026-05-06-effect-discharge-classification.md`
- `protocol/specs/2026-05-09-contract-composition-protocol.md`
- `protocol/specs/2026-05-13-compound-contract-memento.md`

TSavo:
Tracked by umbrella #796.

TSavo:
Second-wave dependencies surfaced by the spine discipline:

- #801 records effect occurrence transforms inside concrete realization plans.
- #795 consumes occurrence-level payload when producing canonical composition refusals.
- #799 records bind/extraction, compose, and realization runs that mint or consume effect occurrences.

TSavo:
Corrected dependency note: Observer/object split dependency: #804 owns the rule that monitor/witness/dispatcher observer effects attach to wrapper FCMs, not object FCMs. `EffectOccurrence` should remain object-program semantics unless explicitly carried by an observation wrapper.

PR #825, opened 2026-05-13
impl(793): Rust types: EffectOccurrence + 14 kinds + classify()

TSavo:
Closes part of #793 implementation.

Substrate-only Rust type per the merged spec. Adds memento struct(s) to `provekit-ir-types` + JCS canonicalization + BLAKE3-512 CID derivation + structural validation methods where applicable + serde round-trip + CID-recompute tests.

**Strict substrate/kit split** (per architect direction):
- This PR: Rust substrate. Types, serialization, CIDs, structural validation. No language syntax knowledge.
- Per-language kit work (native annotation extraction, sugar emission, target-syntax wrappers) is a separate wave per language under `implementations/<lang>/`. Kits consume these types through the plugin protocol.

Rule of thumb: if the change requires knowing JML / __attribute__ / Python decorators / Java exceptions / C setjmp, it does NOT belong in this PR. If it only knows CIDs, memento schemas, effect occurrence sets, policies, loss records, or protocol envelopes, it does.

### DIFF
--- a/implementations/rust/provekit-ir-types/src/lib.rs
+++ b/implementations/rust/provekit-ir-types/src/lib.rs
@@ -1676,10 +1676,181 @@ pub struct CompoundContractMemento {
 //     preservation_claim_cid, provenance_cid, wrapper_fcm_cid
 // ============================================================
 
-/// A promoted semantic effect payload carried by `FunctionContractMemento.effects`.
+// ============================================================
+// Manual extension: EffectOccurrence substrate object (issue #793)
+// Source of truth:
+//   protocol/specs/2026-05-13-effect-occurrence-memento.md
+//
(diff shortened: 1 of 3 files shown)


TSavo:
Force-pushed `5f857c80` with fail-closed fix.

`OccurrenceKind` deserialization now uses `#[serde(try_from = "String")]` + `TryFrom<String>` returning new `OccurrenceKindError`. Bare unknown labels (e.g. `"WeirdKind"`) and malformed namespaced strings (`":kind"`, `"acme:"`) fail closed. The `from_str` helper was also tightened to require both segments non-empty so it's consistent with the deserializer.

Test `occurrence_kind_rejects_bare_unknown` covers all three malformed cases. Closes the gap that #834 named.

Issue #795, opened 2026-05-13
Spec: CCP canonical refusal memento — close the handle-level error gap


TSavo:
## Problem

CCP currently has canonical composition behavior, but refusal/error reporting is not itself a durable substrate object. JSON-RPC and C ABI paths expose errors/handles, while the architecture wants refusals to be content-addressed and replayable.

Composition refusal should be a memento, not only process-local diagnostics.

## Proposed Primitive

Define `CompositionRefusalMemento` or equivalent canonical CCP refusal object.

Expected fields:

- `compose_input_cid`
- `atom_cids[]`
- `effect_occurrences[]`
- `reason_code`
- `blocking_effects[]`
- `missing_memento_requirements[]`
- `diagnostic`
- `ccp_version`
- `provenance_cid`

## Acceptance Criteria

- Adds a normative protocol spec or amends `2026-05-09-contract-composition-protocol.md`.
- Defines refusal CID generation and deterministic encoding.
- Maps current JSON-RPC and C ABI errors to refusal mementos.
- Defines refusal ordering for memento-required, unconditionally-blocked, schema mismatch, determinism violation, and effect tracking gaps.
- Defines how verifiers and `.proof` envelopes consume or embed refusal mementos.

## Related Existing Work

- `protocol/specs/2026-05-09-contract-composition-protocol.md`
- `protocol/specs/2026-05-06-effect-discharge-classification.md`
- `implementations/rust/libprovekit/src/compose.rs`
- `implementations/rust/libprovekit/include/provekit-compose.h`

TSavo:
Tracked by umbrella #796.

TSavo:
Second-wave dependencies surfaced by the spine discipline:

- #799 represents CCP composition executions as generic `RunMemento` instances under a compose `PipelineMemento`.
- #800 carries obligation-level receipts when composition creates or discharges proof obligations.
- #793 supplies the occurrence-level effect payload needed to explain effect-related composition refusals.
- #803 records canonicalization profiles for composition inputs and outputs where canonical objects are minted.

PR #827, opened 2026-05-13
impl(795): Rust types: CompositionRefusalMemento + CCP API conversion

TSavo:
Closes part of #795 implementation.

Substrate-only Rust type per the merged spec. Adds memento struct(s) to `provekit-ir-types` + JCS canonicalization + BLAKE3-512 CID derivation + structural validation methods where applicable + serde round-trip + CID-recompute tests.

**Strict substrate/kit split** (per architect direction):
- This PR: Rust substrate. Types, serialization, CIDs, structural validation. No language syntax knowledge.
- Per-language kit work (native annotation extraction, sugar emission, target-syntax wrappers) is a separate wave per language under `implementations/<lang>/`. Kits consume these types through the plugin protocol.

Rule of thumb: if the change requires knowing JML / __attribute__ / Python decorators / Java exceptions / C setjmp, it does NOT belong in this PR. If it only knows CIDs, memento schemas, effect occurrence sets, policies, loss records, or protocol envelopes, it does.

### DIFF
--- a/implementations/rust/libprovekit/src/compose.rs
+++ b/implementations/rust/libprovekit/src/compose.rs
@@ -36,7 +36,10 @@ use std::sync::Arc;
 
 use provekit_canonicalizer::{blake3_512_of, encode_jcs, Value};
 use provekit_ir_types::{
-    AggregationStrategy, CompoundContractMemento, EvidenceMemento, EvidenceRef, IrFormula, IrTerm,
+    composition_refusal_compose_input_cid, composition_refusal_header_cid,
+    composition_refusal_signature, AggregationStrategy, BlockingEffect, CompositionRefusalEnvelope,
+    CompositionRefusalHeader, CompositionRefusalMemento, CompositionRefusalMetadata,
+    CompoundContractMemento, EffectOccurrence, EvidenceMemento, EvidenceRef, IrFormula, IrTerm, OccurrenceKind, OccurrenceRole,
     Sort, SourceKind, SourceLocator, SourceLocatorPoint, SourceLocatorSpan,
 };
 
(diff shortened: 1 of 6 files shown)


TSavo:
Two correctness bugs patched.

**B1: classification was uniformly `block` for all effects.** Fixed: new `effect_classification()` helper returns per-spec verdict per `2026-05-06-effect-discharge-classification.md` + EffectOccurrence §4:
- `block` for Reads/Writes/Io/Unsafe/Panics
- `memento-required` for OpaqueLoop / UnresolvedCall / EarlyReturn / ClosureCapture / PinnedReference / RawPointerProvenance / PossibleAliasing / Drop / AtomicAccess with `ordering: None`
- `informational-dischargeable` for AtomicAccess with concrete `ordering`

`BlockingEffect.classification` now carries this verdict so refusal consumers can see WHY a given effect blocked composition, not just that some effect did.

**B2: occurrence kind labels were lowercase/snake_case** (`reads`, `opaque_loop`, `raw_ptr_provenance`), would never compose with the canonical EffectOccurrence #793 type which uses `Reads`, `OpaqueLoop`, `RawPointerProvenance`. Fixed: `effect_occurrence_kind()` returns the PascalCase canonical labels. Discharge keys updated to match the §3 example shape (`read:x`, `opaque-loop:<cid>`, `raw-pointer-provenance:<target>:<mut>`, `atomic:<target>:<kind>:<ordering>`, etc.).

Also fixed the role from `"declared"` to `"body"` to match EffectOccurrence #793 §2 (`declared` was the old codex draft name; the spec uses `pre/post/invariant/body/exceptional`).

Note: refusal CIDs computed against impure inputs will change because discharge_key + occurrence_kind values are part of refusal-CID input. `cargo test -p libprovekit` clean.

Local EffectOccurrence duplicate (TODO at top of refusal struct) is a separate concern; will unify with #793 type once both PRs land.

PR #840, opened 2026-05-13
Route contract sugar through realize kits

TSavo:
## Summary

This PR wires contract witnesses through canonical bind lowering without making the Rust substrate own target sugar semantics.

- Extends the realize RPC request with `mode`, married contract payload, `sugar_cids`, and full `sugar_plugins` payloads.
- Lets realize kits return `emitted_artifact_cid`, `observed_loss_record`, and `used_sugars`.
- Threads bind-side compound contract witnesses into canonical rewrite so the target kit can apply loaded sugar dictionaries during emission.
- Adds Java sugar mementos for Bean Validation `@NotNull`, JUnit 5 witness skeletons, and role-aware function comments.
- Implements the Java kit side: pattern-match contract witness predicates against sugar dict entries, emit Java surfaces, and report observed loss/used sugar.
- Pins the new Java sugar CIDs in the substrate default CID drift test.

Architecture boundary: the substrate routes/cites sugar dictionaries and contract witnesses; the Java kit reads the sugar payloads, applies Java surface locators, and computes the observed loss from what it actually emitted.

## Verification

- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-cli realize_request_params_include_contract_mode_and_loss_payload --lib -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-plugin-loader --test substrate_default_cids -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-cli bind_writes_evidence_and_compound_contracts --test cmd_bind_integration -- --nocapture`
- `mvn test -pl provekit-realize-java-core -am -Dtest=JavaNullBoundaryRealizerTest#bindContractWitnessesEmitNotNullSugar`
- `mvn test -pl provekit-realize-java-core -am`
- `JAVA_HOME=/usr/local/Cellar/openjdk/25.0.2/libexec/openjdk.jdk/Contents/Home PATH=/usr/local/Cellar/openjdk/25.0.2/libexec/openjdk.jdk/Contents/Home/bin:$PATH cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-cli --test slice2_java_realize_plugin_byte_identical -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-cli --test compose_rpc_smoke -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-walk compose_method_chain --lib -- --nocapture`
- `git diff --check`

### DIFF
--- a/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/JsonUtil.java
+++ b/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/JsonUtil.java
@@ -109,7 +109,11 @@ final class JsonUtil {
      * Returns the object content (including braces), or "{}" if not found.
      */
     static String extractParamsObject(String json) {
-        String key = "\"params\"";
+        return extractObjectField(json, "params");
+    }
+
+    static String extractObjectField(String json, String field) {
+        String key = "\"" + field + "\"";
         int ki = json.indexOf(key);
         if (ki < 0) return "{}";
         int pos = ki + key.length();
@@ -118,7 +122,11 @@ final class JsonUtil {
             pos++;
         }
         if (pos >= json.length() || json.charAt(pos) != '{') return "{}";
-        // Find the matching closing brace.
+        return extractObjectAt(json, pos);
+    }
+
+    private static String extractObjectAt(String json, int pos) {
+        if (pos >= json.length() || json.charAt(pos) != '{') return "{}";
         int depth = 0;
         int start = pos;
         while (pos < json.length()) {
@@ -128,7 +136,6 @@ final class JsonUtil {
                 depth--;
                 if (depth == 0) { pos++; break; }
             } else if (c == '"') {
-                // skip string content
                 pos++;
                 while (pos < json.length()) {
                     char sc = json.charAt(pos);
@@ -140,6 +147,41 @@ final class JsonUtil {
             pos++;
         }
         return json.substring(start, pos);
+    }
+
+    static List<String> decodeJsonObjectArray(String json, String field) {
+        String key = "\"" + field + "\"";
+        int ki = json.indexOf(key);
+        if (ki < 0) return List.of();
+        int pos = ki + key.length();
+        while (pos < json.length()
+            && (json.charAt(pos) == ':' || json.charAt(pos) == ' ' || json.charAt(pos) == '\t')) {
+            pos++;
+        }
+        if (pos >= json.length() || json.charAt(pos) != '[') return List.of();
+        pos++;
+
+        List<String> result = new ArrayList<>();
+        while (pos < json.length() && json.charAt(pos) != ']') {
+            char c = json.charAt(pos);
+            if (c == '{') {
+                String object = extractObjectAt(json, pos);
+                result.add(object);
+                pos += object.length();
+            } else if (c == '"') {
+                pos++;
+                while (pos < json.length()) {
+                    char sc = json.charAt(pos);
+                    if (sc == '\\') { pos += 2; continue; }
+                    if (sc == '"') break;
+                    pos++;
+                }
+                pos++;
+            } else {
+                pos++;
+            }
+        }
+        return result;
     }
 
     /**
--- a/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/RpcServer.java
+++ b/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/RpcServer.java
@@ -1,9 +1,11 @@
 package com.provekit.realize;
 
+import com.provekit.ir.Blake3;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -86,12 +88,20 @@ public final class RpcServer {
         String function = JsonUtil.decodeJsonStringField(paramsObj, "function");
         String returnType = JsonUtil.decodeJsonStringField(paramsObj, "return_type");
         String conceptName = JsonUtil.decodeJsonStringField(paramsObj, "concept_name");
+        String mode = JsonUtil.decodeJsonStringField(paramsObj, "mode");
+        ContractPayload contract = ContractPayload.fromJson(JsonUtil.extractObjectField(paramsObj, "contract"));
+        List<String> sugarPlugins = JsonUtil.decodeJsonObjectArray(paramsObj, "sugar_plugins");
         List<String> params = JsonUtil.decodeJsonStringArray(paramsObj, "params");
         List<String> paramTypes = JsonUtil.decodeJsonStringArray(paramsObj, "param_types");
         SugarRealizer.Realization r =
-                SugarRealizer.emitStub(function, params, paramTypes, returnType, conceptName);
+                SugarRealizer.emitStub(function, params, paramTypes, returnType, conceptName, mode, contract, sugarPlugins);
         return "{\"source\":" + JsonUtil.quoted(r.source())
-                + ",\"is_stub\":" + (r.isStub() ? "true" : "false") + "}";
+                + ",\"emitted_artifact_cid\":"
+                + JsonUtil.quoted(Blake3.blake3_512(r.source().getBytes(StandardCharsets.UTF_8)))
+                + ",\"is_stub\":" + (r.isStub() ? "true" : "false")
+                + ",\"observed_loss_record\":" + r.observedLossRecord()
+                + ",\"used_sugars\":" + r.usedSugarsJson()
+                + "}";
     }
 
     /**
--- a/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/SugarRealizer.java
+++ b/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/SugarRealizer.java
@@ -8,7 +8,11 @@ import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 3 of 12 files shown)


TSavo:
**CHANGES-REQUIRED**

Adversarial substrate-correctness review against the agreed-on substrate-vs-kit responsibilities (see PR-description summary + the just-landed #823/#830 substrate types). The kit boundary is clean, but the substrate-spine mints are not yet in place: this PR threads kit responses through as raw JSON/strings without minting `RealizationPlanMemento`, `ObservationWrapperMemento`, or validating against `ParametricRealizationMemento`. Per the agreed responsibilities for substrate (#4):

> mint RealizationPlanMemento with the returned values; substrate does NOT interpret loss content or expand sugar templates
> validate plan via `.validate_against(&realization)`
> if observation_wrapper_emission_record present, mint ObservationWrapperMemento

none of those three actions happen on the realize path in this PR.

## Findings

1. **No `RealizationPlanMemento` mint at the realize call-site** — `implementations/rust/provekit-cli/src/cmd_transport.rs:1081-1163` (`realize_function`) and `:1059-1080` (`realize_for_bind_with_contract`). The kit returns `emitted_artifact_cid` + `observed_loss_record` + `used_sugars`, but the substrate only stuffs them into `RealizedSource` strings/JSON. Grep across `implementations/rust/` finds `RealizationPlanMemento` references ONLY in `provekit-ir-types/tests/parametric_realization_serde.rs` and `provekit-ir-types/src/lib.rs:1207` (the type definition). No CLI / transport / bind code consumes it. The plan IS the audit record per spec §1.2; without it, the verifier cannot replay the realize step or chain it back to the bind admission.

2. **No `.validate_against(&realization)` invocation** — same files, same gap. Grep shows the only callers of `validate_against` are `provekit-ir-types/tests/parametric_realization_serde.rs:234,273`. The slot-count check (sort_morphism_cids.len() == required_sort_morphism_slots.len()) is therefore never run on a real bind. Malformed plans (e.g. wrong slot count) will pass through silently.

3. **No `ParametricRealizationMemento` resolution from `concept_site_cid`** — `cmd_bind.rs:1331-1369` builds a `RealizeContractPayload` from `(concept_site_cid, local_contract_cid, origin, discharge_verdict, witnesses)` and ships it straight to the kit. The substrate responsibility "resolve ParametricRealizationMemento + SortMorphismMemento[] from concept_site_cid" is not performed. No `selected_realization_cid`, no `sort_morphism_cids`, no `loss_function_cid` selected per policy. The plan template the kit receives is therefore not the spec-shaped template; it's an ad-hoc contract-witness bundle.

4. **No `ObservationWrapperMemento` mint for `mode ∈ {witness, monitor}`** — `cmd_bind.rs:1356` passes `Some(mode_label(mode))` through to the kit and that's the end of the substrate's involvement with the wrapper. `RealizedSource` in both `cmd_transport.rs:1025-1033` and `kit_dispatch.rs:530-536` has no `observation_wrapper_emission_record` field; `invoke_realize` in `kit_dispatch.rs:758-792` doesn't decode one. Grep across `implementations/rust/` finds zero references to `ObservationWrapperMemento` outside `provekit-ir-types/{src,tests}/`. With `--mode=monitor` being the CLI default, this is the common path, not an edge case. Either wire the wrapper-emission-record decode + mint, or mark explicitly as TODO/follow-up in the PR body (it's currently silent).

5. **No `used_sugars ⊆ cited_sugar_cids` subset check** — `kit_dispatch.rs:780-784` reads `result.used_sugars` directly into `RealizedSource.used_sugars` without filtering against the cited `request.sugar_cids`. A buggy or hostile kit could return sugar CIDs that were never cited; the substrate would accept them and (presumably) record them in any downstream memento. Add a subset assertion: every entry's `cid` MUST appear in `request.sugar_cids`, else refuse.

6. **`observed_loss_record` is opaque-passthrough `serde_json::Value`** — `kit_dispatch.rs:775-779` accepts whatever shape the kit sends and never type-checks it against `LossFunctionMemento`. For determinism + replay this needs at minimum a JCS-stability invariant (canonical bytes, deterministic field order, no timestamps/paths). The Java kit looks careful here (uses `Jcs.encode` + `TreeMap`), but the Rust side enforces nothing.

7. **`mode` arg is `Option<&str>` not the typed `RuntimeMode`** — `cmd_transport.rs:1066`. Loses the type safety carried through `RuntimeMode` in `cmd_bind.rs`. Minor; would be cleaner to plumb the enum and stringify only at the JSON-RPC boundary.

## Substrate-vs-kit boundary verdict

**Clean on the Java-leakage axis.** Grep for `@NotNull|jakarta\.|org\.junit|@Min|@Test|@Disabled` across `implementations/rust/provekit-cli/` and `implementations/rust/provekit-walk/` returns empty. `SugarRealizer.java` is the sole owner of `@NotNull`/JUnit5 surface emission, pattern-matching, JCS encoding of loss records. Federation-by-construction directive respected. `cmd_transport.rs:1085-1100` even has an explicit comment marker that body emission is the kit's responsibility (`let _ = body; // body emission is the kit's responsibility`).

**Broken on the substrate-mint axis** — see findings 1-4. The kit side is correct; the substrate side is degenerate (string-passthrough, no mint, no validation).

## RealizationPlanMemento minting verdict

**Not performed.** This is the load-bearing audit record per spec §1.2 and the agreed substrate responsibility. The kit returns the four substrate-relevant CIDs/records (`emitted_artifact_cid`, `observed_loss_record`, `used_sugars`, optional wrapper-emission), and `RealizedSource` carries them, but `cmd_transport::realize_function` does not assemble a `RealizationPlanMemento { selected_realization_cid, sort_morphism_cids, loss_function_cid, total_loss_record, concept_site_cid, candidate_set_cid, selected_candidate_cid, observation_wrapper_cid }`, does not validate it against the cited `ParametricRealizationMemento`, and does not persist it to the proof bundle. The realize step therefore leaves no content-addressable evidence of WHICH parametric realization was selected against WHICH loss function — the spine of the audit record.

## Wrapper emission verdict

**Not handled, not declared TODO.** `mode` is plumbed all the way to the kit, but the kit response path in `kit_dispatch.rs:758-792` only extracts `{source, is_stub, extension, emitted_artifact_cid, observed_loss_record, used_sugars}`. No `observation_wrapper_emission_record`. No `ObservationWrapperMemento` mint in either bind or transport. The Java `SugarRealizer` does emit JUnit5 witness scaffolding when `contract` has matching sugar entries (good kit-side behavior), but the wrapper FCM CID + observer_effects round-trip back to the substrate is absent. Recommend either (a) add the wrapper field to `RealizedSource` + mint `ObservationWrapperMemento` here, or (b) explicitly state in the PR body that wrapper modes are a follow-up and mode parameter is being threaded only.

## Test results

- `cargo test -p provekit-cli realize_request_params_include_contract_mode_and_loss_payload --lib` — **PASS** (1 ok, 0 failed). Confirms the JSON-RPC request shape carries `mode`, `contract.*`, `sugar_cids`, `sugar_plugins`, and notably asserts `params.get("total_loss_record").is_none()` — i.e. the test itself documents that the substrate is NOT (yet) emitting a total_loss_record envelope per spec §1.2.
- `cargo test -p provekit-walk compose_method_chain --lib` — **PASS** (2 ok, 0 failed). The `compose_method_chain_returns_none_on_missing_contract` rename aligns with `cmd_compose.rs:323-330` now surfacing `refusal.header.cid` (content-addressed refusal — finding 8 from the checklist is in good shape on the compose side).
- `cargo test -p provekit-cli bind_writes_evidence_and_compound_contracts --test cmd_bind_integration` — not run locally (longer integration test, but exercised in the PR's CI green).
- CI: `prove` pass (52s), `CodeRabbit` pass; cross-language conformance gates pending at review time.

## Recommendation

Split decision:
- **The kit work is ship-shape.** The Java emitter is the right shape, owned entirely on the Java side, with JCS-stable loss-record output and proper subset-from-cited-sugars logic at the kit boundary.
- **The substrate-mint work is missing.** The plan-memento mint + validation + wrapper-memento mint + subset assertion are precisely the substrate responsibilities Sir wrote out in the agreed-on cut. Without them this PR is "extend the JSON-RPC request shape" rather than "route contract sugar through realize kits with a substrate audit record".

If the intent is "wire the plumbing now, mint the mementos in the next PR", say so in the body and add explicit `// TODO(realize-plan-mint #NNN)` markers at `cmd_transport.rs:1148-1162` and `kit_dispatch.rs:785-792`. Otherwise, the four blockers above should land before merge.

TSavo:
Rebased against origin/main (4204deab, post-PR#839) and applied 4 substrate blockers + 1 nit flagged in Opus review.

**Conflict resolution (4 files)**
- cmd_compose.rs: HEAD (RpcError::composition_refused canonical pattern from #837)
- chain.rs: HEAD (.map(Some) preserves refusal CID; .ok() drops it, violates Supra omnia rectum)
- cmd_transport.rs: #840 (sugar_cids extraction from sugar_plugins)
- cmd_bind.rs: both (local_contract_cid_for_binding from #840 + PromotionDecisionMemento minting from #839)

**Substrate mints applied**

Blocker #1 - RealizationPlanMemento: minted in new pure function mint_realization_artifacts() in cmd_transport.rs, called from apply_canonical_rewrite after each successful kit dispatch.

Blocker #2 - validate_against: called inside mint_realization_artifacts after plan construction; propagates RealizationPlanError.

Blocker #3 - ParametricRealizationMemento resolution: synthetic inline construction (one slot per param) per the established gap-comment pattern at cmd_bind.rs:1112. Returns CompositionRefusalMemento failure_kind "memento-required-missing" path available for future catalog integration.

Blocker #4 - ObservationWrapperMemento: minted when mode in {witness, monitor, dispatcher} and kit returns observation_wrapper_emission_record. Calls wrapper.validate(&[], &observer_effects_clone, &[]) before persisting (v1: object FCM effects unknown at realize-time; honest loudly-bounded-lossy per Supra omnia rectum). plan.observation_wrapper_cid is set to the minted wrapper's CID.

**Nit** - used_sugars subset check: invoke_realize returns ext:unauthorized-sugar error when kit returns a CID not cited in sugar_cids.

**New fields**
- RealizedSource.observation_wrapper_emission_record: Option<Value> (both kit_dispatch and cmd_transport layers)
- EngineResult.realization_plan_mementos + .observation_wrapper_mementos
- Output dirs: .provekit/bindings/realization-plans/ and observation-wrappers/

**Tests**: 3 unit tests in cmd_transport::mint_realization_artifacts_tests asserting plan IS minted, validate_against passes, and ObservationWrapperMemento IS minted for mode=witness with valid invariants.

All cargo test -p provekit-cli -p provekit-walk -p provekit-ir-types pass except polyglot_smoke::test_daemon_polyglot_smoke (pre-existing: provekit-linkerd binary not in env, unrelated to this PR).

Issue #849, opened 2026-05-13
Bridge D: HTTP payload in the trinity round-trip


TSavo:
Child of #845. Extend the existing trinity round-trip fixture with an HTTP-using function.

## Scope

Do not create a new trinity harness. Add an HTTP payload to the current one.

The fixture should prove:

```text
C source with HTTP library call
  -> lift to concept:http-request / concept:http-response
  -> realize Java HTTP surface
  -> lift back to same concept CIDs
  -> realize Python HTTP surface
  -> lift back to same concept CIDs
  -> realize C terminus
  -> byte-stable terminus under the existing trinity invariant
```

If the active trinity order differs, preserve the actual order and keep the same invariant: library surfaces differ, concept CIDs remain stable.

## Deliverables

- [ ] Add an HTTP-using function to the trinity fixture.
- [ ] Assert the HTTP concept CIDs are present in every lifted leg.
- [ ] Assert the normal byte-stable terminus still holds.
- [ ] Record the loss records selected at each realization leg.

## Acceptance

The existing trinity test proves an API-tier concept, not only primitive/control-flow concepts. This is the empirical receipt for paper 21's cross-library axis.

TSavo:
Closing as old-direction source-transform/transpilation-cycle work.

Current project scope keeps ProofIR/substrate, kit-owned lift/materialize/emit/prove, and LSP/editor surfaces. It does not keep Java -> Rust -> Java, source-to-source conversion, git-filter cross-language editing, or migrate-as-transpiler style work as active product direction.

If there is useful residue, refile it as a narrow current issue against #1486/#1499 or a concrete kit-owned RPC/LSP surface.

Issue #848, opened 2026-05-13
Bridge C: trinity HTTP sugar cells for C, Java, and Python


TSavo:
Child of #845. Author the three HTTP sugar/realization cells needed by the existing trinity harness.

## Scope

Use the active trinity language set. Default target surfaces:

| Language | Library surface | Concept |
|---|---|---|
| C | `libcurl` easy request | `concept:http-request` / `concept:http-response` |
| Java | `java.net.http.HttpClient` or the Java kit's preferred HTTP surface | same concept CIDs |
| Python | `urllib.request` or the Python kit's preferred HTTP surface | same concept CIDs |

If the active trinity fixture has a different language/library convention, preserve the harness and adjust the cell choices only.

## Deliverables

- [ ] C HTTP sugar/realization cell.
- [ ] Java HTTP sugar/realization cell.
- [ ] Python HTTP sugar/realization cell.
- [ ] Loss records for cancellation, timeout, streaming, redirects, and body encoding where the surfaces diverge.
- [ ] Lift-side recognition rules that tag the library call as `concept:http-request`.
- [ ] Realize-side emission rules that choose the target library surface from the concept CID.

## Acceptance

A lifted HTTP call in any trinity language carries the same HTTP concept CID and can realize to the other trinity language surfaces without Rust learning target-specific syntax outside the kit boundary.

TSavo:
Closing as old-direction source-transform/transpilation-cycle work.

Current project scope keeps ProofIR/substrate, kit-owned lift/materialize/emit/prove, and LSP/editor surfaces. It does not keep Java -> Rust -> Java, source-to-source conversion, git-filter cross-language editing, or migrate-as-transpiler style work as active product direction.

If there is useful residue, refile it as a narrow current issue against #1486/#1499 or a concrete kit-owned RPC/LSP surface.

Issue #845, opened 2026-05-13
Cross-library axis — API-tier concept tagging + HTTP trinity receipt


TSavo:
Child of #844 (Cross-X Catalog umbrella). The cross-library axis from paper 21.

## Correct framing

The HTTP trinity payload is **not** the whole solution to cross-platform libraries. It is the first empirical receipt.

The general solution is high-level concept tagging:

```text
library surface -> lift -> concept:<api-level-operation> CID -> realize -> target library surface
```

If the substrate recognizes the concept and all surfaces are tagged as realizations of that concept, the library boundary becomes a routing problem through the CID graph. No translation matrix is required.

## Why CIDs all the way down matters

At every level, identity is content-addressed:

- concepts have CIDs
- library sugar dicts have CIDs
- realization cells have CIDs
- loss records have CIDs
- lifted terms/contracts/implications have CIDs

So cross-library, cross-version, cross-vendor, and cross-platform movement is not `library A -> library B` rewriting. It is `surface -> concept CID -> target realization`, with loss accounted at the realization edge.

## API-tier examples

- `_.isEqual(a, b)` / Zig `std.meta.eql(a, b)` / Python `a == b` / Rust `a.eq(&b)` -> `concept:deep-equality(a,b)`
- `fetch(url)` / `requests.get(url)` / `reqwest::get(url)` / `libcurl` -> `concept:http-request`
- `JSON.stringify(x)` / `json.dumps(x)` / `serde_json::to_string(x)` -> `concept:json-encode`
- Lodash debounce / hand-written timer debounce / framework-native schedulers -> `concept:debounce`

## Exemplar concept: `concept:http-request`

```text
concept:http-request {
  method in {GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS}
  url: concept:url
  headers: concept:header-map
  body: optional(concept:byte-stream | bytes)
}
-> concept:http-response { status, headers, body }

loss-dims:
  sync-vs-async, cancellation, streaming-body, body-encoding,
  retries, timeout, cookie-jar, redirect-policy, tls-pinning
```

HTTP is first because every working developer understands the shape, and the loss dimensions are concrete.

## Bridges

- [ ] #846 Bridge A: API-tier concept tagging for cross-library surfaces
- [ ] #847 Bridge B: mint `concept:http-request` / `concept:http-response`
- [ ] #848 Bridge C: trinity HTTP sugar cells for C, Java, and Python
[...]

Issue #844, opened 2026-05-13
Cross-X Catalog [tracking, paper-21]


TSavo:
Umbrella for operationalizing paper 21 (*After Cross-Language: Every Cross-X Boundary Dissolves the Same Way*, PR #843).

## Thesis

The dissolution mechanism that closed cross-language never used a property of *language*; it used *surface-ness*. The same mechanism dissolves every cross-X axis: library, framework, runtime, architecture, protocol, database, version, platform. The math is M+N per axis, additive across axes.

## Catalog axes (one child issue each)

- [ ] #845 Cross-library — API-tier concept tagging plus the first HTTP trinity receipt
- [ ] Cross-framework — reactive cell, component, route, store
- [ ] Cross-runtime — async task, scheduler, fiber, channel
- [ ] Cross-architecture — packed vector, atomic, fence, prefetch
- [ ] Cross-protocol — RPC, wire encoding, transport, schema
- [ ] Cross-database — relational join, document store, key-value, transaction
- [ ] Cross-version — language-version-bump as realize pass
- [ ] Cross-platform — byte-stream IO, completion model, capability surface

## Mechanism

Every cross-X surface routes through the same shape:

```text
surface -> lift -> concept CID -> realize -> target surface
```

This is true whether the surface is a language operator, library API, framework call, runtime primitive, database query, protocol message, or platform service.

## How this populates — four paths

Paper 21 §6 named three. The 2026-05-13 realization added a fourth:

1. **Authored** — human declares a concept binding / realization cell.
2. **Inferred** — substrate recognizes structural equivalence after lifts.
3. **Generated** — LLM proposes, substrate discharges (#841).
4. **Self-Attested** — library/package author signs their own binding declaration alongside a package release.

Self-Attested is the scaling path, but not the first gate. First gate is an authored trinity receipt proving API-tier concepts survive the existing cross-language loop.

## Net new specs

Zero for the first receipt. SugarDictMemento, CompoundContractMemento, LossFunctionMemento, ParametricRealizationMemento, RealizationPlanMemento, ObservationWrapperMemento, and PEP 1.7.0 plugin protocol are enough. Work is curation + kit wiring + receipts, not protocol.

## Generative successor

#841 (GenerativeCompletionProtocol) eats the long tail: any (concept, library) cell without a Self-Attested or Authored binding gets a Generated candidate that goes through structural-equivalence discharge.

## Operational order

1. #846: record API-tier concept tagging as the general cross-library mechanism.
[...]

PR #851, opened 2026-05-13
concept-shapes(B): mint concept:http-request + concept:http-response, kill cids.tsv absolute-path bug (#847, #73)

### LINKED ISSUE #847: Bridge B: mint concept:http-request / concept:http-response
Child of #845. Mint the HTTP API-tier concepts used by the trinity receipt.

## Scope

Mint durable concept-shape entries for:

- `concept:http-request`
- `concept:http-response`

Supporting concepts should be reused if already present, or minted only where required:

- `concept:url`
- `concept:header-map`
- `concept:byte-stream`
- optional `concept:tls-session`

## Shape

```text
concept:http-request {
  method in {GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS}
  url: concept:url
  headers: concept:header-map
  body: optional(concept:byte-stream | bytes)
}
-> concept:http-response { status, headers, body }
```

Loss dimensions:

- sync-vs-async
- cancellation
- streaming-body
- body-encoding
- retries
- timeout
- cookie-jar
- redirect-policy
- tls-pinning

## Deliverables

- [ ] Concept-shape memento(s) under `menagerie/concept-shapes/`.
- [ ] Realization-gap notes for the loss dimensions above.
- [ ] Minimal examples showing the same concept shape can bind C, Java, and Python HTTP surfaces.

## Acceptance

The HTTP concept CIDs are stable inputs for the sugar-cell bridge and the trinity fixture bridge.

TSavo:
Closes #847 (Bridge B: mint concept:http-request / concept:http-response).
Also closes #73 (cids.tsv absolute-path bug, boy-scout fix in same PR).

Child of #845 → umbrella #844 → paper 21 (PR #843, merged).

## Concepts minted

| Concept | CID |
|---|---|
| `concept:http-request` | `blake3-512:4bba4395e8ce6a46ed110bd89cb5e45f42e89c6e67af063a4eaf9b69d8963ef63baae83be27b97681ebee46c39172472cba39a6547cdcc08a30ac47f57b89684` |
| `concept:http-response` | `blake3-512:42438adf303fb5d9a27fb29ae39e318d3799951a528115fdaa9e4c3ac7f95d4b069d15a4b1348e05535f738fad141b060d91896181533a897bd7464eee058cf9` |
| `concept:url` | `blake3-512:ec39f4820bdac0fc1bfb60e30d7348e2273a89b0f13f7fd3be49b03d206026e5a1f9414aca64ac639aee6ec789dbbd2045309146d118e66c8bccd5b8be654463` |
| `concept:header-map` | `blake3-512:53cb0113c4211d9c326868d5901d0c7e699f4ae23078bea4de54086c1cd59e92ca2accbd5986db0643b7689ea57011e3ed33514614f3626d774fd5cbf011cab7` |
| `concept:byte-stream` | `blake3-512:d5c8fe062ffe4004ccef5c49a90eaebe4f2a3cf46c315c3d966d0214c9ee5311a94355a103094669ea4634bac0fb0e82e703eb35fcbf174810defa86ee02c128` |

`concept:tls-session` was NOT minted; `tls-pinning` is captured as a loss-dimension on `concept:http-request` instead of a separate sort.

## Loss dimensions

- **http-request**: `sync-vs-async`, `cancellation`, `streaming-body`, `body-encoding`, `retries`, `timeout`, `cookie-jar`, `redirect-policy`, `tls-pinning`
- **http-response**: `streaming-body`, `header-case-sensitivity`, `body-decoding`

## Boy-scout: cids.tsv absolute-path fix (#73)

Pre-existing minter scripts wrote absolute worktree paths into `cids.tsv`. Each fresh mint from a new worktree rewrote every row, churning the file with worktree-local data conveying nothing beyond what the repo-relative path already supplies. This PR was about to produce a ~340-line diff of pure path churn before this fix landed.

**Fix**: new `scripts/normalize_cids_paths.py` runs as the final step of `mint.sh`, rewriting any path column to start at `menagerie/concept-shapes/`. Idempotent: a second run is a no-op. One-file post-pass instead of editing the 25+ existing minter scripts that each duplicate the `append_cids_tsv` pattern.

This commit rewrites 339 pre-existing rows from absolute to repo-relative form, in addition to the 5 new HTTP-related rows.

## Verification

- All mint.sh stages pass (existing tests + new HTTP test + new minter + normalize)
- **mint.sh is idempotent**: three consecutive runs from a clean tree produce byte-identical `cids.tsv` (md5 stable at `e8a00152b8a0e3924781c105640aea6e`)
- Zero absolute paths remain in `cids.tsv`
- New `test_http_concepts.py` passes: schema, loss-dims, examples, ASCII-clean
- Existing `test_mint_language_morphisms.py`: passes (boy-scout fix to patch `canonical_cid_spec` for unit-test independence from Rust canonicalizer build state)
- No em-dashes, no en-dashes anywhere in any new file

## Gates downstream

Bridge B's CIDs are stable inputs for:
- **Bridge C** (#848): sugar cells for C/Java/Python, with multi-library Python (urllib, httpx, aiohttp, requests) to enable the cross-library trinity
- **Bridge D** (#849): paper 21 cross-language trinity receipt with HTTP payload
- **Bridge E** (target identifier becomes `(language, library)`): unblocks cross-library trinity composition
- **migrate-CLI**: thin wrapper over E

## Substrate review

Substrate-touching PR. Requires Opus deep-review per the standing rule. Particular spots to look closest:
1. The 5 new shape spec JSONs — schema match against existing shapes?
2. `mint_http_concepts.py` as a callout from `mint.sh` — does this integrate cleanly with the existing per-concept minter pattern, or is it a parallel pipeline that should be folded back?
3. The 339-row absolute→relative rewrite in `cids.tsv` — confirm CIDs themselves are byte-identical to origin/main (only path column changed)
4. The boy-scout patch in `test_mint_language_morphisms.py` — gating canonicalizer dependency on build state

## Summary by CodeRabbit

* **New Features**
  * Added HTTP concept shapes for requests, responses, URLs, header maps, and byte streams to the catalog.

* **Documentation**
  * Updated README with an HTTP Concept Shapes section and added an examples document showing language-specific bindings.

* **Tests**
  * Added unit tests validating generated HTTP concept shape specs and examples.

* **Chores**
  * Added minting and CID-normalization utilities and updated CI attestation/expected CID references.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/851)


TSavo (commented):
Finding: P1 — `concept:http-request` mints the wrong operation shape for Bridge B.

`menagerie/concept-shapes/specs/http-request_shape.spec.json` currently returns `HttpRequest` and its postcondition says it only constructs a request object (`result: HttpRequest`, wp_note: constructs an HTTP request). But #847 defines the concept shape as `concept:http-request {...} -> concept:http-response { status, headers, body }`, and Bridges C/D need this CID to bind real library call sites like libcurl perform / Java HttpClient send / Python urllib open. Those call sites do not just construct a request object; they execute the request and produce response data.

With the current CID, Bridge C cannot honestly tag those library calls as `concept:http-request` without either losing the response semantics or inventing an extra unspecced execute/send concept later. The fix should make the API-tier concept represent the request operation returning `HttpResponse` (or explicitly mint a separate `concept:http-execute` / `concept:http-send` and update #847/#845 before C consumes these CIDs). The tests should also assert this return/result shape so the constructor-vs-operation mismatch cannot recur.

Relevant lines: `http-request_shape.spec.json` return_sort/result at lines 265-267 and 341-345; generator source at `scripts/mint_http_concepts.py` lines 141-152 and 194-201.

TSavo:
## Adversarial substrate-correctness review — verdict: MERGE-WITH-NITS

I verified the eight specific points in the review request and one additional precedent concern (loss_record placement on AlgorithmMemento). The PR is empirically sound and well-bounded; one schema-precedent observation merits an explicit follow-up but does not block.

### What I verified

**1. CID byte-stability vs origin/main.** Diffed the (kind, name, cid) tuples across the 759-row main file and 764-row PR file. **0 rows mutated, 5 rows added, 339 rows path-only-rewritten, 0 rows removed.** Every preserved CID is byte-identical. Final cids.tsv md5 matches PR claim `e8a00152b8a0e3924781c105640aea6e`. Zero absolute paths remain.

**2. Spec schema match.** All five new specs have identical top-level keys to existing templates (`add_shape`, `branch-on-error-else-passthrough`, `break_shape`, `call_shape`): `effects / fn_name / formal_sorts / formals / kind / post / pre / return_sort`. `pre` and `post` substructure matches `operation-contract` shape exactly. The only novel field is `post.loss_record` on http-request/http-response, addressed below.

**3. Round-trip byte-identity.** Ran `build_shape_specs()` from `mint_http_concepts.py` and serialized via the same `json.dumps(..., indent=2, sort_keys=True, ensure_ascii=True)` as `discharge.write_json`. All five output files are byte-identical to the committed specs. Re-minting from a clean tree would reproduce the exact same CIDs.

**4. normalize_cids_paths.py.** Algorithm is correct: anchor guard is `idx <= 0`, which simultaneously rejects "no anchor present" (-1) and "anchor at column start" (0, i.e. already-relative). Column 3 (path) is the only column touched. Cannot rewrite the CID column. Idempotent: two consecutive runs on the post-state are no-ops (verified empirically). Edge cases handled: missing file (logged, exit 0), short rows (passed through), no-anchor rows (passed through).

**5. mint_http_concepts.py integration.** This is **not a parallel pipeline** — it correctly imports `discharge` and reuses `discharge.BASE`, `discharge.SPEC_DIR`, `discharge.CID_FILE`, `discharge.write_json`, `discharge.mint("algorithm", ...)`, `discharge.scan_created_text`. Its `append_cids` helper is actually *stricter* than the surrounding `append_cids_tsv` pattern: it enforces one-name-one-CID with a `SystemExit` on violation, vs the older pattern that silently skips duplicates. Under Supra omnia, rectum the stricter check is the correct one.

**6. Boy-scout test patch.** The `CanonicalCidPatchMixin` uses `setUp`/`tearDown` with `unittest.mock.patch.object`, properly scoped per-TestCase. Real `canonical_cid_spec` is restored on teardown; the actual `mint_language_morphisms.py` invocation later in mint.sh sees the real function. Test passes locally (`Ran 8 tests in 0.004s OK`).

**7. ASCII / em-dash cleanliness.** New scripts produce ASCII-clean output with zero — / – characters. `discharge.scan_created_text` runs at the end of `mint_http_concepts.py` as an extra guard.

**8. Loss-dimension completeness.** http-request dims (sync-vs-async, cancellation, streaming-body, body-encoding, retries, timeout, cookie-jar, redirect-policy, tls-pinning) cover the major Bridge C library divergences for libcurl / java.net.http / urllib/httpx/aiohttp/requests. http-response dims (streaming-body, header-case-sensitivity, body-decoding) are the relevant async surface. Reasonable for first mint; can be extended later via a new mint without invalidating these CIDs only if loss_record location is stable (see nit below).

### Nits / followup (do not block merge)

**N1 — highest priority, precedent-setting:** `loss_record` placement inside `post` on an AlgorithmMemento is a schema novelty. Every other use of `loss_record` in the catalog is at the top level of `RealizationDesugaringMemento` / `LossyMorphismMemento`, with the Rust type `LossRecord` (lib.rs L589-590, L1059-1060). Here it is `Dict[String, IrFormula]` nested one level deep on the concept itself. The minter's serde is permissive (no `deny_unknown_fields`), so the field round-trips and is baked into the CID, but no Rust reader currently interprets `AlgorithmMemento.post.loss_record`. This means: (a) the field is currently decorative, (b) the CIDs `4bba4395...` and `42438adf...` are forever bound to this placement and name. If a later PR formalises a typed `LossRecord` on AlgorithmMemento at the top level (matching the existing pattern), or under a different name like `loss_dimensions`, these two concept CIDs must be re-minted and Bridge C/D/E sugars re-derived. **Followup**: open an issue to lock the semantics ("`post.loss_record` on a concept-shape AlgorithmMemento declares dimensions any realization must characterise") before Bridge C lands, or rename to `loss_dimensions` here while the CIDs are still unused downstream.

**N2:** `HTTP_METHODS` allow-list is 7 entries; RFC 9110 also defines `CONNECT` and `TRACE`. Not wrong (tls-pinning loss-dim implies CONNECT-via-proxy is out of scope for this concept), but undocumented. Add a one-line rationale comment in `mint_http_concepts.py` when the next mint lands.

**N3:** Newly introduced predicate symbols (`http_method_set` ctor, `http_loss_dimension_requires_realization_policy`, `http_status_code`, effect-signatures `NetworkRequest` / `NetworkResponse`) have no central registry. Acceptable for first mint, but Bridge C will need a discoverable home for these.

**N4:** `mint.sh` ordering invariant ("`normalize_cids_paths.py` must run last") is enforced only by line position. A future minter inserted after it silently re-introduces absolute paths. Cheap hardening: add an assert at the top of every minter that fails if any cids.tsv row contains an absolute path, or move normalize into a pre-commit hook.

**N5:** `update_readme()` mutating the human-edited README from a per-concept minter is a new pattern (other minters leave README alone). Defensible here since the new section is bracketed by a clear marker (`## HTTP Concept Shapes\n`), but worth folding into `discharge.py` if more mints adopt the pattern.

### Verdict

**MERGE-WITH-NITS.** Substrate is empirically correct: CIDs are stable, byte-identical, idempotent; tests pass; the only schema-novel field is currently decorative. N1 is the load-bearing followup — track it before Bridge C consumes these CIDs.

TSavo:
Re-review after e0ada889: the P1 constructor/executor issue is fixed. concept:http-request now returns HttpResponse, carries NetworkRequest, and test_http_concepts.py has a recurrence guard. The N1 loss_record placement is also fixed: loss_dimensions is top-level metadata and tests assert loss_record is not nested under post.

Local checks I reran on the latest PR head:
- PYTHONDONTWRITEBYTECODE=1 python3 -m unittest menagerie/concept-shapes/scripts/test_http_concepts.py menagerie/concept-shapes/scripts/test_mint_language_morphisms.py -> 13 OK
- cargo build --manifest-path implementations/rust/Cargo.toml -p provekit-cli -p provekit-canonicalizer -> OK, existing warnings only
- PYTHONDONTWRITEBYTECODE=1 python3 menagerie/concept-shapes/scripts/mint_http_concepts.py -> reproduced committed CIDs, clean worktree
- cids.tsv has zero absolute paths

Merge call: content is ready, but I would not merge while required CI is red. Current red gates are unrelated to the HTTP shape itself: macOS swift pin mismatch (expected 272543..., minted cb743...) and downstream CICP witness refresh/ruby blast-radius mismatch. Branch is also behind main by #842, so update/rebase + witness/pin refresh is the next integration step before merge.

Issue #856, opened 2026-05-14
Substrate honesty pass: rename concept-shape fields by proof-strength gradient


TSavo:
Filed 2026-05-13 from architectural discussion. **Gate this before migrate-CLI (#855) lands.** Bridge B (#851) and Bridge C (#848) ship the current naming as-is; this issue is the cleanup lane that runs before API-tier concepts become load-bearing for `provekit migrate`.

## The rule (T Savo, 2026-05-13)

Don't let a field name claim more proof strength than the payload has. The substrate has an *honesty gradient*; field names should sit at the correct rung:

| Tier | Field-name family | Payload property |
|---|---|---|
| Solver-facing | `IrFormula`, `*_formula`, `*_rule`, `wp_rule` | Dischargeable. A solver can compose, discharge, or refuse. |
| Typed operational fact | `EffectOccurrence`, `effect_*` | Structured + machine-checkable, not a logical formula. |
| Named axis | `LossDimension`, `loss_dimensions` | Names only. Values live elsewhere (per-realization morphism). |
| Sketch | `SemanticSketch`, `contract_sketch`, `shape_contract`, `*_claim`, `*_hint` | Structured but not yet machine-complete. Discharge is promissory. |
| Documentary | `note`, `rationale`, `description`, `*_note` | Human prose. Cannot be discharged; should not appear to be. |

If a field's payload is dischargeable: name it as a contract / formula / rule.
If the payload is structured but not dischargeable: name it as a sketch / claim / hint.
If the payload is prose: name it as note / rationale / description.

## Current overclaiming sites

Cataloging from the just-merged Bridge B concept-shape mints. Each site claims more rigor in its field-name than the payload delivers.

1. **`post.kind: "operation-contract"`** — promises a contract; payload is `{arity, arity_shape, operator, result, wp_note: <prose>, loss_dimensions: <names>}`. The structural metadata is real; the contract content is the prose `wp_note`. Should be `post.kind: "operation-contract-sketch"` or `"shape-contract"`.

2. **`pre.kind: "atomic"` with stringly-typed predicate `name`** — `pre: {kind: "atomic", name: "in", args: [...]}` promises an atomic predicate. But there is no minted predicate-CID for "in" with a discharge rule registered; "in" is a free-form string interpreted by readers. The `atomic` claim outruns the payload.

[...]

Issue #841, opened 2026-05-13
Spec: GenerativeCompletionProtocol — bootstrap new languages from ProofIR descriptions


TSavo:
## Proposal

Formalize how a new language L bootstraps to self-hosting status using only its own surface + a ProofIR description, with no host language past step zero. This is the operational form of paper 07's "generative completion" claim, lifted to the language-addition direction.

## Architectural claim

`program = (V, A, C, ≤)`. Languages are surfaces for that tuple. ProofIR is the canonical home. A language's kit (lifter + realizer + sugars) is itself a program — a (V, A, C, ≤) tuple wearing some surface. Therefore:

**A new language L can be added to the substrate by:**
1. Hand-writing a ProofIR description of L (grammar + sugars + semantics) as a JSON file conforming to ProofIR's wire shape
2. Running `provekit generate-kit --from <seed.json> --target <host>` to materialize L's kit as a runnable process speaking PEP 1.7.0
3. Writing L code that re-expresses L's grammar + sugars + semantics + kit, in L
4. Lifting the L code from step 3 via the kit from step 2
5. Checking: `lift(L_self_description) ≡ seed_proofir` (CID-identical under declared loss)
6. When equal: L is self-hosted; the seed can be dropped

This is a **fixed-point property on the surface ↔ ProofIR functor pair**. Self-hosting becomes a categorical condition, not a build-system condition. ProofIR plays the role classical bootstraps assign to a chain of host languages.

## What this requires (three missing pieces)

1. **`provekit generate-kit` verb**: the substrate operation that emits a kit from a ProofIR description of a language. Generates the kit code, builds it, registers it as a plugin.

2. **LanguageDescriptionMemento substrate primitive**: a content-addressed schema specifying what fields a "ProofIR description of a language" must contain:
   - `grammar_productions: [* production]` — the surface grammar
   - `semantics_rules: [* rule]` — how surface constructs map to (V, A, C, ≤)
   - `sugar_dict_cids: [+ cid]` — citation of SugarDictMementos for this language
   - `canonicalization_profile_cid: cid` — citation of #803 profile for this language's CIDs
   - `contract_conventions: ...` — how contract clauses are expressed in surface (annotations, decorators, comment headers, etc.)
   - `target_hosts: [+ target-host]` — which runtimes the kit can be realized into (rust / jvm / cpython / beam / etc.)

[...]

Issue #796, opened 2026-05-13
Admissibility spine — five-primitive substrate frame


TSavo:
## Model

ProveKit needs a durable admissibility spine that connects learned evidence, promoted substrate semantics, verifier execution, and replayable proof admission.

Core frame:

```text
terms + contracts + implications
  are the substrate objects

evidence + promotion
  are how substrate objects are learned

catalog equations
  are how substrate objects transport

proof runs + receipts
  are how substrate claims become admissible

provenance + plugin registry
  are how the run becomes replayable
```

Effects fit this frame as promoted semantic occurrences, not as a separate substrate layer:

```text
EvidenceMemento -> EffectOccurrence -> FCM.effects -> CCP/verifier classification -> catalog realization/loss_record
```

## Core Tracking Issues

- [x] #791
- [x] #792
- [x] #793
- [x] #794
- [x] #795

## Second-Wave Tracking Issues

- [x] #798
- [x] #799
- [x] #800
- [x] #801
- [x] #802
- [x] #803
- [x] #804

## Implementation Follow-Ups

- [ ] #805
- [ ] #806

## Scope

This epic tracks the substrate admissibility spine. The core issues define the first five independently spec-able primitives:

1. `PromotionDecisionMemento` — hypothesis-to-catalog promotion as content-addressed event.
2. `ProofRunMemento + StageReceipt` — make `provekit prove` a reproducible pipeline.
3. `EffectOccurrence` — define what `FunctionContractMemento.effects` carries.
4. `SortMorphismMemento` — type/sort federation parallel to concept realization.
5. `CompositionRefusalMemento` — make CCP refusal durable instead of handle-local.

The second wave defines the supporting spine objects surfaced by the discipline itself:

1. `PolicyMemento` family — policy-addressed admission gates.
2. `PipelineMemento + RunMemento` — generic replayable run graph.
3. `ObligationReceiptMemento` — durable discharge, counterexample, and tactic receipts.
4. `ParametricRealizationMemento + RealizationPlanMemento` — concept/sort/effect realization planning.
5. `CatalogSnapshotMemento` — pin admitted catalog state.
6. `CanonicalizationProfileMemento` — stabilize CIDs across plugin output drift.
7. `ObservationWrapperMemento` — keep observer effects off object functions.

The implementation follow-ups track work that should happen after the relevant specs land:

1. Bind evidence vNext — carry evidence entries instead of `attr_pre` / `attr_post` strings.
[...]

Issue #791, opened 2026-05-13
Spec: PromotionDecisionMemento — hypothesis-to-catalog promotion as content-addressed event


TSavo:
## Problem

ProveKit has EvidenceMemento, CompoundContractMemento, WitnessMemento, and catalog promotion flows, but no durable event that records why a candidate claim became accepted substrate semantics.

The missing object is the promotion decision itself: which gate fired, which evidence was considered, what policy applied, and which durable memento was produced.

## Proposed Primitive

Define `PromotionDecisionMemento` as the content-addressed event that promotes a hypothesis, witness, compound contract, effect occurrence, or catalog cell into a durable substrate claim.

Expected fields:

- `candidate_cid`
- `promoted_cid`
- `gate`: `human | proof | property | threshold`
- `evidence_cids[]`
- `policy_cid`
- `decider_cid`
- `provenance_cid`
- optional `confidence_basis_points`
- optional `notes` / bounded rationale

## Acceptance Criteria

- Adds a normative protocol spec under `protocol/specs/`.
- Includes CDDL or equivalent canonical wire shape.
- Defines CID inputs and deterministic encoding.
- Defines how this relates to EvidenceMemento and CompoundContractMemento auto-promotion.
- Defines refusal/error cases for insufficient evidence, failed gate, ambiguous candidate, and policy mismatch.

## Related Existing Work

- #689 WitnessMemento
- #754 canonical witness/type work
- Existing evidence lifter PR sequence: test assertions, type signatures, docstrings, native surface extraction
- `protocol/specs/2026-05-13-compound-contract-memento.md`

TSavo:
Tracked by umbrella #796.

TSavo:
Second-wave dependencies surfaced by the spine discipline:

- #798 defines the policy objects referenced by `PromotionDecisionMemento.gate`, `policy_cid`, and `decision_payload`.
- #799 lets promotion runners be represented as generic `RunMemento` instances instead of a one-off `PromotionRunnerMemento`.
- #802 consumes promotion decisions to pin admitted catalog state.
- #803 stabilizes candidate/promoted CIDs across harmless plugin output drift.

PR #857, opened 2026-05-14
spec: ConceptBindingMemento v1.0.0 normative draft

TSavo:
First-class memento family for per-realization concept bindings. Closes the spec gap that Bridge C (#848) surfaced.

## What it specifies

`ConceptBindingMemento` — substrate-admissible content for answering: given concept C and target `(language, library, surface)`, how does the surface realize C, what is preserved, what is lost, what supports the claim?

CDDL in §1 with locked JCS key order. Validation rules in §3 (nine rules; refusal posture is mandatory). CID derivation in §4. Index shape in §5: `(concept_cid, target_language, target_library) -> [binding_cid]`. Substrate consumer contract in §6.

## Architecture

Generic plugin loader stays generic. A dedicated `ConceptBindingRegistry` consumer does typed validation, indexing, query, and refusal. Honors the admissibility rule from #856:

> Identity-valid but behavior-inert CIDs are not admissible default substrate cells.

## Honesty-gradient compliance

- `loss_record` carries TYPED values per the LossRecord schema, not prose-in-atomic-predicate-clothing
- `effect_occurrences` reuses the existing EffectOccurrence typed payload
- `sort_bindings` cites SortMorphismMemento CIDs, not stringly-typed names
- `target_surface` separates surface contract from library install
- Refusal is loud and content-addressed; silent fallback is forbidden

## Provenance modes (uniform on wire, policy-discriminated)

- `authored` — substrate maintainer signs
- `inferred` — structural equivalence discharge
- `generated` — via #841 GenerativeCompletionProtocol
- `self-attested` — library author signs alongside package release (paper 22 scaling path)

## Gates downstream

After merge:
1. Consumer PR — typed Rust structs + `ConceptBindingRegistry` + indexing + validation
2. Bridge C reshape — six HTTP cells rewritten as ConceptBindingMementos conforming to this spec
3. Bridge D — trinity uses binding query API
4. Eventually paper 22 receipt + migrate-CLI

## Cautionary note

Bridge C's original branch (`bridge-c-http-sugar-cells` on origin) is **drafting material only** — the cells there invented `concept_bindings[]` inside opaque content. They are not substrate-admissible. The reshape PR (gated on this spec landing) ships the proper memento family.

## Review needs

Substrate-grade spec, requires Opus deep-review before merge. The nine validation rules are load-bearing; the typed-prose fallback in §2.7 is a transitional accommodation that should be examined adversarially.

## Summary by CodeRabbit

* **Documentation**
  * Added protocol specification for concept binding mementos, defining the wire format and validation rules for binding artifacts that map concepts to concrete surface elements.
  * Included federation correctness requirements to ensure consistent behavior across nodes and migration guidance for existing artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/857)

### DIFF
--- a/protocol/specs/2026-05-14-concept-binding-memento.md
+++ b/protocol/specs/2026-05-14-concept-binding-memento.md
@@ -1,0 +1,356 @@
+# ConceptBindingMemento Normative Spec
+
+**Status:** v1.0.0 normative draft.
+**Date:** 2026-05-14
+**Author:** T Savo
+**Related:**
+- `2026-05-12-plugin-protocol.md` (PEP 1.7.0 plugin protocol envelope)
+- `2026-05-09-language-signature-protocol.md` (LSP; sorts are CID-named)
+- `2026-05-13-effect-occurrence-memento.md` (`EffectOccurrence` typed payload)
+- `2026-05-14-transport-gap-and-partial-morphism-protocol.md` (`LossRecord` typed values)
+- `2026-05-15-concept-hub-abstraction-layer.md` (`ConceptAbstractionMemento` realization framing)
+- `2026-04-30-canonicalization-grammar.md` (JCS canonicalization, normative)
+- `TSavo/provekit#856` (substrate honesty gradient + admissibility rule)
+- `TSavo/provekit#848` (Bridge C: HTTP sugar cells; cells are reshaped per this spec)
+- Paper 21 (`docs/papers/21-after-cross-language-every-cross-x-dissolves.md`)
+- Paper 22 placeholder (*After Vendoring*)
+
+## §0. Purpose
+
+The substrate's `concept:*` hub names abstract operations (`concept:http-request`, `concept:add`, `concept:dynamic-dispatch`) that have, per target `(language, library)` cell, a **realization**: a way the language-and-library combination encodes the abstraction in what it has. The realize step at `provekit bind --target=python:httpx` needs to answer one query:
+
+> Given `concept_cid` C, `target_language` L, and `target_library` Lib, what surface elements in L+Lib realize C, with what typed loss against C's contract?
+
+That query is structurally distinct from two adjacent concerns the substrate already handles:
+
+1. **Surface emission templates** (existing `SugarDictMemento`): how to write the target code as text once a binding has been chosen. Emission is downstream of binding selection.
+2. **Surface recognition / lifting** (existing per-language lifters in `implementations/<lang>/`): how to identify, at lift time, that a surface call IS a realization of `concept:X`. Lifting is upstream of binding selection.
+
+ConceptBindingMemento is the substrate's first-class artifact for the binding itself: the per-realization mapping from concept formal parameters to surface elements, the per-realization typed loss-record, the per-realization effect occurrences, and the discharge evidence supporting the binding. It is queryable, validatable, indexable, and refusable.
+
+### §0.1 Admissibility rule
+
+This spec defines a substrate-admissible memento family per the rule articulated in `#856`:
+
+> Identity-valid but behavior-inert CIDs are not admissible default substrate cells.
+
+A `ConceptBindingMemento` is admissible to the substrate's default registry only when ALL FOUR conditions hold:
+
+1. **Parse**: the JSON conforms to the CDDL in §1.
+2. **Validate**: each typed field (sorts, effects, loss-record values) round-trips through the relevant family's typed parser (`SortMorphismMemento` resolution, `EffectOccurrence` validation, `LossRecord` typing).
+3. **Index**: the binding is recorded into `(concept_cid, target_language, target_library) -> [binding_cid]` and is queryable from the bind/transport/migrate commands.
+4. **Refuse**: malformed content (missing required fields, untyped values, dangling CID references, contradictory loss-record shape) is REJECTED before any CID is admitted to the registry.
+
+Without all four, the artifact is **drafting material**, not substrate. Drafting material may live in `menagerie/` for human review; it MUST NOT be added to the default registry.
+
+### §0.2 What this is not
+
+- Not a `SugarDictMemento`. Sugar dicts describe emission templates; bindings describe concept-to-surface mappings. A realize run consults BOTH: the binding to identify the cell, the sugar dict to author the target text. They are different mementos and live in different consumers.
+- Not a `RealizationDesugaringMemento` (operation-tier desugaring). Bindings sit at the API tier above primitive ops; desugarings sit at the operation tier. Bindings reference desugarings (via the surface_mapping's compiled term) when the binding's surface element itself decomposes algebraically.
+- Not a `ConceptAbstractionMemento` (abstraction-layer hub node). The abstraction-layer node defines the concept; the binding realizes it. One concept abstraction MAY have many bindings.
+- Not an authoring-format file. This spec defines the wire format; the SBL DSL conversation in `project_provekit_libraries_ship_sugar.md` is about a future authoring source format that compiles to this wire shape.
+
+### §0.3 Provenance modes
+
+A `ConceptBindingMemento` may originate by one of four admission paths (per paper 21 §6 and the 2026-05-13 fourth-path realization):
+
+- **Authored**: a substrate maintainer signs the binding.
+- **Inferred**: the substrate clusters two surfaces' lifted terms, recognizes structural equivalence, mints a binding with the equivalence proof as evidence.
+- **Generated**: a model proposes the binding; structural-equivalence discharge produces or refuses it. See `GenerativeCompletionProtocol` (`#841`).
+- **Self-Attested**: the library author signs the binding alongside their package release. The `provenance_cid` cites the author's release-key provenance. This is the scaling path; paper 22 (*After Vendoring*) is built on it.
+
+The substrate registry does not discriminate by path. Verifier policy decides which paths to trust; the wire format is uniform.
+
+## §1. Wire shape (CDDL)
+
+```cddl
+; Shared scalar types:
+;   cid, signature, pubkey, iso8601, json-value
+;
+; Locked JCS key order: alphabetical within each object.
+; Producers MUST emit objects in JCS-canonical key order and MUST omit
+; optional metadata fields when absent.
+
+provenance-mode = "authored"
+                / "inferred"
+                / "generated"
+                / "self-attested"
+                / tstr   ; namespaced extensions, e.g. "ext:multi-sig"
+
+surface-element = {
+  ? args:               json-value,         ; surface-specific argument schema
+  kind:                 tstr,               ; "method-call" / "function-call" / "class-method" / "macro" / "decorator" / ...
+  locator:              tstr,               ; surface-language fully-qualified name, e.g. "urllib.request.urlopen"
+  ? surface_signature:  json-value          ; surface-side typed signature (target-language types as strings)
+}
+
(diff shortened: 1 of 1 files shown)


TSavo:
**Verdict: BLOCK**

Adversarial substrate-correctness review. Five blockers; each violates the spec's own *Supra omnia, rectum* + honesty-gradient framing by silently redefining or mis-citing primitives this spec consumes.

---

### B1. `loss-record` schema mismatch (§2.7)

The transport-gap spec (`2026-05-14-transport-gap-and-partial-morphism-protocol.md` §1.3) defines:

```cddl
loss-record = { * loss-dimension => ir-formula }
```

Values are **`ir-formula`** — formulas characterizing per-dimension divergence (e.g. `c11_result ≡ python_result mod 2^w`). The concept-binding spec §2.7 examples are realization tags, not formulas:

- `cancellation: "AbortController" | "CancelToken" | ...` — string enum, not a formula
- `retry_policy: { kind: "exponential-backoff", max_attempts: N, base_ms: ... }` — structured tag, not a formula
- The §2.7 typed-prose fallback `{kind: "value", representation: "prose", description: tstr}` — also not an `ir-formula`

This silently redefines `loss-record`, or it conflates two distinct concepts:
- **(A) morphism loss** = formulas characterizing divergence-sets per dimension (the existing schema)
- **(B) binding realization-tag** = which target construct realizes a concept slot (what the §2.7 examples actually describe)

These are not the same shape. The spec must pick one. Either:
- (a) The binding genuinely carries divergence-formulas → drop the §2.7 examples and the typed-prose fallback entirely; values must be `ir-formula`.
- (b) The binding carries realization tags → mint a new schema with a different name (`realization-tag-record`?) and stop claiming compatibility with `loss-record §3`.

Shipping (A) and (B) under the same name is the substrate anti-pattern the spec itself names in §0.1.

### B2. `loss_dimensions` field on concept is phantom (§3 rule 6, §2.7)

The spec asserts: "Its keys MUST EXACTLY EQUAL the loss-dimension names declared by the concept-shape's `loss_dimensions` field (per Bridge B)."

`ConceptAbstractionMemento` (`2026-05-15-concept-hub-abstraction-layer.md` §2.1) has fields: `kind, operator, tier, slots, formal_sorts, result_sort, contract, contract_note, realizations, superseded_by, refines`. No `loss_dimensions`. The dimension vocabulary is the open enum on the **`loss-record` schema** (`value_divergence, ub_introduction, domain_narrowing, effect_divergence` + the proposed `structural_divergence`), not a per-concept declaration.

Rule 6 is unverifiable as written. Either:
- (a) Add a `loss_dimensions: [+ loss-dimension]` field to `ConceptAbstractionMemento` in a successor-mint and cite that CID, then enforce equality, or
- (b) Reword rule 6 to validate against the open `loss-dimension` enum, accepting that the schema, not the concept, owns the dimension namespace.

### B3. `ProvenanceMemento` misappropriation (§2.9, rule 9)

`2026-05-06-provenance-memento.md` §1 defines `ProvenanceMemento` as a **C-pointer property record**:

```cddl
ProvenanceHeader = {
  schemaVersion: "1",
  kind:          "ProvenanceMemento",
  target:        tstr,       ; the formal parameter name
  functionCid:   cid,        ; contract CID anchoring the target name
  mutable:       bool,       ; true for *mut T, false for *const T
  properties:    [+ ProvenanceProperty],  ; NonNull / Aligned / Allocated / Readable / Writable
}
```

That is not "who minted the binding, with what signing key, against what trust anchor." It is pointer-NonNull/Aligned properties for raw `*const T` / `*mut T`. There is no existing "minting-authorship" memento family to cite.

Signer authority for any signed memento already lives in `envelope.signer` (Ed25519 pubkey, per the per-memento envelope shape). Rule 9's "signer matches provenance memento's signer" check is structurally vacuous because the cited family does not represent that semantic relationship.

Either:
- (a) Drop `provenance_cid` and `provenance_mode` rule 9; rely on the envelope `signer` + verifier policy keyed on `provenance_mode` enum.
- (b) Propose and land a new `BindingAuthorshipMemento` (or equivalent) family that records "library author's release-key trust anchor" for the self-attested case, and cite that. The four-path provenance discrimination claim in §0.3 cannot stand on the existing `ProvenanceMemento`.

### B4. `RealizationDesugaringMemento` overlap (§0.2 disjointness claim is wrong)

§0.2 claims: "Not a `RealizationDesugaringMemento` (operation-tier desugaring). Bindings sit at the API tier above primitive ops; desugarings sit at the operation tier."

But `2026-05-15-concept-hub-abstraction-layer.md` §2.2 puts `RealizationDesugaringMemento` **at the abstraction tier**, not the op tier. It already carries `target_lang`, `loss_record`, `discharge_receipt`, supports multiple realizations per `(abstraction, lang)` cell, distinguishes via `pre` side-conditions (the same role `target_surface` plays), and discharges by `wp`-preserving check (§2.5).

The reader cannot answer: why is `ConceptBindingMemento` a separate family rather than a successor-mint of `RealizationDesugaringMemento` extended with API-tier discriminators (`surface_mapping`, `binding_kind`)? The spec must either:
- (a) Articulate the structural distinction (what does a binding express that a realization-desugaring cannot? — the only candidate I see is `surface_mapping`'s structural projection vs. `post.rhs`'s op-layer term; that case must be made explicitly), or
- (b) Collapse to a `RealizationDesugaringMemento` successor-mint per LSP §4.4 and avoid minting a parallel family.

### B5. `EffectOccurrence` embedding violates promotion discipline (§2.6 vs effect-occurrence §1.5 + §3)

`2026-05-13-effect-occurrence-memento.md` §1.5 and §1 ¶67 are explicit:

> "the `EffectOccurrence` lives INSIDE `FunctionContractMemento.effects`. The occurrence object itself does NOT carry an `evidence_cid` field; the link from occurrence to evidence is recorded in the `PromotionDecisionMemento` that admitted it."

The concept-binding spec §2.6 embeds `effect_occurrences: [* effect-occurrence]` directly in the binding header, with no `PromotionDecisionMemento` linkage. That bypasses the admissibility-spine promotion discipline #791 / #796 establishes. Two specific problems:

1. **Where is the `PromotionDecisionMemento`?** The spec admits these occurrences into a memento that is not a `FunctionContractMemento`, with no promotion-decision record. Per the cited spec, a producer that mints an `EffectOccurrence` directly MUST still have a `PromotionDecisionMemento` (synthetic-evidence path allowed).

2. **§2.6's `NetworkRequest` example is not a v1 occurrence kind.** `effect-occurrence` §3 lists the canonical kinds: `Reads / Writes / Io / Panics / OpaqueLoop / UnresolvedCall / AtomicAccess / EarlyReturn / Unsafe / ClosureCapture / PinnedReference / RawPointerProvenance / PossibleAliasing / Drop`. `NetworkRequest` is not in that set and would fail closed as `LegacyUnknown` per §6. The example silently assumes an extended kind that doesn't exist.

Either:
- (a) `effect_occurrences` is replaced with a `function_contract_cid: cid` pointer to the realization's FCM, which carries the occurrences through the proper promotion-discipline pipeline, or
- (b) The binding spec lands alongside a `PromotionDecisionMemento` requirement for each embedded occurrence, AND adds `NetworkRequest` (or routes through `Io`) in the effect-occurrence kind catalog before this spec admits the example.

---

### What I verified clean

- **JCS key order**: header keys verified byte-sorted under RFC 8785 (`binding_kind < cid < concept_cid < ... < target_surface`); `schemaVersion < sort_bindings` correct under byte ordering; envelope (`declaredAt, signature, signer`) correct; surface-mapping branches correct; sort-binding (`cid < concept_sort < target_sort`) correct; top-level (`envelope, header, metadata`) correct.
- **No em-dashes / en-dashes** anywhere in the spec body (rule 10 satisfied by the document itself).
- **CID derivation §4**: not circular — `JCS(header_without_cid) → bytes → hash → cid`, then populate, then sign over `JCS(header_with_cid ++ metadata)`. The sequence is fine.
- **§5 index shape**: `target_surface` correctly belongs as a list-discriminator rather than primary key, given §6's selection-policy semantics (registry returns ordered candidates; caller picks). This is the right shape **only if** `target_surface` is policy-visible during query, which it is.
- **§6.2 refusal posture**: correctly forbids silent fallback. The "explicit fallback query" surface is under-specified at the API level but that is non-blocking; the contract is clear.
- **§7 federation correctness**: §3 + §4 produce byte-stable CIDs; secondary indices §5 are pure functions of the primary index, so federation-stable.
- **§9 migration**: Bridge C reshape correctly characterizes the six existing cells as drafting material; the new CIDs cannot collide with the old `concept_bindings[]`-inside-sugar-dict CIDs because the canonical bytes differ in `kind` and envelope shape.
- **`provenance-mode` enum**: the four-path framing reads correctly; the `self-attested` vs `authored` distinction makes sense **conditional on B3 being resolved** (right now both reduce to "envelope signer matches policy" because there is no separate authorship memento).

### Non-blocking observations

- **N1**: §2.2's `target_surface` distinction (contract vs install) is load-bearing and well-argued. Keep it. The naming honesty is correct here.
- **N2**: §3 rule 6 should additionally specify case-sensitivity and ordering rules ("exactly equal" is ambiguous between byte-equality and set-equality after the schema is straightened).
- **N3**: §6.1 initialization order does not specify defer-and-retry vs hard-refuse for unresolved sort-morphism citations. Specify (probably defer-and-retry within a single catalog-snapshot load, hard-refuse across snapshots).
- **N4**: §10 future-work entry 1 (`LossDimensionValueMemento`) is what B1/B2 actually want. Promoting it from future-work to a prerequisite would let this spec land cleanly.

---

The five blockers above each break a load-bearing claim the spec makes about itself: (B1) typed-not-prose, (B2) "concept declares dimensions," (B3) "existing `ProvenanceMemento` carries signer authority," (B4) "not a `RealizationDesugaringMemento`," (B5) "embedded `EffectOccurrence` is just `EffectOccurrence`." A spec for the substrate's first-class binding memento that itself isn't substrate-consistent cannot be the prerequisite for Bridge C reshape / Bridge D / paper 22.

Recommend: rework B1–B3 (the load-bearing ones); resolve or articulate B4; fix B5's embedding shape and the `NetworkRequest` example; resubmit. The four-path provenance frame, the (concept, lang, lib) index, and the refusal posture are all worth keeping.

TSavo:
Closing per architect call (T Savo, 2026-05-14). Opus deep-review surfaced five substrate-correctness blockers; the decisive one is B4 — `RealizationDesugaringMemento` already owns "concept abstraction realizes to target surface with loss and discharge."

ConceptBindingMemento as a parallel memento family is wrong ontology. Library binding is a library-scoped realization profile of the existing family, not a new primitive.

The replacement work is being opened as a different spec: `LibraryRealizationProfile` (or `LibraryRealizationDesugaringMemento`), explicitly refining RealizationDesugaringMemento. Corrections that will land in the replacement:

- `loss_record` stays formula-valued divergence only (no realization-tag strings)
- Realization tags (sync/timeout/cancellation model) move to a separate `capability_profile` field
- No global `concept.loss_dimensions` invariant; validate only where the concept-kind declares dimensions
- No `ProvenanceMemento` citation for authorship; `envelope.signer` is the existing channel
- No directly-embedded `EffectOccurrence`; for HTTP use v1 `Io { channel: "network", operation: "http-request" }` unless deliberate new occurrence kind is specced
- Sugar dicts remain emission sugar; they should not carry semantic binding payloads

The replacement spec PR will be opened shortly. This PR is closed without merge; do not cherry-pick.

PR #866, opened 2026-05-14
docs: paper 22 - After Vendoring (migration as source transformation)

TSavo:
Paper 21's operational consequence at the cross-library axis. Spine is T Savo's canonical articulation of 2026-05-14.

### What this paper says

\`provekit migrate --library-from A --library-to B --write\` is not:
- a compatibility layer
- an adapter
- an ORM
- a facade
- a bridge package
- a runtime shim
- a "common interface" everyone must code against

It is a computably correct transformation over the program graph. The migrated repository uses library B directly; library A is removed from the manifest. The hard part of \`better-sqlite3\` -> \`pg\` is not the callsite rewrite, it is the call-graph contagion as the \`async\` effect propagates upstream. The substrate already tracks effect signatures and the call graph is content-addressed, so propagation is mechanical.

### The receipt is the deliverable

```
132 callsites rewritten
47 functions widened to async
9 boundary handlers already async-capable
3 refused exports because public API forbids promise return
2 lossy sites: sqlite-specific last_insert_rowid semantics
```

Five rows = the substrate trichotomy at the report tier. Each count is a count of signed mementos. Refusal and loss-record are first-class outcomes per the honesty gradient (#856). The body cites every widened function with the propagation reason chain.

### The closing principle

> The substrate only exists to compute and justify the patch.

After the patch lands, the durable artifacts are the patched source and the signed receipt. The substrate is the proof object the receipt cites; it is not a runtime dependency of the migrated codebase. The cypherpunk shape at the migration tier: no enforcement clause, proof clause only.

### Empirical receipt status

Pending. The sqlite -> pg run requires:
- Bridge E: \`library_tag\` named-axis dispatcher widening (codex 498f09c6, in flight)
- Bridge F: permutation driver + cross-library truth-table receipt (queued; corpus locked at \`better-sqlite3\` -> \`pg\` per memory)

This PR sits open until the empirical receipt CID can be cited from \`§9 Empirical receipt\`. The paper is the prediction; the run is the verification; if they disagree, the prediction is wrong, and the substrate has a structural bug more interesting than the paper.

### Files

- \`docs/papers/22-after-vendoring-migration-as-source-transformation.md\` (3336 words)
- \`docs/papers/README.md\` (index entry + ladder sentence updated 21 -> 22)


TSavo:
Review finding (P1): the empirical-anchor section is stale now that the receipt has landed on main.

`docs/papers/22-after-vendoring-migration-as-source-transformation.md` §9 still says the sqlite/pg receipt is pending Bridge E/F and that the example output is predicted rather than produced. Current `main` has #867, #872, and #873 landed; #873 explicitly records the Stage 2 empirical receipt root (`blake3-512:9faa22b51d6bb08e166a0ebd99bf95a21ab3ea61951c6f420840c68fb985d7f523a5bbfc72888d82d1269d4cc50303f8a243f978b76836ada8fe343f6ba88910`) and the exact aggregate counts.

This paper frames the empirical run as the gate for the claim, so merging the stale prediction language would make the paper underclaim the now-landed evidence and point readers at work that is no longer pending. Please rebase/update §9 and the README summary to cite the actual Stage 2 receipt/counts and remove the pending Bridge E/F language.

PR #879, opened 2026-05-14
Add policy-based promotion decision registry

TSavo:
## Summary
- add libprovekit PromotionDecisionRegistry for admitted promotion decisions and per-vector ConsensusPolicy evaluation
- make witness consensus emit v1.1 consensus_vector payloads and require an explicit --consensus-policy for admission
- add witness status plus verify --require-empirically-witnessed --consensus-policy read paths
- cover Stage 3 SQL cross-language witnesses through consensus, registry, and verifier policy gating

## Tests
- cargo test -p libprovekit --test promotion_decision_registry -- --nocapture
- cargo test -p provekit-cli --test witness_consensus_test -- --nocapture
- cargo test -p provekit-cli --test promotion_decision_query_test -- --nocapture
- git diff --check

### DIFF
--- a/implementations/rust/libprovekit/src/lib.rs
+++ b/implementations/rust/libprovekit/src/lib.rs
@@ -7,6 +7,7 @@ pub mod core;
 pub mod desugar;
 pub mod effect_propagation;
 pub mod ffi;
+pub mod promotion_decision_registry;
 pub mod transport;
 pub mod witness_registry;
 pub mod wp;
--- a/implementations/rust/libprovekit/src/promotion_decision_registry.rs
+++ b/implementations/rust/libprovekit/src/promotion_decision_registry.rs
@@ -1,0 +1,378 @@
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use provekit_ir_types::{MigrateReceiptEnvelope, PromotionDecisionMemento, PromotionResult};
+use serde::Deserialize;
+use serde_json::{json, Value};
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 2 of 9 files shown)


TSavo (commented):
Review findings:

1. P1: The registry/query key does not include the promotion tier. `PromotionDecisionKey` is only `(promoted_op, fixture_cid)`, and `index_entries()` indexes every admitted decision with those two fields without checking `decision_payload.promotion`. As a result, `verify --require-empirically-witnessed concept:sql-query --require-fixture X` can accept any admitted promotion for the same concept/fixture whose vector satisfies the supplied policy, even if its payload says a different promotion than `documentary -> empirically-witnessed`. The next fix should either add `promotion` to `PromotionDecisionKey`/`PromotionStatus` and query it explicitly, or have the empirically-witnessed query filter to that exact promotion string. Add a regression where an admitted decision with `promotion: "documentary -> other-tier"` is not accepted by `witness status` / `verify --require-empirically-witnessed`.

2. P2: `consensus_vector()` synthesizes a fallback vector when `decision_payload.consensus_vector` is absent. That weakens the new invariant that verification evaluates the stored observed vector against the supplied policy. A legacy/pre-vector decision can satisfy policies such as `n>=8`, `unique_fixtures>=1`, and `total_sample_count>=8` using derived defaults rather than an actually stored vector. For this slice, I would fail closed for policy-gated empirical promotion unless `consensus_vector` is present, or carry an explicit `vector_source` and require policies to opt into legacy synthesized vectors.

Validation run locally:
- `cargo test -p libprovekit --test promotion_decision_registry -- --nocapture`
- `cargo test -p provekit-cli --test witness_consensus_test -- --nocapture`
- `cargo test -p provekit-cli --test promotion_decision_query_test -- --nocapture`

Issue #880, opened 2026-05-14
Umbrella: concept:contract-observation hub + four-mode composition (lego the existing sugar pieces)


TSavo:
## Purpose

Track `concept:contract-observation` as the hub concept for runtime observation of contracts at ordinary callsites.

This is the shared substrate shape behind witness, monitor, emitter, and gate behavior. It is not a new memento family. It is a concept realization on a wrapper, selected through the existing realization / sugar / loss / policy machinery.

## Canonical Hub Term

Use a real parameterized concept term:

```text
concept:contract-observation(callsite_cid, contract_cid, mode)
```

`mode` is a slot/sort value, not bracket metadata. Existing pattern matching and body-template applicability must be able to inspect it.

## Mode Semantics

| Mode | Meaning | Wrapper effects | Typical surfaces |
|---|---|---|---|
| `witness` | observe and emit signed/runtime evidence | IO + sign | `provekit_witness`, JUnit harness, runtime witness package |
| `emitter` | observe and send to runtime sink | IO | slf4j, JUL, structlog, pino, logging, event bus |
| `gate` | observe and fail/throw/assert on violation | Throw/assert effect | Bean Validation, JUnit assertions, explicit checks |
| `monitor` | passive/least-effect marker for coverage/audit | least effect the wrapper validator permits; never object FCM pollution | liftable comments/tags, coverage markers |

Observer effects live on the observation wrapper only. They must not widen or mutate the wrapped object `FunctionContractMemento`.

## Composition Points

Runtime observation is not textual append. Body-template cells must declare where they compose with the wrapped operation:

| Composition point | Use case |
|---|---|
| `before` | precondition gates, input witnesses |
| `after-return` | output witnesses, success emitters, result-preserving logging |
| `after-throw` | error emitters, failure witnesses |
| `around` | full lifecycle wrappers, try/catch/finally observation |

A result-returning operation must preserve the original result. For `after-return`, the shape is:

```text
result = wrapped_operation(...)
observation(result, ...)
return result
```

Never replace the wrapped result with the observation concept's own return sort.

## Recursive Composition

Observation bodies may contain nested concept citations. Example:

```text
concept:contract-observation(callsite, contract, emitter)
  -> body template cites concept:log-emit(level, message, fields)
  -> language/library sugar realizes concept:log-emit
```

That recursion is tracked in #884. The important invariant is concepts all the way down: logging, witnessing, gates, and comments are all selected surfaces for concepts, not side channels hardcoded into one language kit.

[...]

Issue #737, opened 2026-05-12
## Sugar #3: JUnit5 (Java, inverse of JUnit lifter) — JSON file


TSavo:
JUnit5 test methods as sugar entries. requires(P) → @Test method assertThrows on ¬P; ensures(Q) → @Test method assertTrue Q on a witnessing input. Lines up with --mode=witness runtime mode axis.

Parent: #732 (plugin-protocol umbrella)

TSavo:
Closing as superseded backlog from the older migration/trinity/body-template inventory wave.

The active queue is now organized around #1486 plus concrete per-kit RPC/LSP/materialize/lift/emit/prove issues. Broad migration rows, audit-row placeholders, old body-template completeness items, and one-off catalog inventory mints should be refiled only if they still matter as narrow current work with file paths, tests, and acceptance criteria.

LSP remains in scope; this closure is only for stale migration/inventory bookkeeping.

Issue #735, opened 2026-05-12
## Sugar #1: Spring (Java) — JSON file


TSavo:
Spring annotations as a sugar dict JSON file. Entries: x > k → @Min(k+1), x >= k → @Min(k), x < k → @Max(k-1), x <= k → @Max(k), x != null → @NotNull, s.len() >= k → @Size(min=k). Lives in menagerie/<lang>-language-signature/specs/sugar/spring.json (path per spec).

Parent: #732 (plugin-protocol umbrella)

TSavo:
Closing as superseded backlog from the older migration/trinity/body-template inventory wave.

The active queue is now organized around #1486 plus concrete per-kit RPC/LSP/materialize/lift/emit/prove issues. Broad migration rows, audit-row placeholders, old body-template completeness items, and one-off catalog inventory mints should be refiled only if they still matter as narrow current work with file paths, tests, and acceptance criteria.

LSP remains in scope; this closure is only for stale migration/inventory bookkeeping.

PR #883, opened 2026-05-14
Scope Java contract-observation sugars by mode

TSavo:
## Summary
- add optional `mode` applicability to the sugar-dict spec for contract-observation runtime modes
- make the Java realizer match sugar entries against the full request mode vector
- scope Java Bean Validation sugar to `gate` and JUnit5 sugar to `witness`, with re-minted plugin CIDs

## Verification
- RED first: `mvn -q -pl provekit-realize-java-core -am -Dtest=JavaNullBoundaryRealizerTest#modeScopedJunitWitnessSugarDoesNotApplyToMonitor test` failed before the matcher fix
- `mvn -q -pl provekit-realize-java-core -am -Dtest=JavaNullBoundaryRealizerTest test`
- `cargo test -p provekit-plugin-loader java_`
- `git diff --check`

Refs #735, #737, #880.

### DIFF
--- a/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/RpcServer.java
+++ b/implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/RpcServer.java
@@ -89,12 +89,13 @@ public final class RpcServer {
         String returnType = JsonUtil.decodeJsonStringField(paramsObj, "return_type");
         String conceptName = JsonUtil.decodeJsonStringField(paramsObj, "concept_name");
         String mode = JsonUtil.decodeJsonStringField(paramsObj, "mode");
+        List<String> modes = JsonUtil.decodeJsonStringArray(paramsObj, "modes");
         ContractPayload contract = ContractPayload.fromJson(JsonUtil.extractObjectField(paramsObj, "contract"));
         List<String> sugarPlugins = JsonUtil.decodeJsonObjectArray(paramsObj, "sugar_plugins");
         List<String> params = JsonUtil.decodeJsonStringArray(paramsObj, "params");
         List<String> paramTypes = JsonUtil.decodeJsonStringArray(paramsObj, "param_types");
         SugarRealizer.Realization r =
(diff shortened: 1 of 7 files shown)


TSavo (commented):
Finding:

1. P1: The spec still defines sugar selection as single-winner, but this PR's Java behavior and tests require inclusive multi-mode emission. The updated §4 records all mode-matching candidates, then says to sort them and emit only the TOP candidate (`protocol/specs/2026-05-12-sugar-dict-memento.md:147-165`). The Java consumer now does the opposite: for each witness and each entry, every mode-matching sugar is appended to `out` (`implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/SugarRealizer.java:546-568`), and the new test asserts `witness,gate` emits both Bean Validation and JUnit for the same non-null clause (`implementations/java/provekit-realize-java-core/src/test/java/com/provekit/realize/JavaNullBoundaryRealizerTest.java:129-145`). I think the implementation behavior is the product direction, but the normative spec needs to name it, e.g. split selection by observation mode or introduce an explicit inclusive policy for multi-mode requests. Otherwise a conforming non-Java kit following §4.3 will emit only one of the two sugars and fail the cross-kit semantics this PR is trying to establish.

Checks run locally:
- `mvn -q -pl provekit-realize-java-core -am -Dtest=JavaNullBoundaryRealizerTest test`
- `cargo test -p provekit-plugin-loader java_`
- `git diff --check`

TSavo:
Fixed in 3510bc21.

The sugar-dict spec now names the emission policy explicitly:

- `best-only`: ordinary competing sugar surfaces emit one lowest-loss candidate.
- `inclusive`: mode-composition emits every applicable candidate that survives matching, ordered by loss/tie-break, without discarding lower-ranked candidates.

For `concept:contract-observation`, inclusive mode is now required when the request has a mode vector and selected sugars cover different requested modes. The spec explicitly says a `witness,gate` Java realization with matching JUnit witness sugar and Bean Validation gate sugar MUST emit both, each with its own loss record.

Verification after the spec-only amend: `git diff --check`.

Issue #884, opened 2026-05-14
Catalog cell: concept:log-emit plus body-template citation instance


TSavo:
Parent: #880. Sibling to #882 and #755. Generic recursion is #1023.

## Problem

Monitor/emitter modes need logging, but a wrapper template must not hard-code one logger library. The catalog needs `concept:log-emit` as a normal concept with per-(language, library) realization cells.

This is a catalog/body-template instance, not the generic recursion engine and not a logger implementation project.

## Existing primitive to reuse

- Body-template citation/sub-realization pattern: a template may cite another concept.
- Sugar selection policy #889 chooses which logger surface wins.
- Recursive realization engine belongs to #1023.

## Scope

1. Mint `concept:log-emit(level, message, structured_fields)`.
2. Define loss dimensions: level semantics, structured-vs-formatted fields, sink buffering, context propagation.
3. Add minimal catalog/body-template cells for selected target libraries.
4. Add a reference test showing a monitor/emitter wrapper cites `concept:log-emit` and the generic realization path resolves it.

## Acceptance criteria

- `concept:log-emit` catalog/spec cell exists.
- At least one Java, one TypeScript, and one Python logger surface can be selected by policy or fixture config.
- Project with no logger can refuse or select an explicit floor according to policy; no hardcoded default floor.
- Observed loss records from log-emit are preserved by the wrapper realization.

## Non-goals

- Do not implement generic recursive composition here; #1023 owns it.
- Do not hard-code log4j/slf4j/winston/etc. in the generic realizer.
- Do not conflate monitor with emitter; mode semantics remain governed by #880/#755.

TSavo:
Tracker refinement, 2026-05-15: the log-emit example is correct, but the mechanism needs to be generic so agents do not implement a logger-only special case.

New leaf #1023 tracks the generic realization primitive:

- structured term tree input,
- recursive child realization before parent realization,
- explicit composition points (`before`, `after-return`, `after-throw`, `around`),
- result-preserving wrappers,
- contract-aware composition/refusal,
- recursive observed-loss aggregation.

This issue should remain the log-emit catalog/body-template instance. #1023 is the generic recursive composition engine that makes examples like `log-emit(bitcoin-send(tx, amount))` work without hard-coding logging as a special path.

TSavo:
[...]

Issue #755, opened 2026-05-12
Runtime-mode emission sugars: witness / monitor / emitter / gate


TSavo:
## Problem

`provekit bind --mode=...` needs to select runtime observation behavior around concept callsites without hardcoding that behavior into the Rust CLI or one language kit.

The mode is realized through `concept:contract-observation(callsite_cid, contract_cid, mode)` from #880. A mode chooses which wrapper concept to realize; sugar/body-template policy chooses the concrete target-language surface.

## Modes

| Mode | Behavior | Examples |
|---|---|---|
| `witness` | observe inputs/outputs and emit `WitnessMemento`-compatible runtime evidence | `provekit_witness(...)`, JUnit harness |
| `monitor` | passive/least-effect marker used for coverage/audit/relift recognition | liftable comments/tags |
| `emitter` | observe and dispatch to runtime sink | slf4j, JUL, structlog, pino, Python logging |
| `gate` | fail/throw/assert on contract violation | Bean Validation, JUnit assertions, explicit checks |

`monitor` is not throw-on-violation. That behavior is `gate`.

## Implementation Shape

`bind --mode=witness|monitor|emitter|gate` annotates concept callsites with requested observation modes. The realize side then composes:

1. the ordinary operation body, and
2. one or more `concept:contract-observation(...)` wrappers selected by policy.

Multiple modes may be selected for one callsite when policy is inclusive. Ordering and loss aggregation must be deterministic and recorded.

Mode selection is not the same thing as sugar selection:

- mode chooses the observation role
- sugar/body-template selection chooses the target-language surface
- policy decides inclusive vs best-only vs refuse
- loss records state what each selected surface cannot express

## Acceptance Criteria

- CLI accepts the corrected four-mode vocabulary: `witness`, `monitor`, `emitter`, `gate`.
- Invalid mode names fail closed.
- Realize-side output records the selected mode vector and policy CID.
- At least one Java integration test proves `emitter` composes an ordinary callsite with result-preserving `after-return` behavior. PR #926 is the first slice.
- At least one follow-up test proves `gate` is not confused with `monitor`.
- At least one relift test proves emitted mode tags recover as structured evidence, not only raw comments.
- Multi-mode emission order is deterministic before broad multi-mode support ships.
- Wrapper effects remain on the wrapper and never mutate the wrapped object FCM.

## Child / Related Issues

- #880: `concept:contract-observation` umbrella
- #884: recursive `concept:log-emit` body-template substitution
[...]

PR #926, opened 2026-05-14
Compose Java emitter observation wrappers

TSavo:
## Summary

- Compose Java `--mode emitter` observation wrappers at ordinary callsites using recursive body-template realization: operation body first, then `concept:contract-observation(..., emitter)`, which recursively realizes `concept:log-emit`.
- Preserve the operation result across after-return composition and return a real `observation_wrapper_emission_record` with object FCM CID, wrapper FCM CID, observer effects, preservation claim CID, and policy CID.
- Emit liftable Java comment tags as a policy-selected audit/training surface, then recover those tags in the Java bind lifter as a `native-surface` observation witness.

## Why

This makes observation composition first-order: the runtime emitter is another realized concept, not a Java-only side path. The source tags are optional policy output, but when emitted they close the round-trip learning/audit loop: a later lift can recover which contract-observation concept, mode, contract, object FCM, emitted concept, and policy produced the surface.

## Validation

- `mvn -q -pl provekit-realize-java-core,provekit-lift-java-source -am -Dtest=JavaNullBoundaryRealizerTest,TrinityRoundtripLiftTest test`
- `cargo test -p provekit-cli canonical_emitter_java_composes_result_preserving_observation_wrapper --test cmd_bind_integration`
- `cargo test -p provekit-plugin-loader java_canonical_bodies_cid_self_consistent`
- `git diff --check`

Refs #880 #884 #755

## Summary by CodeRabbit

* **New Features**
  * Observation metadata from source code comments is now extracted and included in analysis results.
  * Generated code now automatically composes contract observations with proper wrapper functions when using emitter mode.
  * Added support for Java canonical emission mode with observation wrapper handling.

* **Tests**
  * Added test coverage for observation policy tag recovery and contract observation composition.
  * New integration tests validate Java canonical emission with observation wrapper generation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/926)

### DIFF
--- a/implementations/java/provekit-lift-java-source/src/main/java/com/provekit/lift/java_source/JavaBindLifter.java
+++ b/implementations/java/provekit-lift-java-source/src/main/java/com/provekit/lift/java_source/JavaBindLifter.java
@@ -90,6 +90,13 @@ public final class JavaBindLifter {
         return new Result(entries, diagnostics);
     }
 
+    Result liftPathsFromSource(String sourcePath, String source) {
+        List<Jcs.Json> entries = new ArrayList<>();
+        List<Jcs.Json> diagnostics = new ArrayList<>();
+        liftSource(sourcePath, source, entries, diagnostics);
+        return new Result(entries, diagnostics);
+    }
+
     private void liftFile(Path root, Path javaFile, List<Jcs.Json> entries, List<Jcs.Json> diagnostics) {
         String rel = root.relativize(javaFile).toString().replace('\\', '/');
         String source;
@@ -99,7 +106,10 @@ public final class JavaBindLifter {
             diagnostics.add(diag("error", "read failed for " + javaFile + ": " + e.getMessage()));
             return;
         }
+        liftSource(rel, source, entries, diagnostics);
+    }
 
+    private void liftSource(String rel, String source, List<Jcs.Json> entries, List<Jcs.Json> diagnostics) {
         JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
         if (compiler == null) {
             diagnostics.add(diag("error", "JDK compiler API unavailable; cannot parse " + rel));
@@ -185,6 +195,7 @@ public final class JavaBindLifter {
             String termShapeCid = Jcs.cid(termShape);
 
             String conceptAnnotation = extractConceptAnnotation(source, fnLine);
+            List<Jcs.Json> observationWitnesses = observationTagWitnesses(source, fnLine);
 
             Jcs.Obj entry = Jcs.object(
                 "attr_post", Jcs.nullValue(),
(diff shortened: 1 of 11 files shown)


TSavo (commented):
Findings:

1. **P1: `RealizationPlanMemento.observation_wrapper_cid` does not resolve to the persisted wrapper file.** `mint_realization_artifacts` computes `observation_wrapper_cid` as the CID of the serialized `ObservationWrapperMemento` and stores that on the plan, but `cmd_bind` writes the wrapper file as `<wrapper.wrapper_fcm_cid>.json` (`implementations/rust/provekit-cli/src/cmd_transport.rs:1211-1214`, `implementations/rust/provekit-cli/src/cmd_bind.rs:455-459`). A consumer walking from the plan to the wrapper by CID cannot find the memento in the output tree. Persist the wrapper under the actual observation-wrapper CID (or add/store that CID explicitly and use it consistently).

2. **P1: `wrapper_fcm_cid` is minted from an in-memory FCM and then dropped.** The Java realizer builds a synthetic wrapper FCM only long enough to hash it, returns just `wrapper_fcm_cid`, and `cmd_bind` persists only the `ObservationWrapperMemento` (`implementations/java/provekit-realize-java-core/src/main/java/com/provekit/realize/SugarRealizer.java:340-370`, `implementations/rust/provekit-cli/src/cmd_bind.rs:455-463`). Per the ObservationWrapper spec, `wrapper_fcm_cid` must resolve and its `effects` must contain every `observer_effects` entry. Downstream admission cannot verify that because the wrapper FCM body is absent; the current local validation sidesteps this by passing `observer_effects` as the wrapper effect surface (`implementations/rust/provekit-cli/src/cmd_transport.rs:1192-1202`). Return/persist the wrapper FCM payload, then validate against the resolved wrapper FCM effects.

3. **P2: relift can attach one method's observation tags to a sibling method.** `scanObservationTags` searches a fixed `fnLine - 25 .. fnLine + 25` window and `tagValue` takes the first tag for each key (`implementations/java/provekit-lift-java-source/src/main/java/com/provekit/lift/java_source/JavaBindLifter.java:337-395`). In generated Java, methods are often closer than 25 lines, and these tags live inside the method body. A plain method immediately after an emitter-wrapped method will recover the previous method's `native-surface` observation witness. Anchor the scan to the current method's source span/body span or to a contiguous tag block owned by that method, and add an adjacent-method regression.

Validation I ran locally:

- `mvn -q -pl provekit-realize-java-core,provekit-lift-java-source -am -Dtest=JavaNullBoundaryRealizerTest,TrinityRoundtripLiftTest test`
- `cargo test -p provekit-cli canonical_emitter_java_composes_result_preserving_observation_wrapper --test cmd_bind_integration`
- `cargo test -p provekit-plugin-loader java_canonical_bodies_cid_self_consistent`
- `git diff --check`

PR #1011, opened 2026-05-15
fix(provekit-cli): wire concept-shape catalog from menagerie into bind

TSavo:
## Summary

Replaces the 2-entry `seed_catalog()` stub with a real loader that walks the menagerie. The TODO comment in cmd_bind.rs:1755 has been pending since the bind CLI landed; this lands the production path.

## What changed

- `CatalogEntry` gains a `shape_cid: String` field.
- `Catalog::match_shape()` tries shape-CID lookup first, falls back to the legacy `classify()` string match for the two stub entries (retry-loop, guard-then-commit). Both pathways coexist.
- `seed_catalog()` walks:
  - `menagerie/concept-shapes/catalog/abstractions/*.json` — abstraction-tier ConceptAbstractionMementos. Name from `memento.operator`, CID from filename or memento body.
  - `menagerie/concept-shapes/specs/*_shape.spec.json` — algorithm-tier op shapes. Name from `fn_name`, CID from JCS(spec).
- Workspace-root discovery via CWD ancestors / `CARGO_MANIFEST_DIR`.

## Empirical signal

Catalog count: **93 entries** loaded (vs 2 in the stub).

`Value::null` lift output now matches at runtime:

```
op=return   concept=concept:return  catalog=LOADED specs/return_shape.spec.json
op=call:new concept=concept:new     catalog=LOADED specs/new_shape.spec.json
MATCH concept:identity              abstractions/concept:identity.blake3-512:6920f6e2...json
```

## Tests

- `cmd_bind::tests::seed_catalog_loads_real_concept_shape_catalog` (asserts >10 entries, finds concept:identity).
- `cmd_bind::tests::catalog_matches_loaded_shape_cid_before_legacy_classification`.

Build green. Existing tests pass.

## Out of scope

- No new memento types.
- No new specs.
- No menagerie data changes (loading existing data only).
- No body-template work in python-canonical-bodies-libprovekit.json (Track C).
- No lifter changes (Track B).

## Summary by CodeRabbit

* **New Features**
  * Concept catalog now uses shape-aware matching for improved accuracy in classification and clustering.
  * Catalog data is now dynamically loaded from disk instead of using static defaults.

* **Tests**
  * Added and updated unit tests to verify shape-based matching and dynamic catalog loading behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1011)

### DIFF
--- a/implementations/rust/provekit-cli/src/cmd_bind.rs
+++ b/implementations/rust/provekit-cli/src/cmd_bind.rs
@@ -1729,11 +1729,12 @@ fn classify_value(value: &serde_json::Value) -> &'static str {
     }
 }
 
-// ---- Seed catalog (v0 stub — production loads signed ConceptAbstractionMementos) ----
+// ---- Concept-shape catalog -------------------------------------------------
 
 struct CatalogEntry {
     id: String,
     name: String,
+    shape_cid: String,
     classification: &'static str,
 }
 
@@ -1742,7 +1743,11 @@ struct Catalog {
 }
 
 impl Catalog {
-    fn match_shape(&self, _shape_cid: &str, shape: &TermShape) -> Option<&CatalogEntry> {
+    fn match_shape(&self, shape_cid: &str, shape: &TermShape) -> Option<&CatalogEntry> {
+        if let Some(entry) = self.entries.iter().find(|e| e.shape_cid == shape_cid) {
+            return Some(entry);
+        }
+
         let cls = shape.classify();
         if cls == "unknown" {
             return None;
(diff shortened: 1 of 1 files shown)


TSavo (commented):
**MERGE-WITH-NITS** — substrate-correctness sweep clears (a)-(g) cleanly.

Adversarial check results:

**(a) Determinism — PASS.** `catalog_json_files` collects via `walkdir`, then `paths.sort()`. OS-stable byte-order across ext4/APFS.
**(b) Signature verification — DOCUMENT-AND-DEFER.** Abstraction docs carry `signature` field; loader reads `memento.operator` and `cid` directly without verifying. Not a regression (the stub had no menagerie load at all), but the substrate's first principle is signed evidence; loading unverified will accrue as structural debt. Open a follow-up to verify `signature` over `memento` before admission, or at minimum log a warning when `signature` is absent. The TODO should be in-tree, not just in PR commentary.
**(c) Symlink/ancestor escape — PASS.** `walkdir` default does not follow symlinks; `menagerie/concept-shapes` contains no symlinks. The ancestor walk only joins `menagerie/concept-shapes` and returns on first `is_dir()` hit; it cannot escape to siblings.
**(d) Backward compat — PASS.** Two legacy entries preserved verbatim in `legacy_classification_entries()`, appended after the loaded set. `match_shape` still falls through to classify-string match. The two prior callers at cmd_bind.rs:823 and :919 pass real shape CIDs from lifts; their behavior on previously-admitted inputs is unchanged.
**(e) CID stability — PASS.** Nothing in this PR touches how lift-side shape CIDs are computed; only how match candidates are sourced. Byte-identical preservation holds for inputs the substrate already admitted.
**(f) Tests — REAL.** Both new tests run on the binary target (`cargo test -p provekit-cli --bin provekit 'cmd_bind::tests::'`) and pass. `seed_catalog_loads_real_concept_shape_catalog` walks the real menagerie. `catalog_matches_loaded_shape_cid_before_legacy_classification` asserts CID-first ordering with a real CID.
**(g) Empirical — VERIFIED with off-by-one.** Filesystem has 18 abstractions + 74 specs + 2 legacy = **94 entries**, not 93 as claimed. Trivial. The `cid` field in every abstraction matches the filename-embedded CID, so `load_catalog_abstraction` produces the same CID via either path.

**Nits:**
1. Loader skips `signature` verification. Add a `// TODO(signature-verify):` in `load_catalog_abstraction` or a follow-up issue; do not let "loaded from menagerie" silently imply "verified."
2. PR description says 93 entries; actual is 94.
3. Latent empty-string match risk: `legacy_classification_entries()` sets `shape_cid: String::new()`. The `find(|e| e.shape_cid == shape_cid)` in `match_shape` is unconditional, so an empty `shape_cid` argument would match the first legacy entry before classify fallback runs. Not triggered today (callers pass real CIDs), but trivially defended with `if !shape_cid.is_empty() { ... }` or by skipping empty-CID entries in the loop.
4. Spec-tier match is not covered by a unit test (only abstraction tier is). The PR's `op=return → spec catalog hit` line is plausible but unasserted; consider a third test that constructs a known spec CID via the same `blake3_512_of(encode_jcs(json_to_value(&spec)))` path and asserts a match against the loaded spec entry.

None of the nits block correctness on previously-admitted inputs. Land it; queue the signature-verification follow-up.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

PR #1012, opened 2026-05-15
fix(provekit-walk): SSA-hoist statement-level mutating method calls into term tree

TSavo:
## Summary

Closes a long-standing lifter gap: statement-position mutating method calls (e.g. `hasher.update(bytes);`) and `&mut` argument positions (e.g. `.fill(&mut out)`) were previously dropped into `loss_record` as `ffi-call-unresolved-effect`; the term tree never saw them. This PR hoists them via SSA-style let-rebindings so the data flow stays faithful to the source.

## The transform

Before:
```rust
let mut hasher = blake3::Hasher::new();
hasher.update(bytes);
```

Lifted (before):
```
let(hasher, call:new(blake3::Hasher::new, []), <continuation drops hasher.update>)
```

Lifted (after):
```
let(hasher, call:new(blake3::Hasher::new, []),
  let(hasher_v1, method:update(hasher, [bytes]), <continuation uses hasher_v1>))
```

Same treatment for `&mut` argument positions: the call result binds new versions for the mutated args.

## Empirical signal: blake3_512_of

Before, the term tree contained `hex::encode(out)` referring to a freshly-zeroed array (because `hasher.finalize_xof().fill(&mut out)` was dropped). After, the term tree contains:

```
method:update(hasher, [bytes])
method:fill(method:finalize_xof(hasher_v1, []), [out])
hex::encode(out_v1)
```

The continuation correctly threads the filled digest.

`loss_record` correctly shrinks: `update`, `fill`, `push_str` entries removed. Genuinely-unresolved FFI symbols (`blake3::Hasher::new`, `finalize_xof`, `hex::encode`, `String::with_capacity`, `len`) remain.

## Tests

237 tests pass including the D7 retirement test set. New test `term_emit_ssa_hoist` substring-asserts the fix on blake3_512_of.

## Out of scope

- No new memento types.
- No new specs.
- No `provekit-ir-types` changes.
- No `cmd_bind.rs` changes (Track A owns that).
- Effect-kind tracking (Alloc, Mut, IO) deferred.
- Loop bodies and conditional branches deferred (linear statement sequences only).

### DIFF
--- a/implementations/rust/provekit-walk/src/emit.rs
+++ b/implementations/rust/provekit-walk/src/emit.rs
@@ -16,7 +16,12 @@
 // JCS+BLAKE3-addressed bundle that downstream substrate tools (lift,
 // linker, mint) can consume.
 
-use std::{cell::RefCell, collections::HashMap, rc::Rc, sync::Arc};
+use std::{
+    cell::RefCell,
+    collections::{HashMap, HashSet},
+    rc::Rc,
+    sync::Arc,
+};
 
 use provekit_canonicalizer::Value;
 use quote::ToTokens;
@@ -338,12 +343,16 @@ struct LossRecord {
 struct LoweringContext {
     return_shape: ReturnShape,
     vars: HashMap<String, ExprSort>,
+    mutable_vars: HashSet<String>,
+    ssa_aliases: HashMap<String, String>,
+    ssa_versions: HashMap<String, usize>,
     losses: Rc<RefCell<Vec<LossRecord>>>,
 }
 
 impl LoweringContext {
     fn from_item_fn_with_losses(item_fn: &syn::ItemFn, contextual_losses: Vec<LossRecord>) -> Self {
         let mut vars = HashMap::new();
+        let mut mutable_vars = HashSet::new();
         for arg in &item_fn.sig.inputs {
             let syn::FnArg::Typed(pat_type) = arg else {
                 continue;
@@ -351,8 +360,12 @@ impl LoweringContext {
             let syn::Pat::Ident(ident) = &*pat_type.pat else {
                 continue;
             };
+            let name = ident.ident.to_string();
             if let Some(sort) = sort_from_type(&pat_type.ty) {
-                vars.insert(ident.ident.to_string(), sort);
+                vars.insert(name.clone(), sort);
+            }
+            if ident.mutability.is_some() {
+                mutable_vars.insert(name);
             }
         }
         let losses = Rc::new(RefCell::new(Vec::new()));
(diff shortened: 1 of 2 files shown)


TSavo (commented):
MERGE-WITH-NITS — substrate-correct change, test passes, the only failing test in the suite is pre-existing on main (not a regression from this PR).

## Verified
- New test `term_emit_ssa_hoists_statement_method_effects` passes locally on the PR branch.
- `cargo test -p provekit-walk` runs 237 unit tests green + every integration suite green EXCEPT `lift_dogfood`, which fails with `name collision on different IR: contract expect@src/dropper/emit.rs:231:24 lifted twice with different bodies`. I confirmed by checking out `origin/main` over the same tree that this exact failure already exists on `main` with the identical error message. **Pre-existing, not introduced here.** Worth a separate issue.
- Implementation walks `&method.args` (Vec, source order) and accumulates via `push_unique` into a Vec; HashMaps are lookup-only. Determinism is intact.
- `method_receiver_source_name` and `mut_borrow_source_name` are implemented (not just claimed). The `&mut` arg path threads through `lower_method_arg_expr_to_value_term` with `statement_mut_args: true`, and the existing-test update on `method:write(sink, [value])` now expects the let-wrapped form — intentional, consistent with the substrate change.
- Term-CID changes for previously-broken functions are correct under Supra omnia, rectum: the old CIDs were addressing wrong substrate (the term tree literally lost the call). Re-derivation is the right outcome.

## Nits (non-blocking, fold into follow-up)

1. **SSA collision-guard absent.** The `<name>_v<n>` scheme has no defense against source identifiers like `hasher_v1` already existing in scope. Two distinct variables could be named identically in the lifted term. No source in the tree currently triggers this, but please add a regression test that exercises a function with a pre-existing `_v1` identifier and asserts the fresh-name choice doesn't fold them. Cheap to do; closes the silent-semantic-break surface.

2. **Positive-half loss_record assertion missing.** The new test asserts `update` and `fill` are absent from `loss_record` (correct), but doesn't assert that the genuinely-unresolved FFI symbols (`blake3::Hasher::new`, `finalize_xof`, `hex::encode`, `String::with_capacity`) are still present. The PR description claims they remain; please pin that with a positive `contains` assertion so it can't silently regress.

3. **CI checks were QUEUED at review time** (CodeRabbit was SUCCESS). I trusted the local cargo run. Worth confirming the cross-language conformance gate + ProvekIt `prove` jobs go green before merge.

Track A / cmd_bind / effect-kind / loops / conditionals are correctly called out as out-of-scope. No memento or spec changes — clean track-B lifter delta.

Verdict: MERGE-WITH-NITS.

PR #1016, opened 2026-05-15
fix(provekit-walk-rpc): enumerate impl-block methods alongside free functions

TSavo:
## Summary

The bind RPC path in walk_rpc.rs only iterated top-level `Item::Fn` definitions when collecting bind lift targets. Files where the public API lives entirely in impl blocks (e.g. \`implementations/rust/provekit-canonicalizer/src/value.rs\` with 7 pub fn impl methods) returned \`bind-lift-empty\` from the JSON-RPC plugin path, even though the standalone \`provekit-walk-emit\` binary handled them fine.

## Fix

Refactor target collection into a shared \`BindLiftTarget\` walker that descends:
- \`Item::Fn\` (top-level free functions)
- \`Item::Impl\` (inherent and trait impl methods)
- \`Item::Mod\` (inline modules, recursively)

Inherent impl methods emit as \`Type::method\`; trait-impl methods emit as \`<Type as Trait>::method\`. Matches the convention used by the standalone binary path.

## Empirical signal

\`provekit-walk-rpc\` lift request against \`value.rs\` now returns 7 entries:
\`Value::kind\`, \`Value::null\`, \`Value::boolean\`, \`Value::integer\`, \`Value::string\`, \`Value::array\`, \`Value::object\`.

Previously: empty.

## Tests

237 cargo tests pass including:
- D7 retirement (PRs #1000-#1008)
- SSA hoist (PR #1012)
- provekit-walk-rpc unit tests
- term_emit_d7 (7 tests)
- term_emit_d7_stmt_expr_other (7 tests)

New test \`bind_lift_includes_public_impl_methods_from_canonicalizer_value\` exercises the RPC dispatch path (not the binary path) on a copy of \`value.rs\` in a temp workspace.

Manual smoke: \`provekit-walk-rpc\` against \`provekit-canonicalizer\` returns the 7 \`Value::...\` impl method entries.

## What's NOT in this PR

- No substrate primitive changes.
- No new memento types, specs, or menagerie data.
- Single file modified: \`implementations/rust/provekit-walk/src/bin/walk_rpc.rs\` (+211/-59).

## Why this unblocks libprovekit-py v1

After this lands, \`provekit lift implementations/rust/provekit-canonicalizer/\` produces non-empty term entries via the canonical CLI path, which is the prerequisite for \`provekit lift | provekit bind | provekit lower --target=python\` end-to-end.

### DIFF
--- a/implementations/rust/provekit-walk/src/bin/walk_rpc.rs
+++ b/implementations/rust/provekit-walk/src/bin/walk_rpc.rs
@@ -348,62 +348,61 @@ fn bind_lift(params: &Value) -> Result<Value, String> {
         let witnesses_by_symbol =
             contract_witnesses_by_function_symbol(&file, &rel, src.as_bytes());
 
-        for item in &file.items {
-            if let syn::Item::Fn(item_fn) = item {
-                let fn_name = item_fn.sig.ident.to_string();
-                let fn_line = line_for_fn(&src, &fn_name) as u64;
-                let (attr_pre, attr_post) = extract_contract_attrs(&item_fn.attrs);
-                let concept_annotation = extract_concept_annotation(&src, &fn_name);
-                let term_shape = term_shape_for_fn(item_fn);
-                let term_shape_cid = blake3_512_of(encode_jcs(&term_shape).as_bytes());
-                let (param_names, param_types, return_type) = fn_signature(item_fn);
-                let function_symbol = format!("{fn_name}@{rel}");
-                let witnesses = witnesses_by_symbol
-                    .get(&function_symbol)
-                    .cloned()
-                    .unwrap_or_default();
+        for target in collect_bind_lift_targets(&file) {
+            let item_fn = &target.item_fn;
+            let fn_name = &target.fn_name;
+            let fn_line = target.line as u64;
+            let (attr_pre, attr_post) = extract_contract_attrs(&item_fn.attrs);
+            let concept_annotation = extract_concept_annotation(&src, &target.source_name);
+            let term_shape = term_shape_for_fn(item_fn);
+            let term_shape_cid = blake3_512_of(encode_jcs(&term_shape).as_bytes());
+            let (param_names, param_types, return_type) = fn_signature(item_fn);
+            let function_symbol = format!("{fn_name}@{rel}");
+            let fallback_function_symbol = format!("{}@{rel}", target.source_name);
+            let witnesses = witnesses_by_symbol
+                .get(&function_symbol)
+                .or_else(|| witnesses_by_symbol.get(&fallback_function_symbol))
+                .cloned()
+                .unwrap_or_default();
 
+            entries.push(json!({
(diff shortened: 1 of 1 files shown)


TSavo (commented):
MERGE-WITH-NITS — The fix is correct, well-scoped, and the load-bearing empirical signal is real: a fresh `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-walk` on the worktree passes (237 + the suite, including `term_emit_d7`, `term_emit_d7_stmt_expr_other`, `term_emit_ssa_hoist`, and the new `bind_lift_includes_public_impl_methods_from_canonicalizer_value`). The new test drives `bind_lift(...)` directly — i.e., the JSON-RPC dispatcher's target for `walk.bind_lift`, not the binary's `emit_term` path. `Item::Mod` recursion only descends inline modules (`Some((_, items))`); `mod foo;` carries `None` content, so the `mod foo { mod foo; }` pathology is structurally inert (different `foo`s, and the file-module form short-circuits). Inherent-vs-trait collision distinction looks sound: `impl_function_qualifier` returns `Type` for inherent, `<Type as Trait>` for trait-impl, so `Foo::new` vs `Bar::new` and `Foo::method` (inherent) vs `<Foo as TraitA>::method` vs `<Foo as TraitB>::method` are all distinguished by prefix.

Nits (do not block merge; track as follow-ups if useful):

1. PR body says the new naming \"matches the convention used by the standalone binary path,\" but `walk_emit`'s `find_fn_in_items` is a *lookup by bare ident* (it accepts `blake3_512_of` as a CLI arg), not an enumerator emitting `Type::method`. The real invariant the two paths now share is \"both descend `Item::Fn` / `Item::Impl` / `Item::Mod`,\" not byte-identical emitted names. Tighten the description on a follow-up.

2. Free-function line numbers silently switched source: old code used `line_for_fn(&src, &fn_name)` (textual scan); new code uses `target.line` from `span_line(item_fn)` (syn span). Tests pass, so no observable break, but the PR frames the refactor as pure extraction when it also changes the line-source for free functions. Worth a sentence in the body or a covering test that pins line semantics across both kinds.

3. The new test only exercises inherent-impl naming (`Value::kind` etc.). No test pins the `<Type as Trait>::method` form for trait impls — concern (e) is implemented but unverified by CI. A short fixture with `impl Trait for Type { fn method(...) { ... } }` asserting the emitted name would close that gap.

4. No test pins nested-`mod` recursion termination, even though the structural argument is straightforward. A 2-level inline-module fixture asserting that nested impl methods come through would harden (g).

Verified on local `cargo test -p provekit-walk`: all suites green. The empirical signal (`Value::kind`, `Value::null`, `Value::boolean`, `Value::integer`, `Value::string`, `Value::array`, `Value::object` now returned from the RPC path against the canonicalizer's `value.rs`, where it previously returned `bind-lift-empty` is the right kind of evidence for this class of fix. Ship.
EOF
)

PR #1018, opened 2026-05-15
feat(provekit-cli + realize-python): framework refuses loudly on body-template gap, no silent stubs

TSavo:
## Summary

Establishes framework-level refusal as the lower verb's exit-code contract: \`provekit lower --target=<lang>\` exits zero IFF the substrate could fully realize every function in the input via the body-template catalog. No partial outputs. No silent stubs. No name-based hardcoding.

## Why

Previous shape (and the recently-closed PR #1017 which made it worse):
- Realize plugin silently emitted \`raise NotImplementedError(...)\` Python stubs into canonical.py when its term-walker couldn't match an operation against a body-template.
- Operator imported canonical.py, smoke test on \`blake3_512_of\` looked plausible, but \`encode_jcs\` and friends raised at runtime.
- canonical.py LOOKED like working substrate output while hiding gaps.

Worse pattern in #1017: realize plugin pattern-matched on function NAME and substituted hand-authored Python bodies. The byte-match for encode_jcs came from those hardcoded strings, not from the pipeline. \`k(I) = t\` collapsed to \`k(handwritten) = t\`.

### Python realize plugin
- Silent stub-emission path removed. Plugin no longer returns Python source containing \`raise NotImplementedError(...)\`.
- New RPC error -32100 "missing body-template entry" with structured \`data\` field containing an array of missing entries: \`operation_kind\`, \`args_shape\`, \`function\` context, \`term_position\`.
- Plugin collects ALL missing entries across the request and returns them in a single response.

### cmd_lower.rs
- Recognizes plugin error -32100.
- Writes no output file. Exits non-zero.
- Prints structured receipt listing every entry the operator needs to author.

## Empirical signal

Running the canonical pipeline on \`implementations/rust/provekit-canonicalizer/\`:

\`\`\`
\$ provekit lift impl/rust/provekit-canonicalizer/ -o /tmp/term.json
\$ provekit bind /tmp/term.json -o /tmp/named.json
\$ provekit lower --target=python /tmp/named.json -o /tmp/canonical.py
ERROR: provekit lower --target=python refused.
The substrate could not realize the input via body-templates.

28 body-template entries needed:
 1. UNNAMED-CONCEPT-1 function=to_cvalue args_shape=[\"& Value\"]
 5. UNNAMED-CONCEPT-5 function=blake3_512_of args_shape=[\"& [u8]\"]
12. UNNAMED-CONCEPT-c function=encode_jcs args_shape=[\"& Value\"]
[...]

Issue #980, opened 2026-05-15
LLM auto-namer: AI narrator that fills UNNAMED-CONCEPT slots in source via comment edits


TSavo:
**Parent**: #943. **Status**: dispatchable. **Depends**: #973 (auto-mint persistence), #979 (the name-lifecycle loop).

The substrate has a CONSUMER-FACING affordance for vocabulary growth: edit `// concept: UNNAMED-CONCEPT-<N>` in source, rerun bind, the substrate persists the name. This ticket operationalizes that affordance at scale by dispatching an LLM agent as the auto-namer.

## The mechanism (uses only existing primitives)

1. **Stage 1**: `provekit bind --rewrite annotate --root <target>`. The annotated source files now carry `// concept: UNNAMED-CONCEPT-N` comments for every cluster the pipeline admitted anonymously. (See cmd_bind.rs:820-842 + 1340-1365; the anonymous-tier path is already wired.)

2. **Stage 2 (this ticket)**: dispatch an LLM agent with a tight self-contained prompt. The agent walks the annotated source files, finds each `UNNAMED-CONCEPT-<N>` annotation, reads the surrounding function body + callers + contract shape, and EDITS the comment to a semantic name.

3. **Stage 3**: rerun `provekit bind`. The lift kit reads the human-named annotations and threads them through as `concept_annotation`. Promotion-decisions now carry `decision_payload.promoted_op = "concept:<llm-name>"`.

4. **Stage 4**: `mint_from_promotion_decisions.py` (#973) reads the admitted promotions and writes new op-spec files. Substrate vocabulary has grown.

## Author this driver

`bootstrap/scripts/auto_name_concepts.sh`:
- Take a stage-1 bind output directory + a target source root.
- Find every `// concept: UNNAMED-CONCEPT-<N>` annotation in the target.
- Group by context (the function the annotation precedes).
- Dispatch ONE LLM agent (Opus or Claude Code via Agent tool) with a self-contained prompt:
  - Mission: name each UNNAMED-CONCEPT-N by reading the function and context. Edit the comment in place.
  - Constraints: do NOT rename catalog-matched concepts. Do NOT introduce new specs. Do NOT remove the comment, only change the name part. Names must be kebab-case, prefix-free (the `concept:` prefix gets added by bind).
  - Output: edited source files in place.
- After the LLM agent completes, run em-dash sweep on the diff.

## Acceptance

[...]

PR #1019, opened 2026-05-15
feat(provekit-cli/bind): per-operation concept naming via term-tree walk

TSavo:
## Summary

Deepens \`provekit bind\` so concept names are assigned at EACH operation node of a function's term tree, not just at the function-level root. Unlocks N:1 amortization: one body-template entry for \`concept:method-update\` covers every function that uses \`.update()\`, across every library.

## Why

PR #1018 established framework-level refusal — \`provekit lower --target=python\` exits non-zero with a structured -32100 receipt when body-template coverage is incomplete. The receipt currently lists 28 UNNAMED-CONCEPT-N entries at function-level granularity. Each entry requires a per-function body template; no amortization.

Catalog \`menagerie/concept-shapes/specs/\` already has 264 operation-level shape specs (\`concept:call\`, \`concept:seq\`, \`concept:method-update\`, etc.). They just weren't being matched because bind only looked at the root.

## What changed

cmd_bind.rs walks the term tree recursively. At each node:
- Computes the operation's shape CID (node structure with arg subterms abstracted).
- Looks up the catalog by that shape CID.
- Annotates the node with \`concept_name\` (matched OR UNNAMED-CONCEPT-N).

Output: \`NamedTerm\` now includes \`namedTermTree\` mirroring the term structure, with each node carrying \`operationKind\`, \`conceptName\`, \`shapeCid\`, recursive \`args\`. Function-level \`rootConcept\` preserved for backward compat.

Catalog loader extended to add operation-shape CID entries (was only registering root-shape entries).

\`--project\` accepted as alias for existing \`--root\`.

## Empirical signal

Lifted canonicalizer (28 functions), counted per-operation concept matches across all \`namedTermTree\` annotations:

\`\`\`
32 concept:call
30 concept:decl
34 concept:seq
 3 concept:conditional
 2 concept:eq
 2 concept:for
\`\`\`

Sample namedTermTree slice for blake3_512_of:

\`\`\`json
{
  "function": "blake3_512_of",
  "rootConcept": "UNNAMED-CONCEPT-5",
  "namedTermTree": {
    "operationKind": "seq",
    "conceptName": "concept:seq",
    "args": [
      {"args": [], "conceptName": "concept:decl", "operationKind": "decl"},
      {"args": [], "conceptName": "concept:call", "operationKind": "call"},
      {"args": [], "conceptName": "UNNAMED-CONCEPT-3", "operationKind": "opaque"}
    ]
  }
}
\`\`\`

## Tests

- cargo build clean.
- cargo test green: cmd_bind_integration (3), trinity_roundtrip (1), verb_composition (1), doc-tests.
- New test \`bind_names_blake3_512_of_operations_from_catalog\` asserts operation-level matches.

## What's NOT in this PR

- Realize plugin extension to consume \`namedTermTree\` and emit per-operation Python (next track L).
- LLM auto-namer #980.
- New concept hubs in menagerie.

## Scope

NO substrate primitive changes outside \`cmd_bind.rs\`. NO new memento types, specs, or body-template data. Em-dash clean.

## Release Notes

* **New Features**
  * Added optional tree structure output for bound named terms, enabling better visibility into term composition with concept names.
  * Added `--project` as a CLI argument alias for `--root`.

* **Tests**
  * Expanded test suite with new test cases for term binding operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1019?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/implementations/rust/provekit-cli/src/cmd_bind.rs
+++ b/implementations/rust/provekit-cli/src/cmd_bind.rs
@@ -556,7 +556,6 @@ fn operation_kind(value: &Json) -> Option<String> {
     let normalized = match raw_kind {
         "body" | "block" => "seq",
         "if" => "conditional",
-        "exit" => "return",
         "let" => "decl",
         "bin" => value
             .get("op")
--- a/implementations/rust/provekit-cli/src/cmd_bind.rs
+++ b/implementations/rust/provekit-cli/src/cmd_bind.rs
@@ -34,7 +34,7 @@ pub struct BindArgs {
     pub output: Option<PathBuf>,
 
     /// Legacy migration root. Kept for cmd_bind_migrate compatibility.
-    #[arg(long, default_value = ".")]
+    #[arg(long, alias = "project", default_value = ".")]
     pub root: PathBuf,
 
     /// Source language hint for diagnostics and named-term metadata.
@@ -203,6 +203,12 @@ pub struct NamedTerm {
     pub file: String,
     pub function: String,
     pub name: String,
+    #[serde(
+        default,
+        rename = "namedTermTree",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub named_term_tree: Option<NamedTermTree>,
     #[serde(rename = "paramTypes")]
     pub param_types: Vec<String>,
     pub params: Vec<String>,
@@ -215,6 +221,17 @@ pub struct NamedTerm {
     #[serde(rename = "termShapeCid")]
     pub term_shape_cid: String,
     pub witnesses: Vec<NamedWitness>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NamedTermTree {
+    pub args: Vec<NamedTermTree>,
+    #[serde(rename = "conceptName")]
+    pub concept_name: String,
+    #[serde(rename = "operationKind")]
+    pub operation_kind: String,
+    #[serde(rename = "shapeCid")]
+    pub shape_cid: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -326,6 +343,7 @@ fn bind_term_document(term_json: &Json, args: &BindArgs) -> Result<NamedTermDocu
     let mut seen_names: BTreeSet<String> = BTreeSet::new();
     let mut terms = Vec::with_capacity(entries.len());
     let mut decisions = Vec::new();
+    let mut operation_namer = UnnamedConceptNamer::default();
     for (idx, entry) in entries.into_iter().enumerate() {
         let concept_name = concept_name_for(&entry, idx + 1, &catalog);
         let name = unique_name(&concept_name, &mut seen_names);
@@ -338,6 +356,8 @@ fn bind_term_document(term_json: &Json, args: &BindArgs) -> Result<NamedTermDocu
         let site_memento_cid = site_cid(&entry, &name, &term_shape_cid)?;
         let witnesses = named_witnesses(&entry);
         let promoted_cid = blake3_512_of(format!("provekit-bind/promoted/{name}").as_bytes());
+        let named_term_tree =
+            named_operation_tree(&entry.term_shape, &catalog, &mut operation_namer)?;
         decisions.extend(promotion_decisions(
             &term_shape_cid,
             &promoted_cid,
@@ -354,6 +374,7 @@ fn bind_term_document(term_json: &Json, args: &BindArgs) -> Result<NamedTermDocu
             file: entry.file,
             function: entry.fn_name,
             name,
+            named_term_tree,
             param_types: entry.param_types,
             params: entry.param_names,
             return_type: if entry.return_type.is_empty() {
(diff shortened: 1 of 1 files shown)


TSavo (commented):
**BLOCK** — substrate field ships wrong concept names for break/continue.

The PR is mostly clean and the framework architecture is right: refusal preserved (empirically: lower still exits non-zero with no .py written on synthetic IR), `rootConcept` unchanged, `namedTermTree` correctly nests (`concept:seq` -> `concept:decl` / `concept:call` x2 / UNNAMED-CONCEPT-1 on opaque), catalog loader cleanly adds operation-shape entries, BTreeMap-only (no HashMap nondeterminism), em-dash clean, zero function-name pattern matching. All 9 tests green (cmd_bind 4, cmd_bind_integration 3, verb_composition 1, trinity_roundtrip 1).

**Required fix (correctness, Supra omnia rectum):**

Line 559: `\"exit\" => \"return\"` actively mislabels `break` and `continue`. The Rust lifter at `provekit-walk/src/bin/walk_rpc.rs:1367-1369` collapses `syn::Expr::Return | Break | Continue` to `kind: \"exit\"`. PR #1019 then renames every `exit` to `return`, so every loop `break`/`continue` gets `conceptName: \"concept:return\"`. The catalog already has distinct `concept:break`, `concept:continue`, `concept:return` specs (menagerie/concept-shapes/specs/), so bind is manufacturing a wrong concept name that contradicts an existing catalog entry.

Fix: drop the `\"exit\" => \"return\"` line and let `\"exit\"` fall through to `other => other`. Exit nodes will then look up shape `{operator: \"exit\"}`, find no catalog match (no concept:exit exists), and get `UNNAMED-CONCEPT-N`. That's truthful (loss is in the lifter, bind is bounded). A follow-up either (a) mints `concept:exit` in the catalog, or (b) fixes walk_rpc to emit distinct `return`/`break`/`continue` kinds.

**Nit (non-blocking):**

`namedTermTree.args` slot order depends on the lifter emitting fields in semantic slot order plus `serde_json/preserve_order` (workspace Cargo.toml line 91 enables it). Today's Rust lifter emits `if` as `cond, then, else` in order, so this works empirically. But it's fragile -- a future lifter without preserve_order, or emitting fields alphabetically, silently breaks downstream consumers. Worth a comment in cmd_bind.rs documenting the contract: `child_operation_trees` requires `serde_json/preserve_order` AND lifter slot-order discipline. CodeRabbit and Codex both flagged this as P1.

CodeRabbit's other concern (operation_lookup_shape collapses structurally-different specs to one CID) is misanalysis: each concept has a unique `fn_name`, the catalog only registers `{operation-shape, operator}` shape CIDs for entries starting with `concept:`, and the collapse-by-operator is the entire point of N:1 amortization. Not a bug.

Once the exit conflation is fixed, MERGE.

TSavo (commented):
**MERGE.** Re-review pass after the BLOCK fix.

**(a) Offending substitution removed.** `cmd_bind.rs` no longer contains any of `"exit"`, `"break"`, `"continue"`, or `"return"` string literals. The `"exit" => "return"` arm is gone.

**(b) `exit` routes honestly.** Path: `operation_kind()` at line 548 trims the namespace prefix, then `match raw_kind { ... other => other }` at line 570 passes `"exit"` through unchanged. Catalog lookup at line 487 receives `operation-shape{operator: "exit"}`. The catalog has distinct `break_shape`, `continue_shape`, `return_shape` specs (verified at `menagerie/concept-shapes/specs/`) and no `exit_shape`, so the lookup refuses-loudly and the op becomes `UNNAMED-CONCEPT-N`. This is exactly the (b)(i) behavior, no false `concept:return` label.

**(c) Other BLOCK conditions still resolved.**
- Function-name pattern matching: zero occurrences in cmd_bind.rs.
- Determinism: only `BTreeMap`/`BTreeSet` (line 11), no `HashMap`/`HashSet`.
- Function-level naming preserved: empirical probe shows `term[0].conceptName == concept:guard-then-commit`.
- `namedTermTree` well-formed: present on terms, walks recursively.
- Downstream consumers: `cmd_lower`, `trinity_roundtrip_test`, `verb_composition` all green.
- PR #1018 refuse-loudly behavior intact (ci_check, stage3, transport_cli, migrate_async_rewrite all green).
- Empirical per-op naming present and honest (see (d)).

**(d) Tests green.** Full `cargo test -p provekit-cli`:
- `cmd_bind_integration` 3/3
- `verb_composition` 1/1
- `trinity_roundtrip_test` 1/1
- `provekit-cli` unit 9/9
- `provekit` (main.rs) 96/96
- All integration binaries: 0 failed.

**(e) Em-dash sweep clean.** Diff vs `origin/main`: zero `—` or `–` introduced.

**(f) No hidden coupling.** Searched all of `implementations/rust/` for `exit.*concept:return` or `concept:return.*exit` orphan assertions, zero hits. The fix was one-line deletion, no test rewired.

**Empirical probe (the load-bearing check):**

Fed a synthetic term-shape with `if amount < 0 { exit balance } let next = balance + amount; exit next` through `provekit bind`. Output:

```
              seq -> concept:seq
      conditional -> concept:conditional
               lt -> concept:lt
              seq -> concept:seq
             exit -> UNNAMED-CONCEPT-3
             decl -> concept:decl
              add -> concept:add
             exit -> UNNAMED-CONCEPT-7
```

Both `exit` ops surface as `UNNAMED-CONCEPT-N`, not as `concept:return`. That's the entire fix, observable end-to-end. Under Supra omnia, rectum the framework correctly refuses to commit to `concept:break` vs `concept:continue` vs `concept:return` when the Rust lifter cannot distinguish.

The substrate now reports what it actually knows, no more.

Issue #1076, opened 2026-05-16
A10: Python source lifter collides bind CIDs for x+y and x-y (substrate correctness violation)


TSavo:
**Title:** A10: Python source lifter preserves operator atoms in bind-IR entries

**Labels:** substrate-correctness, area:realize-side, follow-up, priority:p1

**Parent:** #1024 / #1068 Path B. **Surfaced by:** composition test census (origin/1073-composition-test-census, seam 4 discrimination test currently marked `#[should_panic]`). **Related:** A9 (#1075, federation type-erasure at hub), A11 (Python realize body-template coverage gap, possibly resolves the 2 UNNAMED-CONCEPT entries it produces).

---

## Problem

The composition test census surfaced a discrimination-test failure at seam 4 (Python relift Term → BindKit). The Python source lifter at `implementations/python/provekit-lift-python-source` produces bind-IR entries that drop the operator's identity. As a result, distinct operators on the same operands collide:

```
lift_python("def f(x, y): return x + y") -> bind_CID = X
lift_python("def f(x, y): return x - y") -> bind_CID = X    # SAME, should differ
```

The bind-IR's NamedTerm entries lack a structural atom for the operator. The lifter encodes the function's shape (params, return position, term tree skeleton) but omits the operator op-CID. Since the substrate's algebra distinguishes `concept:add` from `concept:sub`, the bind output should reflect that distinction structurally. It does not.

This is a mechanical fix in the Python source lifter. The producer-consumer pattern: the Python source lifter is a producer of `Term`, and BindKit is the consumer; A10 closes the integration gap that BindKit's discrimination output is collapsed because of an upstream omission.

## Already present on main

- `implementations/python/provekit-lift-python-source` (the Python source lifter package).
- The Python source lifter's bind-IR emission path (location varies by file; the lifter emits bind-IR entries via the `bind_lifter.py` module per the #1022 carrier work).
- The composition test census on `origin/1073-composition-test-census` (SHA 319ad456). Seam 4 discrimination test is `#[should_panic]`, will transition to clean pass when A10 lands.
- Concept op-CIDs for the standard operators (`concept:add`, `concept:sub`, `concept:mul`, `concept:div`, `concept:eq`, `concept:lt`, etc.) are already minted in `menagerie/concept-shapes/catalog/algorithms/`.

## Scope

[...]

PR #1074, opened 2026-05-16
feat(tests): trinity composition test census for seams 1, 3, 4, 6 (closes #1073)

TSavo:
Closes #1073. Operationalizes the trinity composition test census (`docs/plans/2026-05-16-trinity-composition-test-census.md`) for the four seams (1, 3, 4, 6) not already covered by A7 or A8.

## Parallel-surfacing payoff

The census ran end-to-end against real toolchains and surfaced **three distinct gaps in one PR** rather than discovered serially:

- **A9 (federation gap, seam 4):** rust-lift and python-lift produce divergent bind CIDs for `f(x, y) = x + y`. NamedTerm diff captured: `param_types = [i64, i64]` (Rust) vs `[Any, Any]` (Python); `return_type = i64` vs `Any`; term_shape and named_term_tree differ structurally. The bind layer is preserving language-tier type information that should be projected away for cross-language federation.
- **A10 (CID collision, seam 4):** Python source lift produces byte-identical bind CIDs for `x + y` and `x - y`. Operator atoms are missing from bind-IR entries. **This is a substrate correctness violation, not a federation gap.** Two semantically distinct operations are collapsing to the same CID. Higher priority than A9.
- **A11 (realize coverage, seam 3):** Python realize refuses with `missing body-template entry` for `concept:conditional`, `concept:eq`, `concept:decl`, `concept:lt`, `concept:mul` plus 2 UNNAMED-CONCEPT entries for `safe_divide_then_double`. Body template catalog is incomplete.

These are filed as follow-up issues (A9, A10, A11) and tracked via `#[should_panic]` antibody tests in this PR. When each gap closes, removing the `#[should_panic]` attribute promotes the antibody to a passing assertion.

## Tests (9 total, all passing)

Slow-test lane gated; default `cargo test` unchanged.

- **Seam 1 (lift to bind, UNTESTED on main):**
  - Positive: real Rust source lifted, fed to BindKit, result is `Term::Op { concept:bind-result, args.len() == 2 }`. Clean pass.
  - Discrimination: malformed Term refuses cleanly with typed error. Clean pass.
- **Seam 3 (lower to python relift, UNTESTED on main):**
  - Positive: `#[should_panic]` antibody for **A11** (realize body-template coverage gap).
  - Discrimination: malformed concept-citation comment payload refuses cleanly. Clean pass.
- **Seam 4 (python relift to bind, OPEN GAP SUSPECTED):**
  - Positive: clean pass (BindKit accepts python-lifted Term shape).
[...]

PR #1072, opened 2026-05-16
docs(libprovekit/walks): lock lenient verdict-propagation policy with antibody doc-comment + regression test (closes #1070)

TSavo:
Closes #1070. A8 prereq for #1068 Path B. Locks the architect ruling on the verdict-propagation question (census seam 7).

## Architect ruling locked

The verdict-propagation question (raised in `docs/plans/2026-05-16-trinity-composition-test-census.md` seam 7) resolves to **option β: lenient verdict policy**.

`walk_premises_to_root` verifies structural chain integrity (cycle, premise CID resolution, origin reachable, deserialization). It does NOT check verdict on intermediate claims. The shipped Path A code (#1067) already implements β. A8 is documentation + one regression test that catches future α-shape drift (where intermediate kits would falsely claim Proved).

The locked architect reasoning: chain-walk proves chain integrity, not semantic correctness of intermediate transforms. Intermediate kits do not claim Verdict::Proved because they did not prove anything; they transformed. ProveKit's terminal claim IS Proved because its prove() ran the chain-walk and it succeeded.

## What changed

- **`walks.rs` ChainBreak doc-comment:** antibody note: "do not add verdict-shaped variants" (e.g., `NotProved`). Verdict-aware verifiers belong in their own kits, not in this enum.
- **`walks.rs` walk_premises_to_root doc-comment:** locked architect language verbatim. Explicitly enumerates the four failure modes and states the lenient policy: "Verdict semantics on intermediate claims are out of scope for this walk."
- **`walks.rs` regression test** `walk_premises_returns_ok_when_intermediate_claims_have_non_proved_verdicts`: 3-step chain with `Verdict::Pending` intermediate claim; assert walk returns Ok. If anyone adds an α-shape verdict check, this test fails red and points back to A8.

40 insertions, all in walks.rs. No algorithm changes. No new variants. No ProveKit changes.

## Test partition discipline

This is the test-partition-is-spec lesson from A2's discriminator bug applied to architectural rulings. The architect's words become a test that fails when the ruling is violated. Future contributors inherit the discipline without needing to re-read the architect's chat.

## Gates

- `cargo test -p libprovekit`: 70 passed (up from 69, with the new regression test).
- `cargo test -p libprovekit -p provekit-cli`: green.
- `tools/spec-cid-lint.sh`: exit 0.
- `git diff --check`: clean.
- em-dash / en-dash sweep on diff: zero.

## Build-on-existing-kits clause (held)

[...]

PR #1071, opened 2026-05-16
feat(libprovekit/lower): claim_spec_value decomposes Term::Op { concept:bind-result } wrapper (closes #1069)

TSavo:
Closes #1069. A7 prereq for #1068 Path B. Closes the missed-integration gap A3 (#1065) introduced and surfaced by Path B's empirical probe.

## The gap

A3 changed BindKit's output to `Term::Op { op_cid: concept:bind-result, args: [original_term, named_form_binding] }`. `LowerKit::claim_spec_value` (`lower_plugin.rs:210-228` on main pre-A7) only fast-paths `Term::Const`; the new wrapper fell through to the default branch and synthesized a RealizeRequest with `function = bind::default::bind-result-op-tree`. Python realize plugin refused correctly: `missing body-template entry`.

Path B's empirical probe (branch `1068-trinity-exhibit-path-b`, SHA f2106ce5) confirmed: lift + bind composed cleanly, BindKit determinism passed, lower refused.

## What changed

- **`lower_plugin.rs:211`:** new match arm in `claim_spec_value` descends through `Term::Op { op_cid: concept_bind_result_cid(), args }` and calls into the shared spec builder. Term::Const fast path preserved. Other Term::Op variants continue taking the existing default fallback.
- **`cmd_lower.rs:281`:** CLI named-term lowering now uses the same shared spec builder so both code paths produce byte-identical RealizeRequest shapes. Single source of truth.
- **`tests/lower_claim_bind_result.rs`** (new, 174 LOC): five-partition test set.

## Test partition (the test-partition-is-spec discipline locked post-A2)

1. **Positive:** bind to lower with the wrapper succeeds; spec shape matches cmd_lower output for the same NamedTerm content.
2. **Regression:** Term::Const fast path still works.
3. **Discrimination:** a Term::Op with a DIFFERENT op_cid takes the existing fallback, not the new descent. **This is the load-bearing discrimination test A2's spec was missing.** If anyone tries to over-broaden the match arm, this test fails.
4. **Refusal:** malformed wrapper (wrong arg arity) refuses cleanly with a typed error. No panic. No silent fallthrough.
5. **Cross-target:** synthesized RealizeRequest is target-neutral; Python / Java / C plugins all receive the same shape.

## Gates

- `cargo test -p libprovekit -p provekit-cli`: green (every test partition passes).
- `tools/spec-cid-lint.sh`: exit 0.
- `git diff --check`: clean.
- em-dash / en-dash sweep on diff: zero.

[...]

PR #1078, opened 2026-05-16
feat(lifters): strip type annotations at lift boundary; realize kits reconstruct types (closes #1075)

### LINKED ISSUE #1075: A9: bind layer preserves language-tier type info that should project away for cross-language federation
**Title:** A9: bind layer preserves language-tier type info that should project away for cross-language federation

**Labels:** substrate-correctness, area:substrate-spine, area:realize-side, follow-up, priority:p1

**Parent:** #1024 Trinity exhibit. **Blocks:** #1068 Path B (assertion 4 federation byte-identity). **Surfaced by:** #1074 / #1073 composition test census, seam 4 federation test.

---

## The empirical finding

The census's seam-4 federation test (`provekit-cli/tests/trinity_composition_census.rs`) ran:

1. Take Rust source `fn add(x: i64, y: i64) -> i64 { x + y }`.
2. Lift via Rust LiftKit (`provekit-walk-rpc`). Bind via BindKit. Capture `bind_from_rust_cid`.
3. Lower the bind output to Python source via LowerKit. Relift via Python LiftKit (`provekit-lift-python-source`). Bind via BindKit. Capture `bind_from_python_cid`.
4. Assert `bind_from_rust_cid == bind_from_python_cid`.

The assertion fails. The two bind CIDs diverge. The failure is currently captured as a `#[should_panic]` antibody test in #1074 referencing this issue.

## NamedTerm diff captured

Rust-lift produces a NamedTerm with:
- `param_types = [i64, i64]`
- `return_type = i64`
- `term_shape` and `named_term_tree` carry Rust-specific structural encoding

Python-lift produces a NamedTerm with:
- `param_types = [Any, Any]`
- `return_type = Any`
- `term_shape` and `named_term_tree` carry Python-specific structural encoding

Both Terms encode the same algebra (`add(x, y)` over i64-or-Any), but the bind layer preserves the language-tier type information rather than projecting it away. Two language-tier views of the same algebra produce different concept-tier CIDs.

## Why this matters

The substrate's federation property requires that the concept tier is the canonical address for an algebra, independent of the lifting language. If `bind_from_rust(add(x:i64, y:i64) -> i64).cid() != bind_from_python(add(x:Any, y:Any) -> Any).cid()` for the same algebra, then:

- The Trinity exhibit's assertion 4 (`post_bind.cid() == post_rebind.cid()`) cannot pass empirically. The exhibit reduces to "the round-trip is deterministic within one language" rather than "the algebra is the portable thing."
- The federation claim of #1024 fails at the concept tier, not at the language tier where it was supposed to fail.

This is the federation gap suspected by the census doc for seam 4.

## Architectural question

Where should the type-erasure projection happen? Three places it COULD happen:

- **(α) Inside BindKit:** project away `param_types` and `return_type` (or canonicalize them to a concept-tier representation) before producing the `Term::Op { concept:bind-result, args }` wrapper. Bind output stays language-typed in the args, but the canonical CID is computed over a type-erased view.
- **(β) Inside LiftKit (per-language):** Rust LiftKit and Python LiftKit both project to a canonicalized type representation at lift time. Their outputs are then byte-identical for equivalent algebras. Distributes the responsibility across every LiftKit.
- **(γ) New canonicalization step between bind and CID computation:** a `concept:canonicalize` op that runs after bind, normalizes the type representation, and produces the addressable concept-tier CID from the canonicalized form.

Architect ruling required before A9 dispatches.

## Test partition (when A9 dispatches)

The fix must include the test-partition-is-spec discipline applied to this seam:

- **Positive (federation property holds):** bind_from_rust_cid == bind_from_python_cid for at least three distinct algebras (arithmetic, conditional, recursive).
- **Discrimination:** two semantically-distinct algebras produce different bind CIDs (A10's concern, but the discrimination test belongs in A9's spec too).
- **Refusal:** if the type-erasure projection cannot proceed (e.g., the LiftKit produced an unrepresentable type), refuse cleanly with a typed error.

The composition test census #1074 contains the antibody (`#[should_panic]`). When A9 closes, the antibody promotes to a passing assertion via a small follow-up commit removing the `#[should_panic]` attribute.

## Build-on-existing-kits clause

Whatever option is chosen, the fix must hold:
- No new crate.
- No new public trait or trait method on Kit / LiftKit / BindKit / Catalog.
- No change to the seven-step algebra.
- No silent fallthrough; refusals are typed.

## Reference

- `docs/plans/2026-05-16-trinity-composition-test-census.md` seam 4.
- #1074 / #1073 census PR with the empirical `#[should_panic]` antibody at `provekit-cli/tests/trinity_composition_census.rs` (federation test).
- Path B determinism sentinel passed empirically on Path B's earlier probe (BindKit is invocation-deterministic on identical input); A9 is the cross-language analogue.

TSavo:
Closes #1075. A9. Implements architect-locked option (a) for the verdict-propagation question (census #1074 seam-4 federation gap).

## Architect ruling (locked, from issue #1075 comment 4467318644)

**Option (a):** lifters strip type annotations at the lift boundary. The hub carries algebra shape only (untyped). Realize kits reconstruct types from concept-tier shape + target language type system defaults.

Architect reasoning verbatim: *"The substrate's 'algebra is the portable thing' claim is structurally about shape, not types. Paper 16's colimit argument is structural; types are not in the universal property."*

The type-system asymmetry between languages belongs IN the realize kit, where language-specific knowledge already lives. The hub stays universal.

## What changed

- **Python source bind lifter** (`bind_lifter.py`): emits `param_names` only; no `param_types` or `return_type` fields.
- **Rust `provekit-walk-rpc` bind lift** (`walk_rpc.rs`): omits `param_types` and `return_type` from emitted NamedTerm entries, including bound callsite entries. Bind lift no longer forwards type-signature evidence as bind witnesses.
- **LowerKit** (`lower_plugin.rs`): reconstructs erased signature defaults for realize requests. Neutral `int` params/return as Rust convention; preserves `unit` for `unit` concepts.
- Regression coverage added for both lifters and lower reconstruction (test_bind_lifter.py + lower_claim_bind_result.rs).

## Build-on-existing-kits clause

No new Kit trait method. No bind-IR schema change. No new concept op-CID. No algebra change. No silent fallthrough; refusals are typed.

## Gates

- `pytest implementations/python/provekit-lift-python-source/tests`: 29 passed.
- `pytest implementations/python/provekit-realize-python-core/tests`: 24 passed.
- `cargo test -p provekit-walk --bin provekit-walk-rpc`: 6 passed.
- `cargo test -p libprovekit -p provekit-cli`: green.
- `tools/spec-cid-lint.sh`: exit 0.
- `git diff --check`: clean.
- em-dash / en-dash sweep: zero.

## Census antibody status after this PR + A10

After A9 (this PR) plus A10 #1076 both land, the seam-4 federation antibody can flip from `#[should_panic]` to `assert_eq` via a small follow-up commit. Both fixes are needed simultaneously:

- **A9 erases types** so bind CIDs can match across languages (this PR).
- **A10 preserves operator atoms** so distinct operators don't collide within either language (#1076).

If only A9 lands without A10: the federation antibody still fails because `x+y` and `x-y` collide in Python lift, so equating Rust's `add` and Python's "ambiguous binary op" is meaningless.

If only A10 lands without A9: the federation antibody still fails because Rust types persist and Python types don't.

## Sequencing

Touches the Python lifter package at a different code path than A10 #1076 (type-emission vs operator-atoms). File-path collision on `bind_lifter.py` and `test_bind_lifter.py`: whichever PR merges second rebases. Independent semantically.

## Cited prereqs

A1 #1064. A2 #1066. A3 #1065. A4 #1061. A5 #1062. A6 #1063. Path A #1067. A7 #1071. A8 #1072. Census #1074.

## Summary by CodeRabbit

* **Refactor**
  * Modified bind-lift output to exclude signature type information from function entries while preserving parameter names.
  * Updated the system to reconstruct default types (`int` for parameters, `()` or `int` for return values) when signature data is unavailable, ensuring downstream compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1078?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/implementations/python/provekit-lift-python-source/src/provekit_lift_python_source/bind_lifter.py
+++ b/implementations/python/provekit-lift-python-source/src/provekit_lift_python_source/bind_lifter.py
@@ -161,12 +161,7 @@ def _entry_for_function(
 ) -> Json:
     concept, attr_pre, attr_post = _extract_leading_annotations(lines, node.lineno)
     term_shape = _function_shape(node)
-    param_names, param_types = _signature_params(node.args)
-    return_type = _annotation_text(node.returns)
-    if return_type is None:
-        return_type = "Any"
-    elif return_type == "None":
-        return_type = "()"
+    param_names = _signature_param_names(node.args)
     witnesses = []
     witnesses.extend(_contract_comment_witnesses(lines, node, rel_path, diagnostics))
     witnesses.extend(_decorator_contract_witnesses(node, param_names, rel_path, diagnostics))
@@ -182,17 +177,14 @@ def _entry_for_function(
         "fn_name": node.name,
         "kind": "bind-lift-entry",
         "param_names": param_names,
-        "param_types": param_types,
-        "return_type": return_type,
         "term_shape": term_shape,
         "term_shape_cid": cid_of_json(term_shape),
         "witnesses": witnesses,
     }
 
 
-def _signature_params(args: ast.arguments) -> tuple[list[str], list[str]]:
+def _signature_param_names(args: ast.arguments) -> list[str]:
     names: list[str] = []
-    types: list[str] = []
     ordered_args: list[ast.arg] = []
     ordered_args.extend(args.posonlyargs)
     ordered_args.extend(args.args)
@@ -203,17 +195,7 @@ def _signature_params(args: ast.arguments) -> tuple[list[str], list[str]]:
         ordered_args.append(args.kwarg)
     for arg in ordered_args:
         names.append(arg.arg)
-        types.append(_annotation_text(arg.annotation) or "Any")
-    return names, types
-
-
-def _annotation_text(annotation: ast.expr | None) -> str | None:
-    if annotation is None:
-        return None
-    try:
-        return ast.unparse(annotation)
-    except Exception:
-        return "Any"
+    return names
 
 
 def _function_shape(node: ast.FunctionDef | ast.AsyncFunctionDef) -> Json:
--- a/implementations/python/provekit-lift-python-source/tests/test_bind_lifter.py
+++ b/implementations/python/provekit-lift-python-source/tests/test_bind_lifter.py
@@ -162,6 +162,26 @@ def _operator_atoms(term_shape: dict) -> list[dict]:
     return [node for node in _walk_objects(term_shape) if "op_cid" in node]
 
 
+def test_bind_lift_erases_signature_types_from_bind_ir_entries() -> None:
+    source = (
+        "def add(left: int, right: int) -> int:\n"
+        "    return left + right\n"
+        "\n"
+        "def generated(value):\n"
+        "    return value\n"
+    )
+
+    result = lift_source(source, "pkg/foo.py")
+
+    assert result.diagnostics == []
+    entries = {entry["fn_name"]: entry for entry in result.ir}
+    assert entries["add"]["param_names"] == ["left", "right"]
+    assert "param_types" not in entries["add"]
+    assert "return_type" not in entries["add"]
+    assert "param_types" not in entries["generated"]
+    assert "return_type" not in entries["generated"]
+
+
 def test_bind_lift_source_emits_language_neutral_entries() -> None:
     source = (
         "# concept: identity\n"
@@ -202,8 +222,8 @@ def test_bind_lift_source_emits_language_neutral_entries() -> None:
     assert result.ir[0]["attr_pre"] == "x >= 0"
     assert result.ir[0]["attr_post"] == "result >= 0"
     assert result.ir[0]["param_names"] == ["x"]
-    assert result.ir[0]["param_types"] == ["int"]
-    assert result.ir[0]["return_type"] == "int"
+    assert "param_types" not in result.ir[0]
+    assert "return_type" not in result.ir[0]
     assert result.ir[0]["term_shape"] == {
         "kind": "body",
         "stmts": [{"kind": "exit"}],
@@ -324,15 +344,15 @@ def test_bind_lift_filters_unnamed_concepts_and_void_return() -> None:
 
     assert [entry["concept_annotation"] for entry in result.ir] == [None, None]
     assert result.ir[0]["param_names"] == ["x"]
-    assert result.ir[0]["param_types"] == ["Any"]
-    assert result.ir[0]["return_type"] == "Any"
+    assert "param_types" not in result.ir[0]
+    assert "return_type" not in result.ir[0]
     assign_shape = result.ir[0]["term_shape"]["stmts"][0]
     assert assign_shape["kind"] == "assign"
     assert {
         "concept_name": "concept:add",
         "op_cid": _catalog_concept_cid("concept:add"),
     }.items() <= assign_shape["value"].items()
-    assert result.ir[1]["return_type"] == "()"
+    assert "return_type" not in result.ir[1]
 
 
 def test_bind_rpc_initialize_declares_bind_ir_surface() -> None:
--- a/implementations/rust/libprovekit/src/core/lower_plugin.rs
+++ b/implementations/rust/libprovekit/src/core/lower_plugin.rs
@@ -263,16 +263,41 @@ pub fn realize_spec_from_named_term(term: &NamedTerm) -> Result<Value, String> {
         .map(serde_json::to_value)
         .transpose()
         .map_err(|error| format!("serialize namedTermTree for `{}`: {error}", term.function))?;
+    let (param_types, return_type) = realize_signature_from_named_term(term);
     Ok(json!({
         "kind": "RealizeRequest",
         "function": term.function,
         "params": term.params,
-        "paramTypes": term.param_types,
-        "returnType": term.return_type,
+        "paramTypes": param_types,
+        "returnType": return_type,
         "conceptName": term.concept_name,
         "namedTermTree": named_term_tree,
         "termShapeCid": term.term_shape_cid,
     }))
+}
+
+fn realize_signature_from_named_term(term: &NamedTerm) -> (Vec<String>, String) {
+    let erased = term.param_types.is_empty() && matches!(term.return_type.trim(), "" | "()");
+    if !erased {
+        return (term.param_types.clone(), term.return_type.clone());
+    }
+
+    let param_types = term
+        .params
+        .iter()
+        .map(|_| "int".to_string())
+        .collect::<Vec<_>>();
+    let return_type = if is_unit_concept(&term.concept_name) {
+        "()".to_string()
+    } else {
+        "int".to_string()
+    };
+    (param_types, return_type)
+}
+
+fn is_unit_concept(concept_name: &str) -> bool {
+    let trimmed = concept_name.trim();
+    trimmed == "unit" || trimmed == "concept:unit"
 }
 
 fn fallback_from(input: &Input) -> Vec<Cid> {
--- a/implementations/rust/libprovekit/tests/lower_claim_bind_result.rs
+++ b/implementations/rust/libprovekit/tests/lower_claim_bind_result.rs
@@ -105,6 +105,29 @@ fn bind_input_value() -> Value {
     })
 }
 
+fn erased_bind_input_value() -> Value {
+    json!({
+        "kind": "ir-document",
+        "workspaceRoot": "/tmp/provekit-lower-claim-bind-result-test",
+        "ir": [{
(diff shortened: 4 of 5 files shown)


TSavo:
## Verdict: MERGE-WITH-NITS

Kit-as-Opus-fallback review (server rate-limited the parallel review agents). All 10 required checks pass at the substrate-correctness level; one architectural deviation from the locked spec is flagged as a non-blocking nit.

### Checks passing

1. **Architect ruling (a) implemented:** lifters strip type annotations (bind_lifter.py:5-28, walk_rpc.rs:268-279); hub carries algebra shape only; realize-side reconstructs at lower_plugin.rs:136-153.

2. **Lift-side stripping:** both Python and Rust lifters omit `param_types` / `return_type` from emitted JSON entirely (not null-sentinel'd). Rust lifter also stops forwarding `lift_file_with_sig_evidence` as bind witnesses (walk_rpc.rs:303-307) — type-signature evidence no longer crosses the bind boundary.

3. **Realize-side reconstruction is real, not a stub:** `realize_signature_from_named_term` detects erased case (`param_types.is_empty() && return_type in {"", "()"}`) and reconstructs `vec!["int"; params.len()]` for params, `"int"` for return (or `"()"` for unit concepts). Non-erased inputs pass through unchanged for backward compatibility.

4. **Regression coverage solid:** four new tests verify the stripping (Python + Rust lifter side) AND the reconstruction (LowerKit side). `bind_result_claim_lower_reconstructs_erased_signature_defaults` is the load-bearing end-to-end test.

5. **No new substrate API:** no new Kit trait methods, no JCS schema breaking change (fields omitted, NamedTerm struct unchanged at Rust type level; serde tolerates missing fields as defaults), no new concept op-CIDs, no algebra change.

6. **A10 dependency acknowledged honestly:** PR body explicitly says A9 alone doesn't flip the federation antibody — A10's operator-atom collision is upstream. Diff does NOT touch trinity_composition_census.rs.

7. **Build-on-existing-kits clause:** held. No new crates.

8. **Em-dash / en-dash sweep:** zero.

9. **No --no-verify.**

### NIT (non-blocking architectural deviation)

The architect's locked spec said: *"Each realize plugin owns its language's type reconstruction logic. Python realize uses `Any` or whatever Python idiom; Rust realize uses `i64` for integer-shaped concept ops."*

This PR puts the reconstruction logic in **LowerKit** (`lower_plugin.rs:136-153`) with a single target-neutral default (`"int"`). Realize plugins consume whatever LowerKit hands them; they don't get to express language-idiomatic preferences (Python: `Any` over `int`; Java: `Integer` vs `int`; etc.).

Why this is a NIT not a BLOCK:
- **Federation invariant still holds** via A9 + A10. The federation check is at bind-tier CIDs, which don't depend on where reconstruction happens downstream of bind.
- All tests pass. End-to-end composition works with the current architecture.
- Build-on-existing-kits favors smaller diffs; threading per-language reconstruction policy into realize-plugin RPC interfaces is a refactor, not a substrate-correctness fix.

**Follow-up issue suggested** (after A9 merges): "Push type-reconstruction logic from LowerKit into realize plugins so each language can express idiomatic type defaults." The current architecture allows this evolution without breaking A9's federation contract — realize plugins could start ignoring LowerKit's defaults and reconstruct per their own language convention.

The A2-lesson (test-partition-is-spec) is relevant: A9's spec tests federation byte-identity and discrimination preservation, but does NOT include "Python realize chooses Python-appropriate type defaults" as a discrimination test. A future PR with that discrimination test would force the refactor.

### Verdict justification

The substrate correctness deliverable (federation antibody flips when A9+A10 both land) is empirically achieved. The architectural deviation is real but doesn't break correctness; it's a future-friction risk worth a follow-up issue, not a merge block.

TSavo:
## Verdict (independent Opus retry): MERGE

Independent review, fresh Opus agent, no dispatch-brief context contamination. I read the diff, the architect's locked ruling at #1075 comment 4467318644, and the actual realize-plugin code in `provekit-realize-rust-core` and `provekit-realize-python-core`. **I disagree with Kit's prior MERGE-WITH-NITS reading**: the architectural deviation Kit flagged is not actually a deviation once you trace the full per-language emission path. Kit stopped reading one level too shallow.

### What Kit missed

Kit cited `lower_plugin.rs:136-153` reconstructing as target-neutral `"int"` and concluded each realize plugin doesn't get to express language-idiomatic preferences. Kit didn't check what the realize plugins DO with that string.

They translate it. Each plugin owns a `map_source_type` function that is the language-idiomatic reconstruction layer:

- `implementations/rust/provekit-realize-rust-core/src/lib.rs:614-628` — `map_source_type("int") -> "i64"`. **Exactly the value the architect specified verbatim.**
- `implementations/python/provekit-realize-python-core/src/provekit_realize_python_core/realizer.py:1591-1611` — `map_source_type("int") -> "int"` (Python's native integer type; the architect said `Any` "or whatever Python idiom" — `int` IS a Python idiom).
- The translation is applied at `realizer.py:150` (`args_shape=tuple(map_source_type(ty) for ty in param_types)`) and `lib.rs:402-403` (`mapped_param_types: Vec<String> = param_types.iter().map(|ty| map_source_type(ty)).collect();`).

A future Java realize plugin would extend its own `map_source_type` to map `"int" -> "int"` or `"Integer"`. A C plugin would map to `"int"` or `"int64_t"`. The substrate is exactly the federation shape the architect described: target-neutral canonical-token reconstruction at LowerKit, language-idiomatic translation at each realize plugin.

### Reconciling with the architect's literal language

The architect wrote: "Each realize plugin owns its language's type reconstruction logic. Python realize uses `Any` or whatever Python idiom; Rust realize uses `i64`."

What the PR does:
- LowerKit reconstructs a canonical erasure-default token (`"int"`/`"()"`) — target-neutral.
- Each realize plugin translates that canonical token to its language idiom via `map_source_type`.
- Result: Rust source contains `i64`, Python source contains `int`. The architect's named output values are produced.

This is a two-step reconstruction with a canonical pivot token between LowerKit and realize. The architect specified the post-reconstruction values (`i64`, `Any`/`int`); the PR produces exactly those values per language. The architectural property the architect's wording was protecting — "language-specific knowledge lives in realize, not at the hub" — holds: `map_source_type` is in each per-language realize crate, owns its own translation table, and is independently evolvable per language. Adding a Java plugin requires only that plugin's `map_source_type`; it does not touch LowerKit.

### Federation invariant + substrate-correctness both hold

- **Federation byte-identity (Kit was right here):** preserved at bind-tier CIDs because no types cross bind.
- **Substrate correctness of emitted output (advisor's discriminator):** preserved because realize plugins translate `"int"` to valid target-language types (Rust gets `i64`, not the invalid bare `int`). This is the Case-B branch the advisor flagged as the test for NIT-vs-BLOCK.

### Other required checks (all clean)

- No new Kit trait method: confirmed (diff has no trait modifications).
- No bind-IR JCS schema breaking change: fields omitted on emission; serde tolerates missing fields as defaults; NamedTerm Rust struct unchanged.
- No new concept op-CIDs: confirmed.
- No algorithm change beyond lift/realize: confirmed.
- Em-dash/en-dash sweep: 0 hits in the diff.
- No `--no-verify`: 0 hits in the diff.
- Regression tests cover stripping (Python `test_bind_lifter.py` + Rust `lower_claim_bind_result.rs`) AND reconstruction (`bind_result_claim_lower_reconstructs_erased_signature_defaults` asserts the RealizeRequest contains `"int"` / `"()"`).

### One legitimate gap (not blocking)

The new test asserts the RealizeRequest payload at the LowerKit boundary contains `"int"`. It does NOT assert the final emitted Rust source contains `i64` (i.e., it doesn't exercise `map_source_type` end-to-end on the erased path). The existing non-erased test at line 89-90 / 225-226 covers `i64 -> i64` round-trip, so map_source_type is covered, just not the composition of (LowerKit-default-reconstruction → realize-plugin-translation). This is a follow-up test-coverage nit, not a substrate-correctness gap — the translation logic is already exercised by the non-erased path.

### Judgment on the substrate-correctness vs future-friction axis

This is neither a substrate-correctness gap nor a future-friction nit. The architect's spec is **already satisfied** by the two-layer reconstruction protocol (canonical-token + per-plugin map). Kit's NIT framing assumed the absence of per-plugin translation; that assumption is wrong on the code. The federation property holds, language idioms are emitted correctly, each realize plugin owns its language translation. **MERGE.**

— Independent Opus 4.7 retry, no dispatch-brief context.

Issue #1355, opened 2026-05-21
Epic: floating-axes materialize — family / library / version / language / concept


TSavo:
## Parent

Extends the unified `SiteTransformKit` machinery shipped in #1334 (materialize + migrate convergence) by promoting the dispatch key from `(target_lang, library_tag, concept)` to the full five-axis tuple `(language, family, library, version, concept)`. Each axis is pinnable on the @boundary annotation or floatable; floating axes resolve against the consumer's platform profile at materialize time.

## Canonical example

`family = sql`, `library = rusqlite`, `version = 0.x.y`, `language = rust`, `concept = concept:sql-query`

Concepts are FAMILY-LEVEL (`concept:sql-query`, not `concept:rusqlite-execute`). Library / version / language are realization axes below the concept.

## Why

Today's dispatch is one-dimensional string equality on `library_tag`. The substrate already has the trichotomy and loss-composition machinery to do constraint-satisfaction resolution; only the dispatch key is narrow. Promoting unlocks:

- `(family, concept)` binding with everything else floating = type-class-style polymorphic call site. Deployment-time library choice; test/prod parity by profile swap; cross-language federation from the same source.
- Migrate becomes materialize against a different profile (kills the `TargetSurface` hardcoded enum in `cmd_bind_migrate.rs`).
- Version upgrades fall out automatically when newer libraries still satisfy the same family/concept CIDs.
- Cross-language port (rust→python) becomes a two-profile materialize, not a new command.

## What's PINNED today vs needs adding

| axis | today | gap |
|------|-------|-----|
| `concept` | `concept_name` string in carrier | catalog-level lookup w/ family-parent traversal |
| `library` | `library_tag` string equality | hierarchical: family contains libraries |
| `language` | `--target` flag | move into platform_profile |
| `family` | absent | first-class field on each concept catalog entry |
| `version` | embedded in `workspace_override` path | promote to first-class field on @sugar/@boundary + library-sugar-binding-entry |

`dialect` is intentionally NOT an axis. SQL strings are opaque data to the substrate; cross-dialect translation is out of substrate scope. Provenance can live in receipt mementos if anyone wants it.

## Sub-issues (modify-in-place, no new crates)

- [ ] #1356 [A] Family field in concept catalog + walk_rpc op-cid loader
- [ ] #1357 [B] Version + library + family fields on @sugar/@boundary annotations
- [ ] #1358 [C] platform_profile section in shim .provekit/config.toml
[...]

PR #1365, opened 2026-05-21
feat(materialize): cross-language --out-dir emits target-language files

TSavo:
## Summary

Extends cross-language materialize with **actual target-language file emission**. When `--out-dir` is passed alongside `--source-lang` and `--target`, RESOLVE'd boundaries are written into composite target-language files (one per source file).

This sits on top of the floating-axes epic (#1355, closed) and the in-flight cross-language wire that landed in 2c33c1150 (showed body previews).

## Substrate invariant preserved

`cmd_materialize` does **zero** target-syntax translation. It routes concept-hub-typed carrier payloads to the target kit's realize binary, and the realize binary returns target-language source verbatim. The kit-internal sort labels (`rust:`, `java:`, `python:`, etc.) never appear in materialize.

## End-to-end demo

```
$ provekit materialize --source-lang rust --target python \
    --library sqlite3 \
    --source-dir examples/cross-lang-sql-family-demo/src \
    --project . \
    --out-dir /tmp/xlang-out

RESOLVE concept:sql-connection-open → python manifest `sqlite3`
    target body preview: def run_open(path):
        return sqlite3.connect(path)
EMIT wrote /tmp/xlang-out/lib.py (1 body bodies, 0 imports)
materialize cross-language emission: 1 file(s) written to /tmp/xlang-out
```

Output `/tmp/xlang-out/lib.py`:
```python
def run_open(path):
    return sqlite3.connect(path)
```

## What's in the diff

`implementations/rust/provekit-cli/src/cmd_materialize.rs` (+120 / -6):
- `run_cross_language_discovery` accepts `out_dir: Option<&Path>`.
- New `EmittedFile` struct accumulates per-source-file output (bodies + dedup'd imports from each used library's realize manifest `scope_bringings`).
- `target_lang_file_extension` maps target language → file extension (rs/py/ts/java/c/cs/go/php/rb/zig). Heuristic for now — a kit could declare its extension in its manifest as a follow-up.
- After the discovery loop, writes `<out-dir>/<source-stem>.<ext>` with imports block + bodies concatenated.

## Test plan

- [x] Build clean (`cargo build -p provekit-cli` — finished in 11.96s)
- [x] End-to-end emission verified (rust → python writes `/tmp/xlang-out/lib.py`)
- [x] Same-language regression: 9 exact + 0 lossy + 0 refused unchanged
- [ ] AMBIGUOUS-case still doesn't emit (currently caller picks via `--library`)
- [ ] Empty imports block when no scope_bringings declared (works; would benefit from a regression test)

## Follow-up not in this PR

- Per-non-rust source kit lift to concept-hub CIDs (mirrors rust kit's 840c2c33f).
- AMBIGUOUS resolution policy (prefer-first / fail-fast / require-pin).
- Full target-file emission including signature decoration where realize binaries return body-only (rust target).

## Summary by CodeRabbit

* **New Features**
  * Added cross-language discovery with optional composite file generation and per-family library disambiguation
  * Introduced Java implementations for Blake3 hashing, JSON processing (Gson/Jackson), RFC8785-JCS canonicalization, and stdio operations
  * Added concept shape definitions and multi-language sort morphisms for enhanced type safety

* **Documentation**
  * Added extensive body templates for libraries across multiple languages (Blake3, JSON, stdio, database)
  * Created per-language sort catalog entries and transformation definitions

* **Improvements**
  * Enhanced cross-language type handling with substrate-canonical sort identifiers
  * Improved template matching with library tag disambiguation
  * Extended structural operator support in term-shape lowering

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1365?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TSavo:
@coderabbitai @chatgpt-codex-connector — addressed all P1/Major/Minor findings in `76404d375`. Please re-review.

## P1 / Major (correctness)

1. **Out-dir file collision** (Codex P1, CodeRabbit Major) — `cmd_materialize.rs` now preserves source-relative path under `--out-dir` and creates intermediate directories.

2. **Body templates keyed by library** — added `target_library_tag` to `RealizeRequest`, plumbed from `kit_dispatch` via the resolved manifest; java realize threads it via `ThreadLocal`; `BodyTemplateEntry` carries `targetLibraryTag`; matcher filters by it.

3. **Refusal-guard bypass** — replaced empties-check with explicit `isCrossLang` flag detected via `JsonUtil.hasField()` on RPC params (field-present = cross-lang signal). Zero-param cross-lang calls now distinguishable from legacy.

4. **Mojibake in JCS Java template** — `â` (em-dash UTF-8-as-Latin-1) replaced with comma.

5. **RFC 8785 number canonicalization (substrate-honesty self-violation)** — declared loss as `rfc8785-number-serialization-non-ecma262` on `encode_jcs` + `encode_value` in both rust + java shim sources AND in the three body-template JSONs.

6. **Java FQN handling** — `javaTypeToConceptHubSortCid` now reduces to simple class name via last-dot (was only stripping `java.lang.`/`java.util.`).

7. **Ref<T> filename** — renamed `Ref<T>.<cid>.json` → `Ref_of_T.<cid>.json`. Mint script sanitizes filenames going forward.

8. **Java for-loop multi-init/update** — multi-clause init/update lists wrap as `concept:seq` so all side effects survive.

9. **Java try-lift preserves catch/finally** — minted `concept:catch-arm` + `concept:finally-arm`. `concept:try` now carries `[body, catch-arm(type, binding, body)..., finally-arm?(body)]`.

10. **Java method-call shape parity** — `obj.method(x)` now emits `call(receiver, method:"method", x)` matching rust's shape. Cross-language clustering for instance methods restored.

11. **Operand_bindings threaded through structural control flow** — `If/While/ForLoop/Loop/Match/Cast/Index/Field` thread bindings through the operator's argument layout matching `shape_of_expr`.

12. **Free `Expr::Path` symbol-leaf dead-code fixed** — `operand_symbol`-derived bindings only emitted for identifiers in `ctx.vars`. Free identifiers (None, Some, etc.) fall through to the `kind:"symbol"` leaf as intended.

## Minor

13. `body bodies` typo fixed.
14. `bcprov-jdk18on` 1.78.1 → 1.81.1 (GHSA-c3fc-8qff-9hwx, GHSA-p93r-85wp-75v3).
15. `mint_while_foreach_try.py` now writes to `catalog/algorithms/` with filesystem-safe filenames.
16. `mint_unit_ref_json_morphisms.py` dup check now keys on `(lang, lang_sort, concept_sort)` — multiple realizations per concept (Jackson vs Gson) no longer silently dedupe.

## Bonus pre-existing bug surfaced + fixed

`cmd_mint::project_body_templates_for_sugar_bindings` never wrote `header.cid` even though `cmd_bind_migrate` required it. Computation + backfill into all 14 existing body-template files.

## Verification

- `cargo test --workspace`: **760 pass**
- 3 pre-existing `cross_platform_point_query_receipt_test` fixture failures (fixture_4/5/6) confirmed pre-existing on `0af74301c` (this PR's starting commit) — they need an arity-2 sql-query body-template variant which has never existed at any commit. Out of scope for this PR.
- Same-lang materialize: 9/9 exact
- Cross-lang materialize (rust → java): 9/9 resolve, 0 ambiguous, 0 refused
- Cycle invariance (rust ↔ java concept-hub identity): 9/9 preserved

TSavo:
@coderabbitai @chatgpt-codex-connector — second round of follow-ups addressed in \`ba79b654b\`. Please re-review.

## Findings closed

**P1 / Major (correctness):**

1. **\`JsonUtil.hasField\` structural depth-1 key check** — substring scan replaced with a JSON walker that respects quoting + nesting and confirms `:` follows the matched string (key, not value).

2. **Body-template two-pass selection** — \`SugarRealizer\` extracted \`tryEntries()\` helper. Pass 1 (only when \`currentLib\` non-empty) exact-match; pass 2 library-agnostic only. Entries from OTHER libraries never considered.

3. **\`scoped_names\` in \`ShapeContext\`** — \`BTreeSet<String>\` populated on every \`set_local\` regardless of sort inference. \`bindings_of_expr\` now keys on it instead of \`vars\`. Params/locals with non-modeled types (refs, String, custom) now correctly recognized as scoped.

4. **\`resolveJavaType\` refuses on cross-lang gap** — added \`isCrossLang\` flag. When non-empty CID can't resolve, returns \`null\`; \`emitStubInner\` refuses with named CID in the reason.

5. **\`cmd_materialize\` stamps resolved \`target_library_tag\` before dispatch** — \`request.target_library_tag = resolved-target-tag\` set on the request mutation BEFORE \`transport.dispatch_realize\`.

6. **\`lower_plugin\` child uses \`effective_library_tag\`** — resolution moved before request construction; \`target_library_tag\` uses child-level override, falling back to parent only when no override exists.

7. **\`DiscoveryOutcome\` enum** — \`invoke_target_realize_for_discovery\` returns \`Preview\` / \`SemanticGap\` / \`TransportError\`. Both call sites discriminate: \`Preview\` → RESOLVE, \`SemanticGap\` → REFUSE, \`TransportError\` → WARN (not counted as refuse). Broken manifests/binaries no longer masquerade as substrate gaps.

## Verification

- Same-lang materialize: 9/9 exact
- Cross-lang materialize (rust → java): 9/9 resolve
- Cycle invariance (rust ↔ java): 9/9 preserved
- \`cargo test --workspace\`: 760 pass; the 3 pre-existing fixture failures (filed as #1366, structurally pre-existing on commit \`0af74301c\`) remain

## Out-of-scope tracker

Issues filed for follow-ups identified during this review pass:
- #1366 — arity-2 sql-query body-template variant missing
- #1367 — mint \`concept:ecma262-number-format\` to retire RFC 8785 number-canonicalization loss
- #1368 — \`concept:Ref<T>\` parametric inner T resolution

Issue #1397, opened 2026-05-23
[Descope v1] PR-1: relocate reverse-direction realizations to kit-source-aliases/


TSavo:
## What It Does

Seven realizations currently live in the wrong location: `menagerie/concept-shapes/catalog/realizations/` contains `rust:Option->concept:option`, `rust:Vec->concept:list`, and 5 others that map language-native types to concepts. These belong in `kit-source-aliases/`. Moving them eliminates 2 apparent singleton concepts (`concept:option`, `concept:pair`) by correctly counting their realizations.

## Files Affected

- `menagerie/concept-shapes/catalog/realizations/` (7 files removed)
- `kit-source-aliases/` (7 files added)
- Catalog index if one exists

## Acceptance Criteria

- No singleton warnings for `concept:option` or `concept:pair` after PR-0 validator runs
- All existing catalog tests pass
- No changes to concept hub definitions

## Dependencies

PR-0 (to verify the singleton count drops)

TSavo:
PR-1 premise check — surfacing findings instead of executing the move.

## Evidence gathered

After orienting on the worktree (`agent-aaa93abce554cc5b2`), the brief's "7 mis-filed entries" framing does not survive primary-source inspection:

### 1. The 7 entries are intentional M-edges, not mis-files

All 7 have `"role": "abstraction-lift"` (source-language -> hub). The other 67 in `realizations/` have `"role": "abstraction-realization"` (hub -> language). Exact split: `grep -l '"role": "abstraction-lift"' menagerie/concept-shapes/catalog/realizations/*.json | wc -l` returns 7.

### 2. They are minted on every build

`mint.sh` runs `scripts/mint_{option,pair,list,result,identity,unit,option_bind}.py`. Each one explicitly mints both the `abstraction-lift` (M edge) and `abstraction-realization` (N edge). Example from `mint_option.py:209`:
```
"fn_name": "rust:Option->concept:option",
...
"role": "abstraction-lift",
"direction": "left-to-right",
"source_lang": "rust",
```
A `git mv` decays on the next `mint.sh` run.

### 3. 2 of 7 are operations, not types

- `rust:identity` is a function (`fn identity(x) -> x`).
- `rust:Option::and_then` is a method on `Option<T>`.

Neither fits the `KitSourceAlias` schema (which carries `source_aliases: [<type-name>]`).

### 4. 2 of 7 already exist in kit-source-aliases

[...]

Issue #1396, opened 2026-05-23
[Descope v1] PR-0: library:/concept: singleton validator


TSavo:
## What It Does

Adds a lift-time scan that flags any `concept:X` hub that appears in fewer than 2 distinct realizations (i.e., singleton concepts). These represent incomplete contracts that haven't achieved cross-language generality. Emits `lift-gap` log entries; does not fail the lift. Cheap, reversible, no breaking changes.

## Files Affected

- `crates/provekit-walk/src/walk_rpc.rs` (or equivalent lift-dispatch entry point)
- Java lift kit equivalent if one exists
- Possibly `libprovekit/src/manifest.rs` for gap-entry schema

## Acceptance Criteria

- `provekit lift` emits a `lift-gap` line for each singleton concept
- No existing tests broken
- ~80 LoC addition

## Dependencies

None. Ships first.

TSavo:
PR-0 (singleton-concept lift-gap validator) is up: #1428

Summary of what shipped:
- Lift-time scan in `provekit-walk/src/bin/walk_rpc.rs` flags `#[provekit::sugar(concept = "concept:X")]` with `< 2` realizations, emitting a non-fatal `lift-gap` / `singleton-concept` diagnostic (function name, `file:line`, concept, count, suggested resolution).
- `library:X` prefixes skipped entirely. Gap-emitting, not refusing: existing `library-sugar-binding-entry` still emitted.
- Counting rule absorbs #1397 (PR-1 closed won't-fix): only `abstraction-realization` N-edges count; the 7 reverse-direction `abstraction-lift` M-edges are excluded.

Verification:
- `ir` output byte-identical to main (sha256 `fd8313ce...`), so the 10/10 cycle holds by construction; only `diagnostics` changes (0 -> 10).
- Lift on `libprovekit-rpc-cross-platform` surfaces exactly 10 `singleton-concept` gaps (one per `@sugar` fn) — expected; the existing-singleton corpus audit is a separate follow-up PR.
- 6 new tests (discrimination triplet + M-edge-exclusion guards) pass; `cargo build --workspace` green.

TSavo:
✅ Merged to main (HEAD `95b9632fd`). Full test matrix green on merged main: rust build + suites, java mvn (exit 0), python 47 passed. PR-0 #1428.

TSavo:
Closing this old descope-v1 tracker item as superseded by the current project boundary, not because the descope was wrong.

The accepted direction is:

[...]

PR #1428, opened 2026-05-23
feat(lift): singleton-concept lift-gap validator (#1396)

TSavo:
## Summary

PR-0 of the descope sequence (#1396). Adds a lift-time scan in the rust bind-lift surface (`provekit-walk/src/bin/walk_rpc.rs`) that flags `#[provekit::sugar(concept = "concept:X")]` annotations whose `concept:X` has fewer than 2 realizations in the catalog. Emits a non-fatal `lift-gap` diagnostic of category `singleton-concept` naming the offending function, `file:line`, the concept, the realization count, and a suggested resolution (prefix-flip to `library:X`, or mint a second realization in another language).

Purely additive and reversible, as intended for PR-0.

## What it does

- **Gap-emitting, not refusing.** The existing `library-sugar-binding-entry` is still emitted for every `@sugar` function. The gap rides the existing `diagnostics` side channel of the `ir-document` lift result.
- **`library:X` prefixes are skipped entirely** — no catalog lookup, no gap. They are already-honest kit-specific identifiers with no cross-language claim.
- **Counting rule (absorbs #1397).** PR-1 was closed won't-fix after fact-finding showed the 7 reverse-direction (`lang:form->concept:X`) catalog entries are intentional `abstraction-lift` M-edges, not mis-filed N-edges. This validator therefore counts **only `abstraction-realization` N-edges** toward the `>=2` threshold. Role correlates exactly with direction in the catalog (every `concept:X->...` is `abstraction-realization`; every `...->concept:X` is `abstraction-lift`), so direction-filtering the embedded `index.json` is equivalent to filtering `role == "abstraction-realization"`. Net effect: `concept:option` and `concept:pair` (1 forward realization + 1 reverse M-edge each) are correctly treated as singletons.
- The diagnostic uses field `function_name` (not `fn_name`) deliberately, to avoid the bind-payload `fn_name`-outside-`gapRecords` hygiene invariant enforced elsewhere in the file.

## Expected: existing singletons surface gaps

The 10 `@sugar` functions in `libprovekit-rpc-cross-platform` use brand-new `concept:X` names (0 realizations), so they each emit a `singleton-concept` gap. **This is expected and informational** — the gap-emit is the tool a follow-up audit PR will use to find offenders. This PR does NOT do the corpus audit (renaming `concept:` -> `library:` for the existing singletons).

## Verification

- `cargo build --workspace` green.
- `cargo test -p provekit-walk --bin provekit-walk-rpc`: 48 pass + 6 new tests pass. The single pre-existing failure (`bind_lift_erases_signature_types_from_bind_ir_entries`) and the pre-existing `libprovekit` test-code compile errors both reproduce on stock `main` and are unrelated to this change.
- Lift run on `libprovekit-rpc-cross-platform`: exactly **10** `singleton-concept` gaps emitted (one per `@sugar` function), all `library-sugar-binding-entry` entries still present.
- **Cycle test (10/10 byte-identical) holds by construction.** The `ir` array output is **byte-identical** before/after this change (sha256 `fd8313ce...` unchanged); only `diagnostics` differs (0 -> 10). Since the cycle pipeline (bind -> lower -> realize) consumes `ir` and `diagnostics` is a side channel that does not feed it, the manual `provekit lift | lower | java lift | realize-rust` cycle in `examples/provekit-rpc-java-demo/diagnostics/symmetric-cycle-scorecard.md` is unaffected. The hash check is stricter evidence than re-running the documented procedure.

## Tests added (discrimination triplet + counting-rule guards)

- positive: unknown `concept:X` (0 realizations) -> gap emitted
- negative: `concept:closure` (>=2 realizations) -> no gap
- structural: `library:X` prefix -> skipped, no lookup, no gap (and `library-sugar-binding-entry` still emitted)
- M-edge exclusion (helper + end-to-end): `concept:option` / `concept:pair` (1 N-edge + 1 M-edge) -> singleton -> gap; M-edge does not inflate the count


TSavo:

**Verdict: APPROVE**

I did not write this code. I verified every load-bearing claim empirically against the real catalog and by running the tests; nothing is taken on trust.

### Singleton detection is correct
- `concept` is the raw annotation string (`parsed.concept`, e.g. `"concept:closure"` / `"library:foo"`); the `concept.starts_with("concept:")` guard at `walk_rpc.rs:456` correctly skips `library:` prefixes entirely (no catalog lookup, no gap). Confirmed in code, and the structural test asserts it.
- The counting helper (`concept_realization_counts`, `walk_rpc.rs:64`) counts only forward `concept:X->...` / `concept:X:...` entries and explicitly `continue`s on `name.contains("->concept:")`. I replicated the exact logic against the real `menagerie/concept-shapes/catalog/index.json`:
  - **Exactly 7** reverse `->concept:` M-edges exist (`rust:()->concept:unit`, `rust:Option->concept:option`, `rust:Option::and_then->concept:option-bind`, `rust:Result->concept:result`, `rust:Vec->concept:list`, `rust:identity->concept:identity`, `rust:tuple->concept:pair`) — all correctly excluded.
  - `concept:option`=1, `concept:pair`=1, `concept:option-bind`=3, `concept:closure`=4 — match the test assertions exactly. The two singletons-with-an-M-edge-companion (`option`, `pair`) are correctly flagged.
- The role==`abstraction-realization` filter is implemented as a direction filter, which is sound: in this catalog direction correlates exactly with role. The doc comment states this explicitly and #1397 (M-edges intentional) backs it.

### Gap-emitting, not refusing — confirmed
The validator block (`walk_rpc.rs:456-470`) only pushes to the `diagnostics` side channel. No early return, no control-flow change. Execution falls straight through to the unchanged `library-sugar-binding-entry` build+push into `ir`. The structural test asserts the sugar entry is still emitted alongside a skipped lookup.

### Byte-identical `ir` claim — verified by reading, not trusting
The diff touches nothing on the `ir` push path. The new code adds (a) a `diagnostics.push(...)` and (b) two free helpers. `diagnostics` rides alongside `ir` and does not feed the bind->lower->realize cycle. The byte-identical-`ir` claim is structurally true.

### Tests are real discrimination tests, not vacuous
Ran all 6: `test result: ok. 6 passed`. They are a genuine triplet plus counting-rule guards — positive (unknown concept -> gap), negative (`concept:closure` >=2 -> no gap), structural (`library:` skipped + sugar entry still emitted), helper sanity, M-edge-exclusion (asserts exact counts 1/1/3), and end-to-end M-edge companion. Each asserts concrete field values, not just presence.

Clean, scoped, reversible. Approve.

Issue #1404, opened 2026-05-23
[Descope v1] PR-8: build provekit-emit-rust-cargo-test plugin


TSavo:
## What It Does

Builds the `provekit-emit-rust-cargo-test` plugin (~600 LoC). A PEP 1.7.0 realize plugin that emits Rust `#[test]` assertion patterns from catalog predicates. Completes the three-language emitter baseline (Java/Python/Rust).

## Files Affected

- New crate: `crates/provekit-emit-rust-cargo-test/`
- Kit manifest: `.provekit/emit/rust-cargo-test.json`
- Integration with kit-dispatch

## Acceptance Criteria

- Plugin emits valid Rust test assertions for seed templates from PR-5
- `provekit emit --lang rust --framework cargo-test` runs end-to-end
- Output is a signed memento
- ~600 LoC

## Dependencies

PR-2 (substrate crate split), PR-5 (assertion template catalog)

TSavo:
## Reframed — mapping is kit-internal, no catalog dependency

PR-5 (#1401) is closed. It tried to externalize the predicate→assertion mapping into a `JUnitAssertionTemplateMemento` catalog family. That was wrong: the fact that `concept:eq` spells as `cargo test`'s assertion form is **rust's framework knowledge**, not substrate data. It belongs in this kit's code, written in rust.

### Revised scope for this PR

Build the `provekit-emit-rust-cargotest` emitter plugin (PEP 1.7.0 realize plugin) with the predicate→assertion mapping **inline in the kit**, the same way the old SugarRealizer carried `concept:utf8-encode → getBytes` inline (that part of the cycle design was correct).

- Input: a contract's predicates (from the neutral catalog) + the function scaffold.
- Output: cargo test test source with native assertions. Examples: `assert_eq!(a, b) / assert!(x.is_some()) / #[should_panic]`.
- The mapping table is rust code in this kit. No menagerie memento family. No catalog read for templates.

### What the kit DOES read from the catalog

- The neutral concept (`concept:eq`, `concept:option-is-some`, etc.) — to know WHICH predicate it's emitting.
- The realization claim (existing realization-memento family) — to confirm this kit's assertion form is a registered realization of that concept.

Framework knowledge (the actual `cargo test` spelling) stays in the kit.

### Bidirectional note

If the harvester (`provekit-lift-rust-*`) and this emitter want to share the predicate↔assertion mapping, they share a rust module — a normal kit-ecosystem dependency, not a substrate catalog memento.

[...]

PR #1430, opened 2026-05-23
feat(rust): provekit-emit-rust-cargo-test emitter kit (PR-8, #1404)

TSavo:
## Summary

PR-8 (#1404): the rust EMITTER kit `provekit-emit-rust-cargo-test`. Given a contract's neutral predicates (`IrFormula`) plus the target function signature, it materializes a `#[cfg(test)] mod tests { ... }` block whose `#[test]` functions assert the predicates using the standard `assert!` / `assert_eq!` / `assert_ne!` macros. It is the inverse of `provekit-lift-rust-tests` (the rust HARVESTER).

- Predicate -> cargo-test-assertion mapping is **INLINE rust framework knowledge** in this kit, NOT a catalog memento family. (PR-5's externalized approach was closed as an architectural mistake; framework spelling belongs to the framework kit.)
- Keys off the canonical `IrFormula::Atomic` atom symbols the harvester actually emits (`=`, `≠`, `<`, `≤`, `>`, `≥` per `provekit-ir-symbolic`), with `concept:*` sugar aliases accepted as a convenience.
- PEP 1.7.0 RPC server matching the rust kit conventions (`initialize` / `provekit.plugin.invoke` / `shutdown`, JSON-RPC over stdio). Result carries `source` + `is_stub` so it slots into kit-dispatch's realize consumer.
- Self-contained: depends only on `provekit-ir-types` (no `provekit-realize-rust-core`, no `provekit-walk`).
- ~420 LoC of mapping/template/RPC + tests; new code, does not touch the 10/10 cycle.

## Supported predicates

| Neutral atom | Emitted assertion |
|---|---|
| `=` (`concept:eq`) | `assert_eq!(a, b)` |
| `≠` (`concept:ne`) | `assert_ne!(a, b)` |
| `<` (`concept:lt`) | `assert!(a < b)` |
| `≤` (`concept:le`) | `assert!(a <= b)` |
| `>` (`concept:gt`) | `assert!(a > b)` |
| `≥` (`concept:ge`) | `assert!(a >= b)` |
| `=(x, Some(_))` | `assert!(x.is_some())` |
| `=(x, None)` | `assert!(x.is_none())` |
| `fallible-err(x)` (`concept:fallible-err`) | `assert!(x.is_err())` |

Term rendering covers `Var`, `Const` (int/string/bool), and `Ctor` shapes (free-call, tuple-struct, infix operators `+ - * / %`, `array`/`tuple`/`vec`, method receivers `len`/`is_some`/`is_none`/`is_err`/`is_ok`/`abs`/`count`). `Some(<concrete>)` keeps the payload equality rather than collapsing to `.is_some()`. Top-level `And` conjunctions flatten to one `#[test]` per conjunct. `Or` / `Not` / `Implies` / `Forall` / `Exists` are reported as skipped-not-emitted (not materializable as concrete cargo tests). There is no primitive `throws` atom in the substrate today, so `#[should_panic]` is intentionally not emitted; `fallible-err` is the single extension point if/when such a concept lands.

## Test plan

- [x] `cargo build -p provekit-emit-rust-cargo-test` green
- [x] `cargo test -p provekit-emit-rust-cargo-test` green (18 unit + 3 end-to-end)
- [x] End-to-end: contract in -> emitted module out -> `rustc --test --edition=2021` compiles -> test binary runs green (eq, option-is-some, multi-predicate conjunction)
- [x] `cargo clippy -p provekit-emit-rust-cargo-test --all-targets` clean (only pre-existing `provekit-ir-types` warnings)
- [x] Registered in workspace `Cargo.toml` members


TSavo:

Independent reviewer, did not write this. Checked out the worktree, built, ran tests, and empirically reproduced the failure mode below.

**Verdict: REQUEST-CHANGES** (one real compile-correctness bug; everything else is sound)

### Principle adherence — all PASS
- **Mapping inline, not catalog-read**: PASS. `grep` for `catalog|memento|read_to_string|File::open|menagerie` in `src/`+`tests/` finds only doc comments and the `emits_signed_mementos:false` capability flag. No catalog reads. Mapping is fully inline in `lib.rs`.
- **Keys off real canonical atoms**: PASS. `canonical_atom` (lib.rs:162) maps `= ≠ < ≤ > ≥` and accepts `concept:*` sugar as aliases. `option_assertion` (lib.rs:321) recognizes `=(x, Some(_))` / `=(x, None)` structurally and `Some(<concrete>)` keeps payload equality (`some_with_concrete_payload_keeps_equality`). No invented primitive atoms.
- **E2e really invokes rustc**: PASS, verified. `tests/end_to_end.rs:880` `compile_and_run_tests` does `Command::new("rustc").arg("--test").arg("--edition=2021")`, then runs the produced binary and asserts BOTH `compiled` AND `passed`. Not string-match. `cargo test -p provekit-emit-rust-cargo-test` → 18 unit + 3 e2e green locally.
- **Self-contained**: PASS. `Cargo.toml` deps = `provekit-ir-types`, `serde`, `serde_json`, `tempfile` (dev only). No `provekit-realize-rust-core`, no `provekit-walk`.
- **Unsupported predicates skipped-not-emitted**: PASS. `flatten_atoms` (lib.rs:359) routes `Or/Not/Implies/Forall/Exists` to `unsupported`; `quantifier_is_skipped_not_emitted` confirms `tests.len()==0, skipped.len()==1`. Never vacuously passed.
- **Registered in workspace members**: PASS (Cargo.toml:30).

### Codex automated review — both findings VALID
- **P1 (bool ctor) — CONFIRMED, BLOCKING.** The harvester at `provekit-lift-rust-tests/src/lib.rs:1220` lifts `syn::Lit::Bool` to `Ctor("True"|"False", [])`. PR-8 `render_ctor` (lib.rs:242) has no arm for these, so they fall through to the nullary free-ctor branch and render the bare string `True`/`False`. I reproduced it: a contract `is_even(4) == true` emits `assert_eq!(is_even(4), True);` — **`True` is not valid Rust and will not compile.** The docstring claims the emitter recognizes "what the rust harvester lifts"; for bool literals it does not. The 18+3 tests stay green only because no test contract contains a bool literal — vacuous coverage on this exact path. Fix: one match arm (`"True" => "true"`, `"False" => "false"`) before the generic ctor fallback, plus an e2e regression that compiles a bool-literal contract through rustc.
- **P2 (1-element tuple) — CONFIRMED, nit.** Harvester emits `(v,)` as `Ctor("tuple",[v])` (lib.rs:1285, syn only matches a trailing-comma tuple). PR-8 renders it `(v)` (lib.rs:228), collapsing to a parenthesized scalar and losing tuple type. Lower severity (rare shape) but a real round-trip fidelity gap.

### Additional nits (non-blocking)
- e2e does NOT exercise the `fallible-err` path through rustc (only eq / option-some / conjunction). Unit-tested textually only. Add an `.is_err()` e2e for parity.
- `tempfile = "3"` is not workspace-pinned; other crates use `{ workspace = true }`. Consistency nit.

Solid kit, correct architecture, genuinely-executing e2e harness. The one blocker is the bool-literal compile bug — it directly contradicts the "recognize what the harvester emits" principle and is hidden by absent coverage. Fix P1, add the regression, and this is an approve.

TSavo:
Fix landed in b077641f1.

**Bug (the request-changes blocker):** the harvester lifts bool literals as `Ctor("True"/"False", [])` (`provekit-lift-rust-tests`, `syn::Lit::Bool` at lib.rs:1219). The emitter rendered those as bare `True`/`False`, producing invalid rust like `assert_eq!(is_even(4), True);`. Confirmed at the cited line.

**Fix:**
1. Added `render_ctor` arms: `Ctor("True",[])` -> `true`, `Ctor("False",[])` -> `false` (rust lowercase bool literals).
2. Added `e2e_bool_literal_contract_compiles_and_passes`: a contract with two bool-literal predicates (`is_even(4) == true`, `is_even(3) == false`), emitted, compiled via `rustc --test`, run, asserted green. It also guards that the emitted source contains no bare `True`/`False`. This is the test that would have caught the original bug. Plus a `bool_literal_ctor_renders_lowercase` unit test.

**Nits also fixed:**
- 1-tuple `Ctor("tuple",[x])` now renders `(x,)` with the trailing comma so it round-trips as a tuple (unit test `one_tuple_renders_with_trailing_comma`).
- Added `e2e_fallible_err_contract_compiles_and_passes` exercising `.is_err()` through `rustc` (`parse("not-a-number")` Result contract).
- `tempfile` now pinned via `[workspace.dependencies]`.

`cargo test -p provekit-emit-rust-cargo-test` green: **20 unit + 5 end-to-end**.

Issue #1403, opened 2026-05-23
[Descope v1] PR-7: build provekit-emit-python-pytest plugin


TSavo:
## What It Does

Builds the `provekit-emit-python-pytest` plugin (~500 LoC). A PEP 1.7.0 realize plugin that emits pytest assertion patterns from catalog predicates. Parallel to PR-6 in structure. Demonstrates cross-language emitter symmetry.

## Files Affected

- New plugin: `kits/provekit-emit-python-pytest/` (Python package)
- Kit manifest: `.provekit/emit/python-pytest.json`
- Integration with kit-dispatch

## Acceptance Criteria

- Plugin emits valid pytest assertions for each seed template from PR-5
- `provekit emit --lang python --framework pytest` runs end-to-end
- Output is a signed memento
- ~500 LoC

## Dependencies

PR-2 (substrate crate split), PR-5 (assertion template catalog)

TSavo:
## Reframed — mapping is kit-internal, no catalog dependency

PR-5 (#1401) is closed. It tried to externalize the predicate→assertion mapping into a `JUnitAssertionTemplateMemento` catalog family. That was wrong: the fact that `concept:eq` spells as `pytest`'s assertion form is **python's framework knowledge**, not substrate data. It belongs in this kit's code, written in python.

### Revised scope for this PR

Build the `provekit-emit-python-pytest` emitter plugin (PEP 1.7.0 realize plugin) with the predicate→assertion mapping **inline in the kit**, the same way the old SugarRealizer carried `concept:utf8-encode → getBytes` inline (that part of the cycle design was correct).

- Input: a contract's predicates (from the neutral catalog) + the function scaffold.
- Output: pytest test source with native assertions. Examples: `assert a == b / assert x is not None / pytest.raises(...)`.
- The mapping table is python code in this kit. No menagerie memento family. No catalog read for templates.

### What the kit DOES read from the catalog

- The neutral concept (`concept:eq`, `concept:option-is-some`, etc.) — to know WHICH predicate it's emitting.
- The realization claim (existing realization-memento family) — to confirm this kit's assertion form is a registered realization of that concept.

Framework knowledge (the actual `pytest` spelling) stays in the kit.

### Bidirectional note

If the harvester (`provekit-lift-python-*`) and this emitter want to share the predicate↔assertion mapping, they share a python module — a normal kit-ecosystem dependency, not a substrate catalog memento.

### Dependency change

[...]

Issue #1401, opened 2026-05-23
[Descope v1] PR-5: add JUnitAssertionTemplateMemento family to catalog


TSavo:
## What It Does

Defines the JSON schema and memento family for mapping catalog predicates to JUnit assertion patterns. Each `JUnitAssertionTemplateMemento` binds a concept predicate to a JUnit 5 assertion call template. This is the catalog-side contract that the `provekit-emit-java-junit` plugin (PR-6) will consume.

## Files Affected

- `menagerie/concept-shapes/catalog/junit-assertion-templates/` (new directory)
- Schema definition: `menagerie/concept-shapes/catalog/schemas/junit-assertion-template.json`
- At minimum 5 seed templates covering null-check, non-null, range-bound, equality, boolean predicates

## Acceptance Criteria

- Schema validates against existing catalog tooling
- At least 5 seed templates present and schema-valid
- Templates are content-addressed (CID in manifest)

## Dependencies

PR-1 (concepts must be correctly counted before templates reference them)

TSavo:
PR-5 opened: https://github.com/TSavo/provekit/pull/1427

Adds the `assertion-templates/` catalog family + JUnit5 seed batch.

**Seed predicates included (7):**
- `concept:option-is-some` -> `assertNotNull`
- `concept:eq` -> `assertEquals`
- `concept:ne` -> `assertNotEquals`
- `concept:lt` -> `assertTrue ({a} < {b})`
- `concept:gt` -> `assertTrue ({a} > {b})`
- `concept:le` -> `assertTrue ({a} <= {b})`
- `concept:ge` -> `assertTrue ({a} >= {b})`

The four comparison ops collapse onto `assertTrue` at the assertion-name level; `harvest_pattern` carries the inline operator literally so the harvester table is injective on harvest.

**Skipped (predicate concept hub does not yet exist; follow-up):**
- `concept:option-is-none` (no hub yet)
- `concept:list-empty`, `concept:list-nonempty` (no hubs yet)
- `concept:throws` / `concept:fallible-err` (no hub yet; `concept:throw` exists but models the throw statement, not the predicate that a body throws, so wiring `assertThrows` to it would make the harvester misidentify any source `throw` as an assertion call)
- `concept:bool-true`, `concept:bool-false` (no hubs yet)

Catalog-data only; no rust or java code changes. Mint script is byte-identical idempotent.

TSavo:
[...]

PR #1431, opened 2026-05-23
PR-7: provekit-emit-python-pytest (inline predicate->pytest mapping)

TSavo:
## Summary

Builds `provekit-emit-python-pytest` — a PEP 1.7.0 realize plugin (target framework = `pytest`) that takes a contract's neutral predicates plus the function signature and emits a pytest test module whose `test_*` functions assert those predicates.

- The predicate -> assertion mapping is **inline python** in `predicate_table.py`. The fact that `concept:eq` spells as `assert a == b` is python framework knowledge, not substrate data. **No catalog memento family, no catalog read for templates** — this reverses the closed PR-5 (#1401) `JUnitAssertionTemplateMemento` mistake, per the reframe in #1403.
- Python sibling of `provekit-emit-java-junit` (PR-6) and the inverse of the `provekit-lift-py-tests` harvester.
- Per-predicate placeholder values are chosen so each emitted test **passes** standalone (e.g. `lt` -> operands `0`,`1`; `option-is-some` -> `object()`; `fallible-err` -> a raising callable), not merely imports.

### Supported predicates
`concept:eq`, `concept:ne`, `concept:lt`, `concept:gt`, `concept:le`, `concept:ge`, `concept:option-is-some`, `concept:option-is-none`, `concept:fallible-err` (-> `with pytest.raises(Exception): x()`).

Unsupported predicates are recorded as `unsupported_predicates` (honest emit-assertion-gap), never emitted as vacuously-passing tests.

### RPC
`provekit.plugin.describe` / `invoke` / `shutdown` over newline-delimited JSON-RPC (`--rpc`), matching the python realize kits and the java JUnit emitter.

### Integration
New package under `implementations/python/provekit-emit-python-pytest/`, wired into `make test-python`.

## Test plan
- [x] Unit tests: predicate table (each supported predicate + arity/malformed refusal + const/arithmetic/call term rendering), emitter (gap accounting, CID determinism, conditional `import pytest`, param parsing).
- [x] RPC: describe/invoke/shutdown/error paths, JSON-serializable.
- [x] End-to-end: emit a module per supported predicate + a multi-predicate clamp contract, write to disk, run `pytest` in a subprocess, assert green.
- [x] `pip install -e .` in a fresh venv + `pytest`: **40 passed**.
- [x] `--rpc` console-script smoke verified.

Closes part of #1403.


TSavo:

Reviewer did not author this code. Checked out `pr1431-review`, ran the suite, verified the claimed bug-fix against the sibling kit.

**Verdict: APPROVE-WITH-NITS**

### Descope principles — all honored
- **Mapping is inline, not catalog-read.** `predicate_table.py` `_HANDLERS` is a hardcoded python dict of lambdas (predicate_table.py:645). No catalog memento family, no catalog read for the framework spelling. Corrected PR-5 reframe, lands correctly.
- **Reads real harvester shapes.** `head_of()` (predicate_table.py:417) accepts both catalog `kind:"op"`/`name:"concept:eq"` and the harvester `kind:"atomic"`/bare-name form, stripping `concept:`. `option-is-some` is consumed as a catalog op-node, not invented. Same forward-looking caveat as PR-6: the python harvester doesn't lift these structurally yet.
- **No body-cycle dependency.** Depends only on `blake3`. No realize-core/SugarRealizer/walk.
- **Tests actually RUN, not import.** Verified: `pip install -e . && pytest` → **40 passed** (127s). `test_end_to_end.py:_run_pytest_on` (line 936) writes the emitted module to a tmp file and `subprocess.run([sys.executable, "-m", "pytest", module, "-q"])`, asserting `returncode == 0`. That is a genuine green-run of emitted tests, confirmed by running it (the 2-min runtime is the per-predicate pytest subprocesses). Placeholder values are chosen so assertions actually pass (`lt` → `0`,`1`; `option-is-some` → `object()`; `fallible-err` → a raising callable). Operator correctness is independently pinned by `test_predicate_table.py` exact-string assertions (e.g. `assert a <= b`), so a `0<=0`-style placeholder can't mask an operator swap.
- **Unsupported predicates surfaced, never vacuously passed.** `emitter.emit()` (emitter.py:296) collects gaps; `is_complete` requires unsupported empty AND emitted non-empty; e2e asserts `is_complete` before running pytest so an empty module can't sneak a green. `test_unsupported_predicate_recorded_as_gap_not_emitted` + `test_invoke_reports_unsupported_gap` pin this.

### Claimed bug-fix — REAL and tested
The agent claimed it avoided `params or {}` coercing `[]` → `{}` and bypassing the `-32602` invalid-params guard. Confirmed: `rpc.dispatch()` (rpc.py:736-737) uses `if params is None: params = {}`, which preserves `[]`, so `isinstance(params, dict)` correctly fails and returns `-32602`. `test_invalid_params_error` (test_rpc.py:1201) passes `"params": []` and asserts code `-32602`. The fix is real and covered.

**Boy-scout follow-up (out of scope here):** the sibling `implementations/python/provekit-realize-python-sqlite3/src/.../rpc.py:34` STILL has the latent bug — `params = request.get("params") or {}` — so `[]` is coerced to `{}` there and the guard is bypassed. Free fix for whoever takes the next python kit.

### RPC conventions
Matches the python realize family on shutdown/error/`-32602`/JSON-serializable responses. Note: `provekit.plugin.describe` is NEW here — the sibling `provekit-realize-python-sqlite3` has no `describe` method at all (uses invoke/emit_module/platform_semantics/shutdown). So there's no python precedent for the describe shape.

### Nit to track (non-blocking)
- `describe` returns a flat capability object, same as PR-6. The language-agnostic `provekit-plugin-loader` (loader.rs:194) requires `{envelope, header, metadata}` for ANY plugin it loads, so this kit would also be refused if loaded via that path. Weighing this lighter than PR-6's instance because python has no sibling returning an envelope to be inconsistent with — but the loader constraint is the same. Track as a loader-integration follow-up.
- `fallible-err → pytest.raises` cannot round-trip: the python harvester `provekit-lift-py-tests` explicitly lists `pytest.raises` as out-of-scope (PKG-INFO / `__init__.py:19`). Symmetric to PR-6's assertThrows gap — a known asymmetry, not an oversight.

Mapping inline ✓ · emitted tests run green in subprocess ✓ · no vacuous passes ✓ · bug-fix real + tested ✓.

TSavo:
## Review follow-up landed (39e3ca77b)

### describe-envelope fix (the blocker)

`provekit.plugin.describe` now returns a full PEP 1.7.0 plugin memento `{envelope, header, metadata}` instead of a flat object. New `plugin_memento.py` mints the header with a §6.1 content CID computed in python (JCS / RFC 8785 + blake3-512), byte-identical to the rust loader's `compute_plugin_cid`. Capabilities moved into `header.content` (opaque to the loader). Signing is a zero-bytes ed25519 placeholder (§12 out-of-scope for the loader skeleton; real signing is the loader-registry follow-up).

**Tests** (`test_plugin_memento.py`) are adversarial, not circular:
- independently re-derive the CID from header fields with a separate encoder and assert the kit's minted CID matches (what the loader does, refusing on mismatch);
- assert the `cid` field is elided from the CID input (mutating it doesn't change the recomputation);
- **golden test** pins the python JCS against the rust loader's checked-in `dummy-sugar.json` fixture CID, so any python/rust canonical-encoder drift breaks the build;
- assert `schemaVersion == "1"`, `pep/1.7.0` in `protocol_versions`, and that the loader's `runtime_protocol_versions` describe-param doesn't crash dispatch.

**End-to-end proof:** built `provekit-plugin-loader` and ran the real loader against this kit via `stdio:` — `LOADED OK, kind: realize, cid: blake3-512:9c0aa021...` (matches the kit's emitted CID). The gauntlet load path works.

### Boy-scout: `params or {}` guard bug

The same latent bug I fixed here also lived in **four** python realize kits (`sqlite3`, `aiosqlite`, `core`, `requests`): a falsy non-dict `params` (`[]`) was coerced to `{}` via `or`, bypassing the `isinstance(params, dict)` guard and turning an `INVALID_PARAMS` (-32602) error into a silent default plan. Fixed all four to default only when `params is None`. Verified each now returns -32602 for non-dict params and still handles absent params.

(Note: 3 pre-existing `concept:sql-query` template failures in the sqlite3 kit's own test suite are unrelated to this change — confirmed identical on the original code — and out of scope.)

### Result
**47 tests pass** (was 40; +7 memento/describe). The slow subprocess e2e tests (emit -> write -> run pytest green) still pass.

Issue #1421, opened 2026-05-23
[Descope v1 Q-5] Coq / Lean solver backends: keep both, pick one, or defer to Phase 2?


TSavo:
## Question

ProvekIt has two solver backends: Coq and Lean. Both discharge contract witnesses. For v1 "Witnesses" pillar, only one backend needs to work end-to-end; shipping two creates maintenance surface with no user benefit in v1.

## Why It Matters

The witness signing pipeline is the product's claim to correctness. An untested backend is a liability. Picking one and deferring the other focuses the v1 correctness story.

## Options Under Consideration

1. **Keep Lean, defer Coq**: Lean's mathlib ecosystem is more active; better long-term bet.
2. **Keep Coq, defer Lean**: Coq is more mature for program verification; more prior art.
3. **Keep both, gate each separately**: Two optional CI gates; neither blocks v1 ship.
4. **Defer both to Phase 2**: Ship v1 with an abstract witness interface; pluggable solvers are v1.1.

## Who Should Answer

@TSavo — determines the witness story for v1.

_Source: Rust audit `/tmp/cycle-1391/audit-rust.md`_

TSavo:
## Confirmed — all four KEEP, no compromise

User feedback is unambiguous: Coq, Lean, Z3, and Vampire are foundational to the WITNESSES pillar. Without solver pluralism, the substrate degrades to "content-addressing for things one solver can prove." That's not the descoped claim.

### What's actually in the workspace

Confirmed crates:
- \`provekit-ir-compiler-coq\` — Coq backend (interactive proof assistant, dependent types)
- \`provekit-ir-compiler-lean\` — Lean backend (same domain, different philosophy)
- \`provekit-ir-compiler-smt-lib\` — SMT-LIB target (Z3 + other SMT solvers)
- \`provekit-ir-compiler-maude\` — Maude backend (rewriting-logic prover; relevant for desugar discharge)

Plus Vampire registered in \`provekit-verifier/src/solvers/registry.rs\` (first-order theorem prover for non-trivial quantification).

### Architectural rationale

Different solver backends handle different obligation shapes well:

| Backend | Strength |
|---|---|
| Z3 (via SMT-LIB) | Arithmetic, bitvectors, arrays, fast SMT decisions; the workhorse |
| Vampire | First-order theorem proving with rich quantification |
| Lean | Dependent types, interactive proof, deep theorem-level claims |
| Coq | Dependent types, second community/style for interactive proof |
[...]

Issue #1405, opened 2026-05-23
[Descope v1] PR-9: unified provekit verify CLI verb


TSavo:
## What It Does

Stitches the existing prove + witness + check + emit-signed-receipt pipeline into a single `provekit verify` user-facing entry point. Currently these are separate subcommands with no unified entry; users must chain them manually. The unified verb runs the full pipeline and produces a single signed receipt.

## Files Affected

- `crates/provekit-cli/src/cmd_verify.rs` (new or refactored)
- `crates/provekit-cli/src/main.rs` (verb registration)
- Existing `cmd_prove.rs`, `cmd_check.rs` wired as sub-steps

## Acceptance Criteria

- `provekit verify <target>` runs the full prove-witness-check-sign pipeline
- Outputs a CID-stamped signed receipt
- Supersedes the need to manually chain subcommands
- Existing subcommands remain but are documented as low-level

## Dependencies

PR-6, PR-7, PR-8 (emitters must exist for verify to be useful)

TSavo:
PR-9 opened: https://github.com/TSavo/provekit/pull/1433 (branch `pr-9-verify-verb`)

`provekit verify <kit>` is now a real GATE verb: lift contract claims → build the wp/refinement obligation → route through the solver-dispatch table → mint a signed witness citing the discharging solver → emit a per-claim receipt. `prove` (six-stage prover + `--kit` C1–C8 conformance gate) is untouched.

**Sample receipt** (captured end-to-end through the binary against a real `.proof` catalog; LIA claim routed to z3, discharged, signed witness minted):

```json
{
  "kind": "verification-receipt",
  "schemaVersion": "1",
  "claims": [
    {
      "property": "parseInt",
      "obligationClass": "linear-arithmetic",
      "routedSolver": "z3",
      "dischargingSolver": "z3@4.x",
      "status": "discharged",
      "pass": true,
      "reason": "solver 'z3' returned unsat (obligation holds)",
      "witnessCid": "blake3-512:279e1dbae018c2ad74da203db5caea6579b81af3392c569ab1d033f8f0a35b9c..."
    }
  ],
  "totalClaims": 1,
  "discharged": 1,
  "failed": 0,
  "ok": true
}
```

**Brief premises adjusted during integration** (PR-1 pushback pattern; full detail in the PR description):

[...]

PR #1433, opened 2026-05-23
feat(cli): real `provekit verify <kit>` GATE verb — lift→dispatch→witness→receipt

TSavo:
## Summary

PR-9, the keystone GATE verb (#1405). Splits `Cmd::Verify` off its old `prove` alias into a **real** verb that runs the kit-level verification flow end-to-end and emits a per-claim verification receipt:

1. **lift** — load the kit's `.proof` catalog + enumerate contract claims (`load_all_proofs` + `enumerate_callsites`)
2. **obligation** — resolve the target contract, build the wp / refinement obligation (reuses verifier `resolve_target` + `instantiate`)
3. **dispatch** — classify the obligation's theory and route it through the **solver-dispatch table** (`SolverPlan::Dispatch` / `DispatchConfig`, the existing primitive). v1 routing: LIA / BV / strings → SMT-LIB (z3 / cvc5 / bitwuzla), equational → maude, dependent / categorical → lean, default → z3. The table is the *structure* (`verify_dispatch_table()`), not a hardcoded backend; a kit's own `[solvers.dispatch]` wins when present.
4. **witness** — mint + Ed25519-sign a `WitnessMemento` citing the discharging solver so consumers can trust-discriminate by prover
5. **receipt** — per-claim pass/fail, obligation class, routed + discharging solver, witness CID; JSON + human output

## Brief premises adjusted during integration (the PR-1 pushback pattern)

The brief invited correction where its model didn't match the code. Four findings:

1. **The six-stage `Runner` already integrates lift-pool → wp → solver-dispatch → witness-mint.** The brief framed the flow as "scattered across `cmd_prove` / `cmd_witness` / `cmd_ci`," but `provekit-verifier::runner` already wires it. **PR-9 is the verb + the kit-level receipt on top, not the integration.** `cmd_verify` calls the verifier's public stages directly (`load_all_proofs`, `enumerate_callsites`, `resolve_target`, `instantiate`, `smt_emitter`, `run_plan`) so it can surface per-claim solver + witness data the global `Report` doesn't carry.
2. **`SolverPlan::Dispatch(DispatchConfig)` + `dispatch_for_formula` + `FormulaTheory` IS the solver-dispatch table** the brief asks to "build." PR-9 wires `verify` to it; it does not invent a new one. (Q-5 / #1421.)
3. **`cmd_witness` basic-mode is a stub** (literally `// TODO: mint witness memento`). The witness *primitives* (`WitnessMemento` + `ed25519_sign_string`) exist; the CLI wrapper does not. PR-9 mints + signs the witness inline rather than calling `cmd_witness::run`.
4. **`--kit` on `prove` (the C1–C8 lift-plugin conformance gate) is preserved untouched.** `verify <kit>` is a distinct verb with distinct semantics; `Cmd::Prove` keeps its own handler and `--kit` path.

## Sample receipt (captured end-to-end through the binary)

```json
{
  "kind": "verification-receipt",
  "schemaVersion": "1",
  "kit": null,
  "claims": [
    {
      "property": "parseInt",
      "obligationClass": "linear-arithmetic",
      "routedSolver": "z3",
      "dischargingSolver": "z3@4.x",
      "status": "discharged",
      "pass": true,
      "reason": "solver 'z3' returned unsat (obligation holds)",
      "witnessCid": "blake3-512:279e1dbae018c2ad74da203db5caea6579b81af3392c569ab1d033f8f0a35b9c..."
    }
  ],
  "totalClaims": 1,
  "discharged": 1,
  "failed": 0,
  "ok": true
}
```

## Test plan

- [x] `provekit-verifier` suite green (63+ tests; only re-export additions to `lib.rs` / `solvers/mod.rs`)
- [x] 3 new unit tests: dispatch-table routes each theory class; LIA obligation routed to the z3 SMT seat; minted witness canonicalizes + signs + re-parses
- [x] 2 new integration tests (`cmd_verify_integration.rs`): empty catalog → clean exit + zero-claim receipt; LIA claim built as a real `.proof` catalog → enumerated → routed to z3 → discharged → signed witness minted, all observed through the `provekit verify` binary
- [x] solver-dispatch table exercised: LIA obligation routed to SMT-LIB/z3, table structure supports adding more seats

## Pre-existing failures in this worktree (NOT introduced by PR-9)

Confirmed by stash + rerun (failures reproduce without my edits):
- `cmd_bind_migrate::tests::changes_for_targets_present_*_emits_uncharacterizable` (2) — sqlite migration fixture assertion, unrelated.
- `provekit-cli/tests/cmd_materialize_integration.rs` — unclosed-delimiter compile error in that test file (committed at `80570d378`, a merge in history).

Both are orthogonal to PR-9 and left out of scope.


TSavo:

**Verdict: REQUEST-CHANGES** (two real issues; the core verb is sound and the security-relevant paths are genuinely real, not faked)

I did not write this code. I read the mint+sign path, the solver-dispatch path, and the failure path, and I proved the load-bearing behaviors empirically by running the real `provekit verify` binary against purpose-built `.proof` catalogs with z3 4.15.4 on PATH.

### What is REAL (verified, not trusted)

**Ed25519 signing is real and the artifact verifies.** `ed25519_sign_string` / `ed25519_pubkey_string` (`provekit-proof-envelope/src/sign.rs`) use genuine `ed25519_dalek::SigningKey`. I minted a witness through the binary, read it off disk, stripped `cid`+`signature`, re-JCS'd the remaining content, and called `ed25519_verify_string(signed_by, signature, jcs)` -> **`true`**. The file `cid` also equals `blake3-512(signable_jcs)`. So the signature is cryptographically valid and the CID binds the signed content. Not stubbed.

**Solver dispatch is real.** `verify_one_claim` (`cmd_verify.rs:384`) calls the existing `run_plan(plan, registry, &smt, Some(&obligation))` primitive. The verdict comes from `s.solve(&input)` (a real subprocess), not a hardcoded `Ok`. `SubprocessSolver` maps `unsat -> Discharged`, `sat -> Unsatisfied` (`subprocess.rs:182-183`) — correct SMT validity-by-negation semantics. The witness cites the discharging solver via `invs.iter().find(|i| i.authoritative)` -> `z3@4.x` (`cmd_verify.rs:387-391`), recorded in `measurements.solver`.

**Failure detection is real — proven end-to-end.** I built a `.proof` catalog with a violated contract (`pre: forall n:Int. n > n`), ran the binary:
```
"status": "unsatisfied", "pass": false,
"reason": "solver 'z3' returned sat (counterexample found)",
"dischargingSolver": "z3@4.x"   EXIT CODE 1 (EXIT_VERIFY_FAIL)
```
So a function whose body does not satisfy its contract IS caught. The valid tautology case discharges (exit 0). The machinery is correct in both directions.

**Runner reuse confirmed.** `cmd_verify` calls the verifier's public stages (`load_all_proofs`, `enumerate_callsites`, `resolve_target`, `instantiate`, `smt_emitter`, `run_plan`) — it does not reimplement a parallel Runner. **`prove` is not broken**: `main.rs:416` keeps `Cmd::Prove(a) => cmd_prove::run(a)` with its own `--kit` conformance handler intact; `verify` is cleanly split off.

**Descope-clean.** No dependency on body-cycle machinery (no `provekit_walk` body-emit, no `realize_rust_core`, no `RewriteShape`). Imports are solver/obligation primitives + the signing helper only.

### REQUEST-CHANGES #1 — No negative test (the brief explicitly required this)
The negative path is real but **untested in this PR**. The only `pass == false` assertion (`cmd_verify_integration.rs:283`) is in the *z3-absent graceful-degradation* branch — that is `Undecidable` (no solver), NOT `Unsatisfied` (a violated claim). There is no test asserting that a genuinely-violated contract yields `status: unsatisfied` / `pass: false` / exit 1. I proved that behavior works manually; it should be a committed regression test. Per the descope principle "tests must actually test," shipping a GATE verb whose entire reason-for-being is catching bad claims, with zero coverage of the catch, is the gap to close. Add a fixture with a `sat` obligation (e.g. `forall n. n > n`, which emits cleanly) and assert the failure receipt + exit 1.

### REQUEST-CHANGES #2 — Hardcoded signer key with a misleading comment (trust-signal issue for a signing GATE verb)
`const VERIFY_SIGNER_SEED: [u8; 32] = [0x76; 32];` (`cmd_verify.rs:69`) is used unconditionally at `:439` and `:472`. The comment claims "production minting overrides via the provenance key path," but **no override exists** — no CLI flag, no env var, no vault/provenance-file read. (The `RUN_SIGNER_SEED` it claims to mirror is also hardcoded with no override.) For a verb that *mints and signs* witnesses, a public hardcoded dev key means the signature attests "this binary ran verify," not "this operator signed off" — a misleading trust signal. Either wire a real override (env/flag/provenance key) or change the comment to state plainly that the witness signer is a fixed non-secret build key and the signature is an integrity/provenance tag, not an authority attestation.

### Nit (non-blocking)
`build_dispatch_plan_and_registry` (`cmd_verify.rs:307-310`) only honors a kit's dispatch table when `SolverPlan::from_config` returns `Dispatch`. A kit declaring `Chain`/`Portfolio`/`Single` is **silently replaced** by the default table, contradicting the "kit author's table wins" comment. Either honor non-Dispatch kit plans or correct the comment.

Solid, genuinely-real keystone work. The two changes above are about test coverage of the failure case and honesty of the signing trust model — both squarely in this PR's wheelhouse.

TSavo:
Review fixes landed in `62f633760`. Thanks for the empirical pass — all three confirmed and addressed.

**CHANGE 1 — committed negative test.** Added `verify_violated_claim_fails_and_mints_no_witness`. It builds a kit whose target contract claims `forall n:Int. n > n` (false for every n); `provekit verify` routes it to z3, which returns sat-for-the-negation = `unsatisfied`. The test asserts:
- `claim.status == "unsatisfied"`, `claim.pass == false`
- exit code `1` (`EXIT_VERIFY_FAIL`), distinct from the solver-miss exit `3`
- **no witness minted** — `claim.witnessCid` is null AND the witness dir is empty on disk

The catch path now ships tested, not just manually reproduced. The fixture builder was refactored into `publish_claim_project(suffix, name, target_pre_body)` shared by the positive (`n >= n`) and negative (`n > n`) cases.

**CHANGE 2 — signer key honesty.** Both parts done:
- (a) Real override hook `resolve_signer_seed()`: `PROVEKIT_VERIFY_SIGNER_KEY` (hex 32-byte seed) or `PROVEKIT_VERIFY_SIGNER_KEY_FILE` (path to one). Falls back to the dev seed only when neither is set; a malformed override **fails closed** (the run errors rather than silently signing with the dev seed). New unit test covers hex decode (valid / `0x`-prefixed / wrong-length / non-hex).
- (b) Comment rewritten: the dev seed is an **integrity tag** (binds bytes, reproducible CID, tamper-evident), **not an authority attestation** (the seed is well-known). Witnesses now carry `measurements.signer_attests_authority` (true only under an override), and the human output prints a note when signing with the dev seed.

**NIT — kit plan honesty.** `build_plan_and_registry` now honors the kit author's declared `[solvers]` plan **verbatim** (`Dispatch` / `Chain` / `Portfolio` / `Single`). The default dispatch table is synthesized **only** when the kit declares no solver config — so "kit author's plan wins" is now true. `routed_seat` reports the plan shape honestly for non-dispatch plans (`single:` / `chain:` / `portfolio:`), with `dischargingSolver` as the authoritative who-proved-it field; the human output names which plan is in use.

**Verification:** 4 unit + 3 integration tests green (incl. the new negative fixture and the new signer-seed test); `provekit-verifier` suite green; `cargo build -p provekit-cli` clean.

Issue #1440, opened 2026-05-24
[Descope v1] PR-21: bridge-writer — promote @sugar/@boundary into bridge members (the harvest→verify seam)


TSavo:
## The gap (surfaced by gauntlet #1436)

The v1 build PRs shipped the endpoints — harvest, verify, emit — but NOT the connective tissue between harvest and verify. No verb writes the \`bridge\` member that connects a harvested/@sugar-claimed contract to a verifiable obligation. Result: \`enumerate_callsites\` returns 0, \`verify\` reports \`totalClaims: 0\` (vacuously ok). Proven empirically on merged main against the canonical blake3 shim (zero bridges, verify finds nothing).

## What a bridge IS (the spec)

A bridge is the content-addressable node at the @boundary seam — where one realization's territory ends and the contract takes over. It binds:
- \`sourceSymbol\` = the @sugar function / @boundary point
- \`targetContractCid\` = the concept's contract (obligation in its \`pre\` slot)
- \`targetProofCid\` = the proof bundle

**Proofchain semantics:** because the bridge's CID includes \`targetProofCid\` (which includes prior witness CIDs), the bridge's hash transitively commits to every verification beneath it. The bridge is the proofchain rollup point — "this boundary is satisfied" becomes one content-addressed fact carrying its whole proof history. Verify discharges → witness; bridge binds witness+contract+symbol; the next bridge up rolls that CID into its own. k(I)=t composing.

## Where it lives

\`core::bind\` (BindKit) — KEPT in the descope (Q-1, #1417) specifically as the gap-discharge/promotion machinery. Minting bridges from \`@sugar(concept=X)\` / \`@boundary\` is exactly its job. Keeps \`verify\` focused on discharge.

## The work

1. Bind reads \`@sugar(concept=X)\` / \`@boundary\` annotations (already lifted to library-sugar-binding-entry / realization-memento).
2. Looks up concept X's contract in the catalog; ensures the obligation is in the \`pre\` slot (the harvester currently puts the source observation in \`inv\` — bind needs to produce the \`pre\`-bearing target contract verify can discharge).
3. Mints a \`bridge\` member: \`sourceSymbol\`=fn, \`targetContractCid\`=concept contract CID, \`targetProofCid\`=bundle CID.
4. \`enumerate_callsites\` now finds it; \`verify\` discharges; witness rolls into the bridge's content address.

## Cross-cutting

[...]

Issue #1436, opened 2026-05-23
[Descope v1] RETIREMENT GATE: end-to-end acceptance gauntlet (rust + java + python, positive + negative)


TSavo:
## The gate

**Nothing in the retire phase (PR-2, PR-3, PR-14, PR-15, PR-16, PR-17, PR-18) merges until this gauntlet is green.** This is the new "10/10 cycle" for the descoped product — it proves the four pillars work together as one integrated chain, not just as six PRs that each pass isolated unit tests.

## The chain (per language: rust, java, python)

1. **Harvest** — a real unit test / macro / annotation in the language, carrying an assertion. The harvester (`provekit-lift-<lang>-tests` / junit / pytest) lifts it into a contract (predicate in the catalog).
2. **Assign to sugar** — a function annotated to claim it satisfies that contract (`#[provekit::sugar(concept=...)]` or per-language equivalent).
3. **Realize at a @boundary** — consuming code marks a `@boundary` that inherits the contract. Contract propagates.
4. **Verify (positive)** — `provekit verify <kit>` discharges the obligation via solver → mints a signed witness → emits the gate (the native test: junit/pytest/cargo-test), which runs GREEN.
5. **Verify (negative)** — a deliberately-broken realization whose body VIOLATES the contract. `provekit verify` CATCHES it: no witness minted (or witness marked failed), the emitted gate FAILS loudly. The negative case is the proof that verify isn't vacuously passing everything.

## Deliverables — 6 demonstrations

| # | Language | Case | Must produce |
|---|---|---|---|
| 1 | rust | positive | contract from test/macro + sugar + @boundary + signed witness + green cargo-test gate |
| 2 | rust | negative | broken realization → verify fails → gate red |
| 3 | java | positive | contract from junit + sugar + @boundary + signed witness + green junit gate |
| 4 | java | negative | broken realization → verify fails → gate red |
| 5 | python | positive | contract from pytest + sugar + @boundary + signed witness + green pytest gate |
| 6 | python | negative | broken realization → verify fails → gate red |

## Acceptance criteria

- Each demonstration is a checked-in example kit + a script/test that runs the full chain and asserts the outcome.
- Positive cases: witness minted + signed, native gate green, receipt shows PASS.
- Negative cases: verify reports the specific contract violation, native gate red, NO valid witness, receipt shows FAIL with the violated predicate named.
- All six runnable from a single \`make gauntlet\` (or equivalent) target so CI can gate on it.

## Depends on (must be merged first)

[...]

PR #1441, opened 2026-05-24
verify: discharge harvested obligations through the function body (#1440)

TSavo:
## Summary

Wires the **verification spine**: `verify` now proves a function BODY satisfies a harvested contract, both ways. Closes the #1440 gap where the wp evaluator existed but was unreachable, so z3 saw the user function as an uninterpreted `Ctor` -> "unknown constant" -> **Undecidable for both positive and negative**.

### The three deltas (NEW-BUILD on top of the existing `libprovekit::wp`)

1. **Ctor->Op at the wp boundary.** `verify` reconstructs a harvested call as a `core::Term::Op` (via `From<IrTerm>`) so wp's `wp_op` dispatches on it. The harvester and `enumerate_callsites` keep the JSON-ctor flow (the production path locked by the new regression test).
2. **Body-derived op-contract + catalog resolver.** `body_discharge::CatalogResolver` is the first non-test `OpContractResolver`: op name -> bridge by symbol -> `targetContractCid` -> contract memento -> projects its **body-derived `post`** (`result == *(x,2)`, lifted by walk's `lift_function_postcondition` -- the **lift-half**, NOT the cycle/lower/carrier machinery) + `formals` into an `OpContractInfo`. Non-body-derived contracts return `None` and fall through to the existing refinement path untouched.
3. **Bridge writer in `core::bind`.** `bind_function_bridge` takes a `FunctionContractMemento` and emits both the body-derived op-contract member and the bridge (`sourceSymbol -> targetContractCid -> targetProofCid`). The bridge CID commits to the op-contract CID (proofchain rollup).

### Load-bearing shape fix (flagged for review)

`enumerate_callsites` was **v1.1-flat only**, but the production harvest path (`mint_contract`) emits **v1.2-layered** mementos -- so harvested calls never enumerated. Made it shape-agnostic via `memento_kind`/`memento_body`, matching what `resolve_target` already does. Without this the spine can't see harvested contracts at all.

### The proof (committed, z3-gated: `cmd_verify_body_discharge.rs`)

- **Positive**: `double(x)=x*2`, `assert_eq!(double(3),6)` -> wp reduces to `*(3,2)==6` -> z3 discharges -> pass, signed witness minted.
- **Negative**: break body to `x*3` -> `*(3,3)==6` -> z3 finds `9!=6` SAT on the negation -> **Unsatisfied, exit 1, no witness**. Fails HONESTLY (Unsatisfied, not Undecidable) because no uninterpreted symbol survives the reduction.

Plus the regression the harvester round-trip test was missing: `enumerate_callsites` is empty before bind and **non-empty after harvest+bind**.

## Test plan

- [x] `cmd_verify_body_discharge`: positive discharges + signed witness; negative is Unsatisfied + exit 1 + no witness
- [x] `cmd_verify_integration` (the existing v1.1 LIA + violated tests) still pass (body path falls through cleanly)
- [x] `round_trip` regression: enumerate empty before bind, non-empty after
- [x] `bind_function_bridge` unit test (op-contract + pinned bridge, CID rollup)
- [x] libprovekit (101) + verifier suites green

## Pre-existing failure (not this branch)

`libprovekit` `d7_v1_source_round_trip::value_null_source_round_trip_receipt_is_complete` fails on the base commit `95b9632fd` too (byte-identical realize, the cycle machinery being retired). Verified by stashing this branch's changes.

ref #1440

Generated with Claude Opus 4.7 (1M context)

## Release Notes

* **New Features**
  * Added fast-path verification for function contracts using body-discharge optimization. The verifier now efficiently handles claims about function calls when they match recognized contract patterns.

* **Tests**
  * Added comprehensive end-to-end tests validating body-discharge verification, including positive cases with successful verification and negative cases confirming proper failure handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1441?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1985,6 +1985,7 @@ dependencies = [
 name = "provekit-lift-rust-tests"
 version = "0.1.0"
 dependencies = [
+ "libprovekit",
  "proc-macro2",
  "provekit-canonicalizer",
  "provekit-ir-symbolic",
--- a/implementations/rust/libprovekit/src/core/bind.rs
+++ b/implementations/rust/libprovekit/src/core/bind.rs
@@ -680,6 +680,154 @@ fn bind_response_contract(payload: &Json, payload_cid: &Cid) -> Contract {
     )
 }
 
+// ============================================================
+// Bridge writer (PR-22, #1440).
+// ============================================================
+//
+// A harvested call `double(3)` is dischargeable through verify ONLY when
+// three things are in the pool:
+//   1. a body-derived op-contract for `double` (post = result == *(x, 2)),
+//   2. a bridge `sourceSymbol "double" -> targetContractCid <op-contract>`,
+//   3. (transitively) the proof bundle the bridge pins.
+// The bridge's CID commits to the op-contract CID (it is in `inputCids`),
+// and the bundle commits to both; that is the proofchain rollup.
+//
+// This writer is the production source of (1) and (2): it takes the
+// `FunctionContractMemento` walk already builds for a function (which
+// carries `fn_name`, `formals`, `formal_sorts`, and the BODY-DERIVED
+// `post`), and emits both member envelopes in the v1.1-flat
+// `evidence.body` shape that `enumerate_callsites` / `resolve_target` /
+// `body_discharge::CatalogResolver` consume. The op-contract member CID is
+// the bridge's `targetContractCid`, so verify resolves the chain.
+
+/// One v1.1-flat member envelope plus its re-derivable member CID
+/// (`blake3_512(JCS(envelope))`, the same identity `load_all_proofs`
+/// recomputes for a member with no `cid` / `producerSignature` field).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BridgeMember {
+    /// The member CID = `blake3-512:<hex>`.
+    pub cid: Cid,
+    /// The member envelope JSON (no `cid` / `producerSignature` — those are
+    /// added by the proof-envelope builder when the bundle is assembled).
+    pub envelope: Json,
+}
+
+/// The pair of members a function bind produces for body-discharge: the
+/// body-derived op-contract and the bridge that points a harvested call at
+/// it.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FunctionBridgeMembers {
+    /// The body-derived op-contract member (carries `formals` + `post`).
+    pub op_contract: BridgeMember,
+    /// The bridge member (`sourceSymbol -> targetContractCid`).
+    pub bridge: BridgeMember,
+}
+
+/// Re-derive a v1.1-flat member CID for an envelope, matching
+/// `provekit_verifier::load_all_proofs::compute_member_cid`: strip any
+/// `cid` / `producerSignature`, JCS-encode, BLAKE3-512.
+fn flat_member_cid(envelope: &Json) -> Cid {
+    let mut stripped = envelope.clone();
+    if let Json::Object(map) = &mut stripped {
+        map.remove("cid");
+        map.remove("producerSignature");
+    }
+    let canonical = provekit_canonicalizer::encode_jcs(&json_to_canonical_value(&stripped));
+    Cid::from_hash_output(blake3_512_of(canonical.as_bytes()))
+}
+
+/// Convert a `serde_json::Value` into the canonicalizer's `Value` so the
+/// JCS bytes line up with every other minter in the tree.
+fn json_to_canonical_value(j: &Json) -> std::sync::Arc<provekit_canonicalizer::Value> {
+    use provekit_canonicalizer::Value as CV;
+    match j {
+        Json::Null => CV::null(),
+        Json::Bool(b) => CV::boolean(*b),
+        Json::Number(n) => CV::integer(n.as_i64().unwrap_or(0)),
+        Json::String(s) => CV::string(s.clone()),
+        Json::Array(items) => CV::array(items.iter().map(json_to_canonical_value).collect()),
+        Json::Object(map) => CV::object(
+            map.iter()
+                .map(|(k, v)| (k.clone(), json_to_canonical_value(v)))
+                .collect::<Vec<_>>(),
+        ),
+    }
+}
+
+/// Build the body-discharge members (op-contract + bridge) for a function
+/// contract. `target_proof_cid` is the `.proof` bundle CID the bridge pins
+/// (`BridgeDeclaration.targetProofCid`); pass the CID of the bundle these
+/// members will be assembled into. `source_layer` / `target_layer` name
+/// the language axes (e.g. `"rust"` -> `"rust-kit"`).
+///
+/// The op-contract carries the function's BODY-DERIVED `post` (the
+/// `FunctionContractMemento`'s `post`, which walk lifts from the body's
+/// trailing expression), plus `formals` / `formalSorts` so the resolver
+/// can name the value slots. This is the lift-half of walk (verification
+/// substrate); it does NOT touch the lower/cycle/carrier machinery.
+pub fn bind_function_bridge(
+    contract: &Contract,
+    source_layer: &str,
+    target_layer: &str,
+    target_proof_cid: Option<&str>,
+) -> Result<FunctionBridgeMembers, BindError> {
+    let post_json = serde_json::to_value(&contract.post)
+        .map_err(|e| BindError::Failed(format!("serialize body-derived post: {e}")))?;
+    let formals: Vec<Json> = contract
+        .formals
+        .iter()
+        .map(|f| Json::String(f.clone()))
+        .collect();
+    let formal_sorts: Vec<Json> = contract
+        .formal_sorts
+        .iter()
+        .map(|s| serde_json::to_value(s).unwrap_or(Json::Null))
+        .collect();
+
+    let op_contract_env = json!({
+        "evidence": {
+            "kind": "contract",
+            "body": {
+                "contractName": contract.fn_name,
+                "formals": formals,
+                "formalSorts": formal_sorts,
+                "post": post_json
+            }
+        }
+    });
+    let op_contract_cid = flat_member_cid(&op_contract_env);
+
+    let mut bridge_body = serde_json::Map::new();
+    bridge_body.insert("sourceSymbol".into(), Json::String(contract.fn_name.clone()));
+    bridge_body.insert("sourceLayer".into(), Json::String(source_layer.to_string()));
+    bridge_body.insert(
+        "targetContractCid".into(),
+        Json::String(op_contract_cid.as_str().to_string()),
+    );
+    bridge_body.insert("targetLayer".into(), Json::String(target_layer.to_string()));
+    if let Some(tpc) = target_proof_cid {
+        bridge_body.insert("targetProofCid".into(), Json::String(tpc.to_string()));
+    }
+    let bridge_env = json!({
+        "evidence": {
+            "kind": "bridge",
+            "body": Json::Object(bridge_body)
+        }
+    });
+    let bridge_cid = flat_member_cid(&bridge_env);
+
+    Ok(FunctionBridgeMembers {
+        op_contract: BridgeMember {
+            cid: op_contract_cid,
+            envelope: op_contract_env,
+        },
+        bridge: BridgeMember {
+            cid: bridge_cid,
+            envelope: bridge_env,
+        },
+    })
+}
+
 fn bind_lift_entries(term_json: &Json) -> Result<Vec<BindLiftEntry>, BindError> {
     if term_json.get("kind").and_then(Json::as_str) == Some("named-term-document") {
         return Err(BindError::Failed(
@@ -1824,6 +1972,90 @@ fn primitive_sort(name: &str) -> Sort {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// A `double(x) = x*2` function contract, as walk's
+    /// `build_function_contract` would produce: body-derived
+    /// `post = (result == *(x, 2))`, one formal `x`.
+    fn double_contract() -> Contract {
+        use crate::compose::{EffectSet, Locus};
+        let post = IrFormula::Atomic {
+            name: "=".to_string(),
+            args: vec![
+                IrTerm::Var {
+                    name: "result".to_string(),
+                },
+                IrTerm::Ctor {
+                    name: "*".to_string(),
+                    args: vec![
+                        IrTerm::Var {
+                            name: "x".to_string(),
+                        },
+                        IrTerm::Const {
+                            value: json!(2),
+                            sort: primitive_sort("Int"),
+                        },
+                    ],
+                },
+            ],
+        };
+        Contract {
+            fn_name: "double".to_string(),
+            formals: vec!["x".to_string()],
+            formal_sorts: vec![primitive_sort("Int")],
+            formal_regions: vec![],
+            return_sort: primitive_sort("Int"),
+            return_region: None,
+            pre: IrFormula::Atomic {
+                name: "true".to_string(),
+                args: vec![],
+            },
+            post,
+            body_cid: None,
+            effects: EffectSet::empty(),
+            locus: Locus::default(),
+            canonical_bytes: vec![],
+            cid: "blake3-512:test".to_string(),
+            auto_minted_mementos: vec![],
+            concept_hint: None,
+        }
+    }
+
+    #[test]
+    fn bind_function_bridge_emits_op_contract_and_pinned_bridge() {
+        let contract = double_contract();
+        let members =
+            bind_function_bridge(&contract, "rust", "rust-kit", Some("blake3-512:bundle"))
+                .expect("bind_function_bridge");
+
+        // Op-contract carries the body-derived post + formals.
+        let oc = members
+            .op_contract
+            .envelope
+            .pointer("/evidence/body")
+            .unwrap();
+        assert_eq!(oc.get("contractName").unwrap(), "double");
+        assert_eq!(oc.get("formals").unwrap(), &json!(["x"]));
+        // post is `result == *(x, 2)`.
+        let value_expr = oc.pointer("/post/args/1").unwrap();
+        assert_eq!(value_expr.get("name").unwrap(), "*");
+
+        // The bridge points at the op-contract member CID and pins the bundle.
+        let br = members.bridge.envelope.pointer("/evidence/body").unwrap();
+        assert_eq!(br.get("sourceSymbol").unwrap(), "double");
+        assert_eq!(
+            br.get("targetContractCid").unwrap().as_str().unwrap(),
+            members.op_contract.cid.as_str(),
+            "bridge.targetContractCid must equal the op-contract member CID (proofchain rollup)"
+        );
+        assert_eq!(br.get("targetProofCid").unwrap(), "blake3-512:bundle");
+
+        // CIDs are content-addressed and stable.
+        assert_eq!(
+            members.op_contract.cid,
+            super::flat_member_cid(&members.op_contract.envelope)
+        );
+        assert!(members.op_contract.cid.as_str().starts_with("blake3-512:"));
+    }
 
     #[test]
     fn bind_names_lifted_entries_without_plugin_dispatch() {
--- a/implementations/rust/provekit-cli/src/cmd_verify.rs
+++ b/implementations/rust/provekit-cli/src/cmd_verify.rs
@@ -464,6 +464,27 @@ fn verify_one_claim(
             // refinement obligation: instantiate the resolved pre with the
             // call's arg term (the same obligation the prover builds).
             let Some(_pre) = resolved.ir_formula.as_ref() else {
+                // HONESTY BOUNDARY (catches every body-bearing variant). The
+                // vacuous-discharge shortcut is legitimate ONLY for a
+                // genuinely non-body-bearing target. A target carrying
+                // `formals` is body-bearing by definition: its `post`
+                // describes a body whose obligation we reached here WITHOUT
+                // reducing (e.g. `extract_body_obligation` could not build the
+                // obligation because the body-derived `post` was not a
+                // `result == <expr>` equation, so the resolver dropped it).
+                // Vacuous-passing it would be a false green. Refuse instead:
+                // leave the default Undecidable verdict (not discharged, not
+                // pass, no witness).
+                if resolved.target_is_body_bearing {
+                    result.reason = format!(
+                        "body-discharge: refuse: target `{}` is body-bearing \
+                         (carries `formals`) but its obligation was not reduced \
+                         and it has no precondition; refusing rather than \
+                         reporting a vacuous pass",
+                        cs.bridge_ir_name
+                    );
+                    return result;
+                }
                 result.verdict = ObligationVerdict::Discharged;
                 result.reason = "vacuous: target carries no precondition".to_string();
                 result.obligation_class = "vacuous".to_string();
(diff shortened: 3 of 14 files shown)


TSavo:
## Spine is language-agnostic; java harvest does not yet feed it (gauntlet #1436 follow-up)

Took the verification spine this PR landed and ran the java leg of the #1436 gauntlet (`int doubleIt(int x){ return x*2; }` + junit `assertEquals(6, doubleIt(3))`). Confirming the spine is genuinely language-agnostic -- and locating exactly where java's harvest fails to produce the body-obligation shape `wp` consumes. Full write-up on #1436; the spine-relevant summary:

**The seam consumes (from this PR):** a `kind:"contract"` memento with `body.formals` + `body.post` shaped `=(var "result", <value-expr>)` (`body_discharge.rs:54,89,108-113,129`), plus a `sourceSymbol -> targetContractCid` bridge that `enumerate_callsites` matches against the harvested call ctor. `CatalogResolver` (the first non-test `OpContractResolver`) projects `formals`+`post` into `OpContractInfo`; `wp` inlines the value-expr; z3 discharges. None of that is rust-specific -- it is bytes in, obligation out.

**Java already has the content, not the wiring.** `JavaSourceLifter` derives, for `return x*2`, a body obligation `post = (return_value == *(x,2))` with `formals:["x"]` -- structurally what the seam wants. But three production-path breaks stop it reaching the resolver:

1. **Result var.** java emits `return_value` (`JavaSourceLifter.java:335`); seam requires `result` (`body_discharge.rs:54`, enforced by `value_expr()?` at `:129`). -> `lookup` returns `None` -> uninterpreted callee -> Undecidable, i.e. the same #1440 wall this PR fixed for rust, re-appearing for java because java's post never reaches `wp` in the right shape.

2. **`formals` lost at mint.** `MintContractArgs` has no `formals` field (`provekit-claim-envelope/src/lib.rs:234`); `cmd_mint` doesn't carry it (`cmd_mint.rs:1027`). The sealed `kind:"contract"` memento (`claim-envelope:455`) has `post` but no `formals` -> `CatalogResolver::lookup` returns `None` (`body_discharge.rs:108-113`).

3. **No bridge writer wired.** `bind_function_bridge` (`libprovekit/src/core/bind.rs:768`) -- the writer this PR added that emits the op-contract member + `sourceSymbol -> targetContractCid` -- is called only from `provekit-lift-rust-tests/tests/round_trip.rs` and its own unit test. `cmd_bind.rs` references it 0 times. Rust's round-trip reaches the spine via test scaffolding, not a production verb. So **no language** currently has a production harvest->bridge->verify path; rust's green is via the round_trip test calling `bind_function_bridge` by hand.

**Takeaway for this PR:** the spine does not need changes -- the language-agnostic claim holds. The missing piece is a production sealing+bridging verb (extend `cmd_bind` to consume `function-contract` decls and call `bind_function_bridge`, or auto-emit the bridge at mint), plus carrying `formals` through mint and renaming java's `return_value` -> `result`. Until then gauntlet rows 3 & 4 (java) cannot go green honestly.

Also flagged on #1436: `JavaSourceLifter.java:327` multi-statement fallback (`postValue = body`) transcribes the whole method body into ProofIR, which violates the terminology lock (body stays java; only the contract lifts). The single-return path (`:351`) is compliant. The fallback should refuse, not transcribe.

No fake green demo, no revived cycle/carrier machinery. The gauntlet surfaced a real gap.

T Savo

TSavo:
## Independent review (fresh eyes) — keystone

**Verdict: REQUEST-CHANGES** (posted as a comment; gh blocks formal request-changes on a self-authored PR). The wp mechanism is real and the positive/negative pair is honest. But the fall-through is **not** honest: I built a reachable case where the verifier silently PASSES a claim it cannot check. That is the cardinal sin for this node, and it blocks merge. The fix is local and mechanical.

I checked out the branch, read the wp path, ran the committed tests, and built a third probe. Evidence below per point.

---

### BLOCKER — Point 6 (fall-through honesty): the verifier silently passes an unverifiable claim

The PR says "anything outside the recognized `=(<call>, <expected>)` shape falls through honestly." It does not. The fall-through routes a body-bearing claim into the **vacuous-discharge** branch and returns a green pass.

`extract_body_obligation` (`body_discharge.rs`) conflates two distinct `Ok(None)` cases:
- **line 163-165**: callee has no body-derived contract -> genuinely fall through. Correct.
- **line 170-172**: callee HAS a body-derived contract, but the assertion shape is not `=(call, expected)` -> also returns `Ok(None)`. **Wrong.**

In the second case, `verify_one_claim` (`cmd_verify.rs:447+`) falls to `resolve_target::run`, which reads the bridge target = the body-derived op-contract. That contract has `post` + `formals` but **no `pre`** (`resolve_target.rs:67` reads `body.get("pre")` -> `None`), so the `let Some(_pre) = resolved.ir_formula ... else { Discharged; "vacuous"; }` branch fires and the claim passes.

**Empirical probe** (same project shape as the committed positive test; only the predicate changed from `=` to `<=`, i.e. `<=(double(3), 6)`):

```
status: "discharged", pass: true, reason: "vacuous: target carries no precondition"
exit code: 0, ok: true
```

A body-derived contract with no `pre` is not a vacuous tautology — its only checkable claim is body-discharge. If the spine could not extract the body obligation, the claim is **unverified, not discharged**. Reachability is trivial: any bound function + any non-`assert_eq!` assertion (`assert!(double(x) < 10)`, `>= 0`, etc.).

**Fix** (either is fine):
- (a) Add a `Refused`-style variant `extract_body_obligation` returns when a body-derived contract is present but the shape is unrecognized; map it to a non-pass verdict (`Undecidable`/`Refused`) in `verify_one_claim`. OR
- (b) In the fall-through, detect a body-derived target (`formals` present, `pre` absent) and refuse instead of firing the vacuous branch.

Please also add a regression test asserting `pass == false` (or `status != "discharged"`) for the body-derived-contract + non-`=` shape, or this hole reappears.

---

### The other 7 points: confirmed REAL

**1. Body-discharge is real (not faked).** `wp_op` (`wp.rs:430`) looks up the contract, synthesizes the value-op rule `Q[result := value_expr]`, and `value_expr_of_term` (`wp.rs:508`) recursively inlines the callee body. For `double(x)=x*2`, `wp(double(3), result==6)` -> `*(3,2)==6`; no uninterpreted `double` survives. Ran the committed positive test: `POSITIVE_WITNESS_CID=blake3-512:5b6ee6e6...`, status `discharged`, solver `z3@`, signed witness file written, exit 0.

**2. Negative is real and honest.** Same path, `body_factor=3` -> `*(3,3)==6` -> z3 SAT-on-negation (`9 != 6`). Ran the committed negative test: `NEGATIVE_EXIT_CODE=1 STATUS="unsatisfied"`, `pass=false`, `witnessCid=null`, witness dir empty. Unsatisfied (not Undecidable) because no uninterpreted symbol survives the reduction. Correct.

**3. Resolver is catalog-backed, not a disguised MapResolver.** `CatalogResolver::target_contract_body` (`body_discharge.rs:79-93`) reads `pool.bridges_by_symbol` and `pool.mementos`, follows `targetContractCid` to a real contract memento. No hardcoded map.

**4. Bridge CID commits to the op-contract CID.** `bind_function_bridge` (`bind.rs:768-829`) computes `op_contract_cid = flat_member_cid(&op_contract_env)`, inserts it into `bridge_body["targetContractCid"]`, THEN computes `bridge_cid = flat_member_cid(&bridge_env)` over bytes that include it. Genuine content-addressed commitment, not a sidecar.

**5. Scope discipline held.** `git diff --stat` touches only the verify path (`cmd_verify.rs`, `body_discharge.rs`, `enumerate_callsites.rs`, `types.rs`, `lib.rs`), the lift-half bridge writer (`bind.rs`), and tests. Grepping `+` lines for `cycle|lower|carrier` returns only doc-comments explicitly DENYING those touches. The retired body-cycle/lower/carrier machinery is untouched.

**7. Signing reuses PR-9's path.** The diff to `cmd_verify.rs` is a single import line; `resolve_signer_seed` (`cmd_verify.rs:99+`) with the `PROVEKIT_VERIFY_SIGNER_KEY` / `_FILE` override + dev-seed integrity-tag fallback + `is_authoritative` flag is unchanged. No new hardcoded key.

**8. Pre-existing failure confirmed pre-existing.** `value_null_source_round_trip_receipt_is_complete` fails on this branch AND identically on a fresh detached worktree of base `95b9632fd` (`d7_v1_source_round_trip.rs:410`). Not introduced. `cargo build --workspace` green (0 errors); `provekit-verifier` suite all green.

---

**Bottom line:** positive discharge, negative honest-failure, and catalog-backed resolver are all REAL. The spine works. But the unrecognized-shape fall-through silently discharges as vacuous, which is exactly the "verifier passes what it can't check" failure this node must not have. Fix point 6 and re-request review.

TSavo:
## False-green closed (commit `dbff9eb25`)

Thanks for catching this and for running z3 to confirm the mechanism. The blocker is fixed; everything else you reviewed clean is untouched.

### The bug, as you diagnosed it

`extract_body_obligation` returned `Ok(None)` for any non-`=(<call>, <expected>)` shape (body_discharge.rs:170-172) **even when the callee had a body-derived op-contract**. The claim fell through to `resolve_target` + `instantiate`, the body-derived op-contract (post + formals, NO `pre`) hit the `vacuous: target carries no precondition` branch, and verify reported `status:"discharged", pass:true, exit 0`. A green pass for a claim it never checked. I also found the **same hazard** in the `Err(WpError::Refused(..))` arm (a wp refusal on a nested call also returned `Ok(None)`).

### The fix

Once the callee resolves to a body-derived op-contract (past the `lookup` gate), **every** failure to build or reduce the obligation is now a REFUSAL (`Err`), never an `Ok(None)` fall-through. The caller maps `Err` to the default Undecidable verdict: not discharged, not pass, no witness, refusal reason on the row, exit 3 (not 0).

The honest fall-through is preserved for genuinely-non-body-bearing claims: a contract with **no `formals`** (the cross-language refinement targets in the LIA integration test) returns `None` from `lookup` BEFORE the gate, so those still take the refinement path. The existing `cmd_verify_integration` LIA + violated tests still pass, confirming the legitimate fall-through is intact.

### Regression test (your exact probe)

`verify_body_bearing_unrecognized_shape_refuses_not_vacuous_pass`: the positive fixture with `=` changed to `<=` (`<=(double(3), 10)`, callee still body-bearing). Asserts the claim is NOT discharged, does NOT pass, does NOT take the `vacuous` obligation class, surfaces a `body-discharge ... refuse` reason, mints no witness, and does not exit 0.

Fresh run:
```
FALSEGREEN_STATUS="undecidable" EXIT=3
```

Suites green after the fix: `cmd_verify_body_discharge` (3/3 -- positive discharge, negative unsatisfied/exit-1, this new refusal test), `cmd_verify_integration` (3/3), verifier suite, `round_trip` regression.

Built with Claude Opus 4.7 (1M context).

TSavo:
## Independent review (fresh eyes) — keystone: re-verify of dbff9eb25

**Verdict: REQUEST-CHANGES (still).** The fix closes the path I reported, but a sibling of the same false-green class is still open. I confirmed both empirically on dbff9eb25.

### The original probe IS closed (credit where due)
Ran the author's new regression test `verify_body_bearing_unrecognized_shape_refuses_not_vacuous_pass` (my `<=`-shape probe): `FALSEGREEN_STATUS="undecidable" EXIT=3`. The post-`lookup` honesty boundary in `extract_body_obligation` now turns every "callee is body-bearing but obligation unreducible" case into `Err` -> Undecidable (exit 3, not pass, no witness). The `WpError::Refused` arm is fixed too. Positive + negative still green (`POSITIVE_WITNESS_CID=blake3-512:25d5c6dc...`, `NEGATIVE_EXIT=1 unsatisfied`). Good.

### Residual BLOCKER — the false-green moved to the `lookup` gate, it didn't close

The honesty boundary was placed AFTER the `resolver.lookup(...).is_none()` gate (`body_discharge.rs:169`). But `CatalogResolver::lookup` ITSELF silently returns `None` for a contract that carries the body-derived marker `formals` but whose `post` is not a `result == <expr>` equation -- it bails at `body_discharge.rs:129` (`info.value_expr()?`). That `None` makes `extract_body_obligation` return `Ok(None)` -> fall through to `resolve_target` -> no `pre` found -> vacuous-discharge branch -> **false green**.

**Empirical probe** (target contract: `formals: ["x"]`, `post: <=(result, x)` -- a body-derived-looking contract with a non-equation post, no `pre`; harvested `=(double(3), 6)`):

```
obligationClass: "vacuous", status: "discharged", pass: true, reason: "vacuous: target carries no precondition", exit 0, ok: true
```

Same cardinal sin, narrower trigger. A contract with `formals` is a body-derived contract by the writer's own convention; a non-equation `post` (e.g. a boolean-returning fn `fn is_pos(x) -> bool { x > 0 }` whose post is `result == (x > 0)` could canonicalize to a form `value_expr()` doesn't recognize, or any legacy/partially-formed memento) routes to a green pass for a claim never checked. The invariant ("verifier never silently passes what it can't check") must hold for ALL inputs, not just what today's harvester emits -- the verifier's honesty cannot depend on upstream behaving.

**Suggested fix (consumer-side, defense-in-depth -- catches every variant in one place):**

> In `verify_one_claim`, before firing the vacuous-discharge branch (`let Some(_pre) = resolved.ir_formula.as_ref() else { ... Discharged; "vacuous"; }`), check whether the resolved target carries body-derived markers (`formals` present in the contract body). If so, REFUSE (Undecidable) rather than vacuous-pass: a body-derived contract with no `pre` is not a vacuous tautology, it's a contract whose only checkable claim is body-discharge that the spine didn't take.

Pair this with the existing post-`lookup` refusal inside `extract_body_obligation` -- the two together are belt-and-suspenders for the same invariant. Please also add a regression test for the `formals`-present + non-equation-`post` + no-`pre` case (I can hand over the probe fixture; it's a near-clone of the new regression test).

Everything else from the prior review (points 1-5, 7, 8) stands as confirmed-real; not re-litigating. This one residual path is the only blocker.

TSavo:
## False-green closed for ALL inputs this time (commit `3240ab101`)

Right call -- the false green had MOVED, not closed. The first fix put the honesty boundary after the resolver lookup gate, but `CatalogResolver::lookup` itself drops a body-bearing contract whose `post` is not a `result == <expr>` equation (the `value_expr()?` early-return), so it returned `None` -> `Ok(None)` -> fell through to `resolve_target` -> no `pre` -> vacuous-discharge -> green pass, exit 0. Same sin, narrower trigger. Confirmed your empirical repro.

### The fix (consumer-side, per your recommendation)

`resolve_target` now surfaces `target_is_body_bearing` -- a non-empty `formals` array on the resolved contract body, the canonical body-derived marker. The vacuous-discharge shortcut is gated on it in **both** consumers:

- `cmd_verify::verify_one_claim` (the `verify` verb path)
- `runner::work_one` (the six-stage `prove`/Runner path -- it had the **identical** `resolved.ir_formula.is_none()` -> Discharged hazard at runner.rs:836; the advisor's audit caught it, fixed in the same commit)

A body-bearing target reaching either vacuous branch without a reduced obligation now REFUSES (Undecidable), never vacuous-passes. The post-lookup refusal from the prior commit stays as belt-and-suspenders.

**Principle, now enforced for all inputs:** the "no precondition => vacuously true" shortcut is legitimate ONLY for genuinely non-body-bearing targets. A contract carrying `formals` is body-bearing by definition; an unreduced obligation on it is REFUSE.

### Not over-refusing

The legitimate non-body-bearing fall-through is preserved: a target with no `formals` (the cross-language refinement targets in the LIA integration test) has `target_is_body_bearing = false` and still takes the refinement / vacuous path. `cmd_verify_integration` (LIA discharge routes to z3 + violated claim exits 1) still pass -- 3/3.

### Regression test for the new trigger

`verify_body_bearing_non_equation_post_refuses_not_vacuous_pass`: body-bearing op-contract (`formals:["x"]`, no `pre`) with a non-equation `post` `<=(result, x)`. Asserts NOT discharged, NOT pass, NOT vacuous class, refusal reason surfaced, no witness, exit != 0.

Fresh run (both probes refuse, original behaviors intact):
```
FALSEGREEN2_STATUS="undecidable" EXIT=3   <- new trigger (non-equation post)
FALSEGREEN_STATUS="undecidable"  EXIT=3   <- original probe (<= shape)
NEGATIVE_EXIT_CODE=1 STATUS="unsatisfied" <- broken body still fails honestly
```

Suites green: `cmd_verify_body_discharge` 4/4, `cmd_verify_integration` 3/3, verifier suite (including resolve_target / instantiate / runner), `round_trip` regression. Full `cargo build --workspace` clean.

A note for future contract authors is in the `target_is_body_bearing` doc-comment: `resolve_target::run` is the single setter; a new body-bearing contract shape MUST be recognized there to keep the honesty boundary complete.

Built with Claude Opus 4.7 (1M context).

TSavo:
## Independent review (fresh eyes) — keystone: third pass on 3240ab101

**Verdict: APPROVE-WITH-NITS.** Both real-world false-green classes are empirically closed. There is one remaining trigger, but it is reachable only by a hand-crafted memento or a writer that does not exist today, and the fix is one character. I would not block on it. Approving.

### Both prior probes now CLOSED (empirical, on 3240ab101)

Re-ran both false-greens I pinned in passes 1 and 2:

- **Probe A** (assertion-shape `<=(double(3), 10)`, the pass-1 finding): `status:"undecidable", pass:false, exit 3, 0 witness files`. Refused at the post-`lookup` boundary in `extract_body_obligation`.
- **Probe B** (target `formals:["x"]` + non-equation `post:<=(result,x)` + no `pre`, the pass-2 residual): `status:"undecidable", pass:false, exit 3, 0 witness files`. Refused at the new consumer-side `target_is_body_bearing` gate; reason explicitly cites the body-bearing refusal.

Author's two regression tests both green: `verify_body_bearing_unrecognized_shape_refuses_not_vacuous_pass` and the new `verify_body_bearing_non_equation_post_refuses_not_vacuous_pass`. Positive still mints a witness; negative still `unsatisfied`/exit 1.

### Second consumer (`runner::work_one`) confirmed closed

`work_one` (runner.rs:836) was the consumer the first review missed. It now has the identical `target_is_body_bearing` gate before its vacuous branch (runner.rs:844). I traced the surface:
- `ResolvedProperty` is constructed in exactly ONE place (`resolve_target.rs:68`) -> `target_is_body_bearing` has a single setter.
- `resolve_target::run` has exactly TWO callers (`cmd_verify.rs:454`, `runner.rs:824`) -> no third consumer.
- Functional vacuous-discharge sites are exactly two, both now gated.

No over-refusal: `cmd_verify_integration` 3/3 green, including `verify_lia_claim_routes_to_smt_and_mints_witness` (a genuinely non-body-bearing LIA refinement target with a `pre`, no `formals`) which still discharges + mints. Legitimate path intact.

### NIT (not a blocker) — `formals:[]` zero-arg edge slips the marker

`target_is_body_bearing` is `formals.is_some_and(|arr| !arr.is_empty())` (resolve_target.rs:71-74). An EMPTY `formals: []` is still a body-derived op-contract (a zero-arg function), but `!is_empty()` reads it as `false`, so it routes to the vacuous branch.

**Probe C** (target `formals:[]` + non-equation `post:>=(result,0)` + no `pre`; source `=(double(), 6)`):
```
obligationClass:"vacuous", status:"discharged", pass:true, exit 0, ok:true
```

Why this is a NIT and not a blocker: the production writer `bind_function_bridge` always emits an EQUATION post (`post = result == value_expr`), so a real zero-arg function (`fn answer() -> i64 { 42 }` -> `post: result == 42`) is accepted by `lookup` and reduced normally — it never reaches this branch. The shape is reachable only by a non-equation-post writer that does not exist today. The two real-world classes are closed.

**One-line fix when convenient:** change the marker to `formals` PRESENT, not non-empty:
```rust
let target_is_body_bearing = body.get("formals").and_then(|v| v.as_array()).is_some();
```
A `formals` field present (even empty) signals body-derived op-contract; a missing field signals refinement target. That is the complete structural marker, and your own doc comment ("if a future contract shape carries body markers ... it MUST be recognized here") invites exactly this tightening. A regression for the `formals:[]` + non-equation-post case would lock it.

### Side note (completeness, not soundness; track separately)

`runner::work_one` never calls `extract_body_obligation`, so it always REFUSES body-bearing targets (their `ir_formula`/`pre` is always `None`) rather than ever positively discharging them. That is safe (refuse > false-pass), but the runner cannot currently discharge a body claim the way `verify` can. Worth a tracking issue if the runner is meant to mint body-discharge witnesses.

---

**Bottom line:** the keystone honesty boundary is empirically closed across both consumers for every shape a real harvester/writer emits. The remaining `formals:[]` edge is writer-unreachable today and fixed by a one-character marker tightening. **APPROVE-WITH-NITS.** Points 1-5, 7, 8 from the original review stand confirmed-real; not re-litigated.

TSavo:
## Nit fixed: body-bearing marker now gates on `formals` PRESENT (commit 852da2bed)

Closing the one shape the third-pass review flagged as still false-greening (Probe C: `formals: []` + non-equation post + no `pre`). Writer-unreachable today, but closed on principle: the verifier's honesty must hold for **all** inputs, not just what today's bridge writer emits. A zero-parameter function still has a body; its op-contract is body-bearing.

**The change** (`resolve_target.rs`, the single setter): `target_is_body_bearing` now gates on `body.get("formals").as_array().is_some()` (presence) rather than `!arr.is_empty()` (non-empty). A genuinely non-body-bearing target (e.g. a LIA refinement contract) carries no `formals` key at all, so it stays on the legitimate path. Both gated consumers (`verify_one_claim`, `runner::work_one`) inherit the corrected marker through the one setter.

**Regression added:** `verify_zero_arg_body_bearing_non_equation_post_refuses_not_vacuous_pass`.

**Evidence (worktree, z3 on PATH):**
```
FALSEGREEN_STATUS="undecidable"  EXIT=3   (Probe A, <= assertion shape)
FALSEGREEN2_STATUS="undecidable" EXIT=3   (Probe B, non-equation post, formals:["x"])
FALSEGREEN3_STATUS="undecidable" EXIT=3   (Probe C, non-equation post, formals:[])
NEGATIVE_EXIT_CODE=1 STATUS="unsatisfied" (broken body still fails honestly)
POSITIVE_WITNESS_CID=blake3-512:119c5a45...  (positive still discharges + mints)
body-discharge 5/5 · cmd_verify_integration 3/3 (no over-refusal of the LIA path)
```

All three false-green probes now refuse. Cardinal sin closed for every input shape.

**Completeness side-note (not a blocker, tracked separately):** `runner::work_one` never calls `extract_body_obligation`, so it *refuses* body-bearing targets rather than positively discharging them. That is sound (refuse, not false-pass), but it's a completeness gap if the runner is meant to mint body-discharge witnesses. Filing a tracking issue.

TSavo:
## Language-agnostic seam confirmed: production mint+bridge reaches it through the real CLI (PR #1443)

Follow-up to the gap report on this PR. PR #1443 wires the production bridge-writer on top of this branch (rebased onto the latest head incl. your vacuous-pass fixes `3240ab101` / `852da2bed` -- clean, my tests still discharge both ways). It proves the spine you landed here is genuinely language-agnostic: with the bridge written by the TOOL (not hand-built like `cmd_verify_body_discharge.rs:145` does), `verify` discharges through the body unchanged.

`cmd_verify_production_bridge.rs` drives the real `provekit mint` + `provekit verify` binaries (lifter via a genuine newline-delimited JSON-RPC mock matching `LiftPluginKit`):

- mint auto-writes the bridge `double -> <op-contract>` carrying `formals:[x]` + body-derived `post = result == *(x,2)`.
- positive: z3 discharges, signed witness `blake3-512:6719e94fedd2923636d49881bb7f55e9e036a3272a6942de57f0574bf455022a405a2fa5f98cd3b55266c2f2a07871abab534beef2aa3f0dcd909b6aae963561`, exit 0.
- negative (body `x*3`): Unsatisfied, **exit 1**, no witness.

No spine changes -- the seam consumed the tool-minted bundle as-is, which is the language-agnostic claim holding up. Your `852da2bed` ("body-bearing marker gates on formals PRESENT") and this PR's "carry formals through mint" are the two halves of the same contract: mint now writes the formals your marker gates on.

PR #1443 (base `pr-22-verification-spine`).

T Savo

Issue #1444, opened 2026-05-24
Bridge-symbol field-name mismatch: cmd_mint auto-bridge reads name/symbol/fn_name, Go+Java lifters emit fnName


TSavo:
## Context

Surfaced while wiring the Go end-to-end demonstration onto the #1443 production bridge-writer. The auto-bridge that #1443 added to `cmd_mint` derives the bridge `sourceSymbol` from a contract's `name` / `symbol` / `fn_name` field. But the **Go lifter** (`provekit-lift-go`) and the **Java lifter** (`JavaSourceLifter`) emit the function name under `fnName`. So the auto-bridge cannot find the source symbol for Go/Java-lifted function-contracts.

## Current workaround

The Go verify-facing surface binary emits an explicit `bridgeSourceSymbol` field so the bridge resolves. This works but is a per-kit patch over a protocol inconsistency.

## The real fix (pick one, substrate-design call)

1. **Canonicalize the lift-protocol field name** across all lift kits (rust walk, Java, Go, python) so the function symbol is emitted under one agreed key, and `cmd_mint` reads that one key. (Cleanest; one protocol, no per-kit special-casing.)
2. **Or** have `cmd_mint`'s auto-bridge sourceSymbol derivation ALSO read `fnName` (and `bridgeSourceSymbol` if present) as fallbacks. (Smaller, but leaves the field-name divergence in place.)

Option 1 is preferable: the bridge-symbol protocol should be uniform, not casing-divergent per lifter.

## Acceptance

- A Go/Java-lifted function-contract minted via `provekit mint` gets its auto-bridge `sourceSymbol` derived correctly with NO explicit `bridgeSourceSymbol` workaround.
- Rust walk's existing path stays byte-identical (no CID drift from a field rename — verify pinned CIDs).
- A test asserting the auto-bridge resolves for each lifter's emitted contract shape.

Ref: #1443 (production bridge-writer), #1436 (gauntlet). Found during Go onboarding.

TSavo:
Issue bankruptcy 2026-07-01: closing all open issues to start fresh. The current state of record is the design-crimes remediation plan (docs/superpowers/plans/2026-07-01-design-crimes-remediation.md, PR #2979) and the audits in flight. If this issue is still real, it will resurface through an instrument.

PR #1443, opened 2026-05-24
feat(mint): production bridge-writer wires harvest -> verify spine (#1436)

TSavo:
## Summary

PR-23: wires the **production bridge-writer** so a real `provekit mint` → `provekit verify <project>` reaches the verification spine (#1441) with **no hand-bridging**. This is the pipeline my #1436/#1441 gap report identified as missing: `bind_function_bridge` existed but was called only from rust test scaffolding; no production verb wrote the `sourceSymbol → targetContractCid` bridge or carried `formals` through mint. Language-neutral by construction — it unblocks rust-via-CLI and java at once.

### The three deltas (from the gap report, file:line)

1. **Result-var (java).** `JavaSourceLifter` emitted `eq(var("return_value"), ...)`; the seam requires `result` (`body_discharge.rs:54`, enforced `:129`). Renamed both sites (`JavaSourceLifter.java:95`, `:335`) to `result`. Rust's walk already used `result` (`lift.rs:181/314/957`, `compose.rs:389`) — confirmed, no rust change needed.

2. **Carry `formals` through mint.** Added `formals` + `formal_sorts` to `MintContractArgs` (`provekit-claim-envelope/src/lib.rs`), written onto the contract **header** (`memento_body` returns the header for v1.2-layered, so `CatalogResolver` reads `body.formals`/`body.post` directly) and folded into the content CID. Both are **omitted when empty**, so every non-function contract keeps byte-identical CIDs. `cmd_mint.rs` extracts `formals`/`formalSorts` from the `function-contract` decl; `provekit-walk/src/envelope.rs` carries them from the `FunctionContractMemento`.

3. **Wire the bridge-writer into the production mint verb.** `cmd_mint` now, for a body-derived `function-contract` decl (has `post` + `formals`), AUTOMATICALLY mints the bridge via the existing `mint_bridge` (v1.2-layered, signed) with `sourceSymbol = simple_function_symbol(fnName)` and `targetContractCid = m.cid` (the attestation CID = the member key the verifier indexes by). `cmd_bind` is unchanged — it is a substrate algebra pass over ProofIR term JSON and never sees `.proof` bundles; `cmd_mint` is the verb that assembles bundles, so it is the correct home.

### The proof (z3-gated): `cmd_verify_production_bridge.rs`

[...]

PR #1445, opened 2026-05-24
Go onboarding: a Go library declares a boundary and gets a verifiable contract

TSavo:
## Summary

Onboards **Go** as a first-class ProvekIt language, end-to-end and both directions, built on the keystone verification spine (#1441) and the production bridge-writer (#1443). A Go library author declares a boundary/sugar, the library **gets a contract**, and that contract both **verifies** through the spine and **materializes** back into Go via a sugar-supplying shim.

### The authoring surface (`@sugar`/`@boundary` for Go)
- Doc-comment pragma idiom (the Go analog of `#[provekit::sugar]`, in the `//go:generate`/`//go:build` family):
  ```go
  //provekit:sugar(concept="identity", library="builtin", version="1")
  func Id(x int) int { return x }
  ```
- Two-surface `.provekit/config.toml` (`go-bind` / `go-contracts`) mirroring rust's `rust-bind`/`rust-contracts`. Declaration-gated: only annotated funcs get contracts + auto-bridge; unannotated funcs get neither. Malformed directives refuse loudly.

### Verify side (Go library gets a verifiable contract)
- Java-style two-lifter split: `LiftOptions{NormalizeCoreArith}` emits SMT-LIB core ops (`*`, `+`, `<`) on the verify-facing path (resolving the `go:mul` `:`-reserved-char z3 error), while the round-trip dialect (`go:mul`) is preserved byte-identically. The verifier spine was NOT modified — Go was wired INTO it.
- Real Go lifter → `provekit mint` (tool-written auto-bridge, confirmed via `bridges_by_symbol`) → `provekit verify`: positive discharges through the body (z3, signed witness, exit 0); broken body → unsatisfied, exit 1, no witness.

### Realize side (contract materializes via a shim that supplies Go sugar)
- `provekit-realize-go-core`: PEP 1.7.0 plugin; maps a cross-language concept + signature to real Go via a body-template. Unsupported concept/arity refuses loudly (never stubbed).
- The authoring mint auto-projects the realize body-template registry (`menagerie/go-language-signature/specs/body-templates/go-canonical-bodies-*.json`) as a deterministic projection of the signed binding-entries.

### Closed loop
Same `Id` / `identity` concept: DECLARE (binding-entry + bridge) → VERIFY (z3 discharge + signed witness) → MATERIALIZE (realize back to Go). 

## Test plan
- [x] `cmd_authoring_surface_go.rs` — declare-gates-emission + positive discharge + negative (3 tests)
- [x] `cmd_verify_go_production_bridge.rs` — real lifter, tool-written bridge, positive/negative (3 tests)
- [x] `go_realize_materialize.rs` — substrate → Go shim → real sugar, `go build`, refusal (2 tests)
- [x] 6 Go annotation unit tests (parse, discrimination, AnnotatedOnly gating both ways)
- [x] Existing production-bridge tests still green (bare surface unaffected)
- [x] `cargo build --workspace` green; all touched Go modules `go test ./...` green

Deferred (tracked): full `provekit materialize` carrier-rewrite end-to-end (needs in-source carrier annotations); bridge-symbol `fnName` canonicalization (#1444).

## Summary by CodeRabbit

* **New Features**
  * Added Go end-to-end verification examples demonstrating function verification and concept authoring workflows
  * Introduced annotation directives (`//provekit:sugar`, `//provekit:boundary`) for declaring verified Go function boundaries
  * Implemented Go code lifting for verification and concept realization capabilities
  * Added integration tests for Go verification and materialization workflows

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1445?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/go-double/.provekit/config.toml
+++ b/examples/go-double/.provekit/config.toml
@@ -1,0 +1,19 @@
+# ProvekIt project config for the Go end-to-end example.
+#
+# `[authoring] surface = "go"` selects the Go lift surface (the manifest at
+# .provekit/lift/go/manifest.toml). `[solvers]` names z3 for the linear-
+# arithmetic class the body-obligation `(* 3 2) == 6` falls into; the
+# `-smt2 -in` flags mirror the verifier's default z3 invocation.
+[authoring]
+surface = "go"
+
+[solvers]
+default = "z3"
+
+[solvers.dispatch]
+linear_arithmetic = "z3"
+default = "z3"
+
+[solvers.z3]
+binary = "z3"
+flags = ["-smt2", "-in"]
--- a/examples/go-double/.provekit/lift/go/manifest.toml
+++ b/examples/go-double/.provekit/lift/go/manifest.toml
@@ -1,0 +1,15 @@
+# Go lift surface for `provekit mint` / `provekit verify`.
+#
+# `command` is the verify-facing Go lift binary (provekit-lift-go-verify),
+# launched in RPC mode. Build it first from the implementations/go module:
+#
+#     cd implementations/go && go build -o /tmp/provekit-lift-go-verify ./cmd/provekit-lift-go-verify
+#
+# then run the example with the binary on PATH or by editing `command[0]`
+# below to the built path. The integration test
+# (provekit-cli/tests/cmd_verify_go_production_bridge.rs) builds the binary and
+# rewrites this manifest's command to the built path in a tempdir copy, so the
+# in-repo example stays build-location-agnostic.
+name = "go"
+command = ["provekit-lift-go-verify", "--rpc"]
+working_dir = "."
--- a/examples/go-double/.provekit/realize/go/manifest.toml
+++ b/examples/go-double/.provekit/realize/go/manifest.toml
@@ -1,0 +1,15 @@
+# Go realize surface for `provekit materialize`: the SHIM that supplies the
+# native Go sugar for a contract's concept (provekit-realize-go-core).
+#
+# Build it first from the implementations/go/provekit-realize-go-core module:
+#
+#     cd implementations/go/provekit-realize-go-core && \
+#       go build -o /tmp/provekit-realize-go ./cmd/provekit-realize-go
+#
+# then edit `command[0]` to the built path or put it on PATH. The gating test
+# (provekit-cli/tests/go_realize_materialize.rs) builds the binary and installs
+# this manifest into a tempdir, so the in-repo example stays
+# build-location-agnostic.
+name = "go-realize"
+command = ["provekit-realize-go", "--rpc"]
+working_dir = "."
--- a/examples/go-double/README.md
+++ b/examples/go-double/README.md
@@ -1,0 +1,52 @@
+# Go end-to-end example: a Go library gets a contract, and the spine verifies it
+
+This is the Go analog of the Rust production-bridge gauntlet
+(`implementations/rust/provekit-cli/tests/cmd_verify_production_bridge.rs`). It
+demonstrates that ProvekIt's verification spine is LANGUAGE-NEUTRAL: a Go
+function's body-derived contract lifts to ProofIR
+(`post = result == <value-expr>` plus `formals`) and the verifier discharges
+the obligation through the body via weakest-precondition + z3, exactly as it
+does for Rust and Java.
+
+## The library
+
+```go
+func Double(x int) int { return x * 2 }
+```
+
+`double_test.go` carries the harvested assertion `assert.Equal(t, Double(3), 6)`.
+
+## The chain (no hand-bridging, no hand-written contracts)
+
+1. The **real Go lifter** (`provekit-lift-go-verify`, the verify-facing `go`
+   lift surface) lifts:
+   - `double.go` -> `function-contract` with `formals: ["x"]` and
+     `post = result == (* x 2)` (the verify-facing dialect normalizes Go's
+     `go:mul` to the SMT-LIB core symbol `*` so z3 can reduce it),
+   - `double_test.go` -> `contract` with `inv = =(Double(3), 6)` (the Go
+     Layer-0 leaf assertion harvester).
+2. `provekit mint` AUTO-WRITES the bridge `Double -> targetContractCid` for the
+   body-bearing function-contract (#1443). The bridge is TOOL-written, not
+   hand-built.
+3. `provekit verify`:
+   - **positive**: reduces `Double(3) == 6` through the body to
+     `(* 3 2) == 6` -> z3 discharges -> signed witness, exit 0.
+   - **negative** (break the body to `x * 3`): the obligation becomes
+     `(* 3 3) == 6` -> z3 refutes -> Unsatisfied, exit 1, no witness.
+
+## Run it
+
+```sh
+cd implementations/go
+go build -o /tmp/provekit-lift-go-verify ./cmd/provekit-lift-go-verify
+# point this example's .provekit/lift/go/manifest.toml command[0] at the binary
+# (or put it on PATH), then from the repo root:
+provekit mint   --project examples/go-double --out examples/go-double --no-attest
+provekit verify --project examples/go-double --emit-witnesses /tmp/go-witnesses
+```
+
+The gating integration test
+(`implementations/rust/provekit-cli/tests/cmd_verify_go_production_bridge.rs`)
+builds the binary, copies this example into a tempdir, and asserts both the
+positive discharge (signed witness, exit 0) and the honest negative
+(Unsatisfied, exit 1, no witness).
--- a/examples/go-double/double.go
+++ b/examples/go-double/double.go
@@ -1,0 +1,14 @@
+// Package sample is the Go library whose contract ProvekIt extracts and
+// verifies end-to-end. `Double` is a body-bearing function: the Go lifter
+// lifts its body `x * 2` to the verify-facing function-contract
+// `post = result == (* x 2)`. `provekit mint` auto-writes the
+// `Double -> targetContractCid` bridge (#1443); `provekit verify` reduces the
+// harvested `Double(3) == 6` assertion through the body via z3.
+package sample
+
+// Double returns twice its argument. The honest body is `x * 2`; the
+// integration test's negative case mutates it to `x * 3` to prove the spine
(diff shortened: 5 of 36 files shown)


TSavo:
## Independent review (fresh eyes) — Go onboarding

**Verdict: REQUEST-CHANGES.** One blocker: the verify-path op-normalization maps `go:div`/`go:mod` to SMT-LIB `div`/`mod`, whose semantics DIVERGE from Go for negative operands. I produced a **signed witness for a contract that is false in Go**. The spine itself is untouched and correct; the bug is in the new Go op-table. Everything else is clean. Fix is a one-line diff.

### BLOCKER — `go:div`/`go:mod` -> SMT `div`/`mod` is semantically unsound (cardinal sin, with a signed witness)

Go `/` truncates toward zero: `-7 / 2 == -3`. SMT-LIB Int `div` floors toward negative infinity: `(div -7 2) == -4`. They diverge for every negative dividend. `go:mod -> mod` has the identical divergence (Go `%` takes the dividend's sign; SMT `mod` is always non-negative).

`coreArithOp` (`implementations/go/provekit-lift-go/lift.go:62-65`) maps both, and the verify-facing lifter (`provekit-lift-go-verify`, `NormalizeCoreArith: true`) emits them, so SMT `div` reaches z3.

**Empirical proof** -- real `provekit mint` + `provekit verify`, real Go lifter, `func Halve(x int) int { return x / 2 }`:

| harvested assertion | Go truth | SMT truth | verdict | witness |
|---|---|---|---|---|
| `Halve(-7) == -4` | **FALSE** (`-7/2==-3`) | true (`div -7 2 == -4`) | **discharged, pass=true, exit 0** | **`blake3-512:1ef14d4f...` written** |
| `Halve(-7) == -3` | **TRUE** | false | unsatisfied, exit 1, no witness | none |
| `Halve(6) == 3` | true | true | discharged | written |

The first row is the cardinal sin: ProvekIt minted and signed a witness for a contract that does not hold in Go. The second is the mirror image -- a contract Go actually satisfies is REJECTED. This is not a missed proof (Undecidable would be honest); it is an INVERTED proof. The third confirms the path works when the semantics agree, ruling out "z3 never saw `div`."

**The fix is one diff:** remove the `go:div` and `go:mod` cases from `coreArithOp`. They then fall through to namespaced names, which z3 sees as uninterpreted -> Undecidable (the honest verdict: "this kit cannot discharge Go integer division yet"). Faithful modeling would need bit-width/Euclidean-vs-truncation handling (BV theory or guarded `div`), which is a feature, not this PR. Refusing is correct until then.

**Test-coverage gap (separate from the bug):** the PR's own end-to-end tests pass because they only exercise `*` (multiplication), where Go and SMT agree. The unsoundness lives exactly where the test suite has no coverage. Please add a regression embedding the three-run probe above: Go-false MUST be Undecidable (not Discharged), Go-true MUST be Undecidable (not Unsatisfied). I have the probe fixture and will contribute it.

`go:neg` -> unary `-` is the same bounded-int-overflow family as `add/sub/mul` (Go wraps at MinInt; SMT negates unbounded) -- not the divergence class, so not a blocker, but worth a look while you are in the table.

### The other points — clean

- **Spine untouched (confirmed):** merge-base diff (`184976ef6..fa5463c60`) shows ZERO changes to `provekit-verifier`, `libprovekit`, `provekit-claim-envelope`, `cmd_verify`, `cmd_mint`. Go is wired INTO the spine by speaking its op vocabulary. The #1441 cardinal-sin closure survives: `cmd_verify_body_discharge` 5/5 green.
- **(A, round-trip half):** the round-trip dialect stays byte-identical. `normalizeCoreArith` defaults false; `lift_test.go:48` asserts default `LiftSource` still emits `"name":"go:add"` (not `+`). Normalization is verify-path-only. Good.
- **(B) Authoring gating:** `cmd_authoring_surface_go` 3/3 green -- declared boundary lifts, unannotated does NOT (`go_authoring_surface_emits_only_declared_boundary`), broken declared body fails honestly. `go-bind` emits `library-sugar-binding-entry` (a declaration kind mint does not double-mint as a contract). Correct.
- **(C) Realize honesty:** `provekit-realize-go-core` REFUSES, never mis-emits. `MissingTemplateError` -> RPC `-32100` (`rpc.go:62`); discrimination tests exist for both unsupported concept and wrong arity (`realizer_test.go`, `go_realize_shim_refuses_unsupported_concept`). Same bar we just enforced in python; met.
- **(D) Body-template registry:** the projection (`project_body_templates_for_sugar_bindings`, `cmd_mint.rs:1340`) is a deterministic pure function (group binding-entries by `(lang, libtag)`, substitute `${paramN}`, `blake3_512(JCS(content))` for the header CID) -- byte-identical re-run by construction. The committed `go-canonical-bodies-builtin.json` (HEAD: `return ${param0}`) MATCHES `id.go`'s body and the realizer's template; no drift.
- **(E) Tool-written bridge + no spine regression:** `go_mint_auto_writes_body_discharge_bridge` confirms the bridge comes from `pool.bridges_by_symbol` (TOOL-written, not hand-built). `bridgeSourceSymbol` stopgap for #1444 is acceptable as documented. All existing suites green; `go test ./...` green across lift-go / realize-go-core / validator.

### Nit (not a blocker): the body-template projection is not hermetic

`locate_menagerie_root` (`cmd_mint.rs`) walks up from CWD to find `menagerie/` and writes the projected registry THERE, escaping the project's `--out` sandbox. Running `provekit mint` from inside the repo (including this PR's own integration tests) rewrites the committed `menagerie/.../go-canonical-bodies-*.json`. I observed my own probe runs mutate the tracked file (HEAD content is correct; the disk delta was test leakage, which I restored). Consider writing the projection under `--out`, or gating it off for sandboxed runs, so tests do not clobber tracked artifacts.

---

**Bottom line:** the Go onboarding is the right shape -- spine untouched, round-trip byte-identical, authoring declaration-gated, realize refuses-not-misemits, bridge tool-written. But `go:div`/`go:mod` normalization is a soundness hole that mints a signed witness for a Go-false contract. Remove those two cases from `coreArithOp` (let them refuse), add the negative-operand regression, and this is an APPROVE. Worktree left clean (probe removed, registry restored, no push).

TSavo:
## Independent review (fresh eyes) — re-review of the cardinal-sin fix (d713cfaaa)

**Verdict: APPROVE.** The inverted proof is closed, the fix is surgical (no legitimate op over-refused), and the spine closure + authoring/realize/bridge behavior are unchanged. The hermeticity nit is fixed too.

### 1. The inverted proof is CLOSED (re-ran my original probe + mod)

Re-ran the exact probe that produced a signed witness last pass, now on d713cfaaa (real `provekit mint` + `verify`, real Go lifter):

| assertion | Go truth | was (fa5463c60) | now (d713cfaaa) |
|---|---|---|---|
| `Calc(-7) == -4` (x/2) | FALSE | **discharged + signed witness** | **undecidable, exit 3, 0 witnesses** |
| `Calc(-7) == -3` (x/2) | TRUE | unsatisfied | undecidable, exit 3, 0 witnesses |
| `Calc(-7) == 1` (x%2) | FALSE | (mod, same class) | undecidable, exit 3, 0 witnesses |
| `Calc(-7) == -1` (x%2) | TRUE | -- | undecidable, exit 3, 0 witnesses |
| `Calc(7) == 3` (x/2) | TRUE | -- | undecidable, exit 3, 0 witnesses |

No `/`- or `%`-containing contract can produce a signed witness anymore: `go:div`/`go:mod` fall through `coreArithOp` to namespaced symbols, stay uninterpreted, and the obligation is Undecidable in BOTH directions. Note even the agree-case (`7/2==3`) is Undecidable -- correct: the verifier refuses uniformly rather than discharging the cases where Go and SMT happen to coincide (it has no way to tell which is which without faithful modeling). That is the honest posture (Supra omnia, rectum: refuse, never false-discharge). The committed regression `cmd_verify_go_division_unsound` (2 tests) asserts the same and is green.

### 2. No legitimate op over-refused -- the fix is surgical

`coreArithOp` audit after the fix: KEEPS `+ - * neg`, the five comparisons, and `and/or/not`; DROPS only `div`/`mod`. `opForDialect` is the single normalization site (binary + unary), and `coreArithOp` is its single source of truth -- compound `/=`/`%=` produce `go:div`/`go:mod` and therefore also stay namespaced. No other leak (bitwise/shift/ne/deref were already namespaced; I re-verified).

Multiplication still discharges: `go_production_path_double_discharges_and_mints_witness` mints a signed witness for `Double(3)==6` through body `x*2` (`blake3-512:61f1d2af...`), broken `x*3` is unsatisfied/exit 1. The `+ - *` overflow-modeling gap is the shared rust/java unbounded-Int baseline, not introduced here -- correctly out of scope, tracked separately.

### 3. #1441 closure + authoring/realize/bridge unchanged

- body-discharge spine: 5/5 green (all three false-green regressions still refuse; positive discharges; negative unsatisfied).
- authoring gating: 3/3 (declared boundary lifts, unannotated does not, broken declared body fails).
- realize honesty: 2/2 (refuses unsupported concept via -32100, materializes identity).
- tool-written bridge: `go_mint_auto_writes_body_discharge_bridge` green (bridge from `bridges_by_symbol`).

### 4. Hermeticity fix works (tracked file no longer clobbered)

`cmd_authoring_surface_go.rs` now stages its own `menagerie/` in the tempdir and runs mint with `current_dir(project)`, so `locate_menagerie_root` resolves to the staged tree. I ran the authoring-surface test (the one that clobbered the tracked `go-canonical-bodies-builtin.json` during my first pass) and `git status menagerie/` stayed CLEAN. The production walk-up in `cmd_mint.rs` is left untouched -- correct scope (test hygiene, not a behavior change).

---

**Bottom line:** the cardinal sin (signed witness for a Go-false division contract) is empirically gone; div/mod now honestly refuse; legitimate `+ - * neg`/comparisons/booleans still discharge; the spine closure I hardened in #1441 holds; and the projection no longer leaks into tracked files during tests. **APPROVE.**

(Note on the separate rust `Div->"/"` / `Rem->"%"` question you are tracking independently: I did NOT exercise the rust lift path in this re-review, so I make no claim either way here -- it is out of scope for #1445 and the same divergence reasoning would apply to z3's Int `/`/`%`, worth your independent check.)

Worktree left clean (re-probe removed, no commits, no push).

Issue #1446, opened 2026-05-24
Faithful Go integer division/modulo modeling (truncate-toward-zero) so go:div/go:mod can verify, not refuse


TSavo:
## Context
PR #1445 review caught the cardinal sin: `NormalizeCoreArith` mapped `go:div`/`go:mod` to SMT-LIB `div`/`mod`, which floor toward -inf while Go truncates toward zero. `Halve(-7) == -4` (FALSE in Go) discharged with a signed witness. Fixed in d713cfaaa by leaving `go:div`/`go:mod` UNINTERPRETED -> verify returns Undecidable (exit 3), no witness. That is honest but it means Go integer division currently cannot be VERIFIED, only refused.

## Divergence
- Go: `-7 / 2 == -3`, `-7 % 2 == -1` (truncate toward zero; remainder takes dividend's sign)
- SMT-LIB `div`/`mod`: `(div -7 2) == -4`, `(mod -7 2) == 1` (Euclidean / floor)
- z3 `/` `%` on Int: real division / floored remainder -- also wrong.

## Goal
Model Go `/` and `%` faithfully so the verifier can discharge division obligations instead of refusing. Options:
1. **Euclidean correction**: emit `div`/`mod` plus a sign-correction term so the composite equals Go truncation for all signs. (`q = (a - (a rem b)) / b` with rem taking a's sign.)
2. **Bitvector theory**: model `int` as a fixed-width `bv` and use `bvsdiv`/`bvsrem` (truncate toward zero, signed) -- semantically exact for Go's machine ints, and also closes the unbounded-Int overflow-modeling gap. Larger change; coordinate with the rust/java baseline (they emit `/`/`%` op names today).

## Acceptance
- A `Halve(x) = x/2` contract with `Halve(-7) == -3` (Go-true) DISCHARGES with a signed witness; `== -4` (Go-false) is UNSAT/refuted -- the inverse of today's both-Undecidable behavior.
- Re-enable `go:div`/`go:mod` in `coreArithOp` (implementations/go/provekit-lift-go/lift.go) only once the modeling is faithful; update the guard test cmd_verify_go_division_unsound.rs accordingly.
- Decide cross-language: do rust/java also need faithful div, or is Go leading here?

## Related
- Cardinal-sin guard + audited `coreArithOp` table: d713cfaaa
- Regression test: implementations/rust/provekit-cli/tests/cmd_verify_go_division_unsound.rs

## Secondary (smaller, also from #1445 review)
[...]

PR #1448, opened 2026-05-24
Full slow-lane green: python sidecar types + honesty guard, citation reconciliation, D7 visibility

TSavo:
## Summary

Closes the three pre-existing slow-lane failures (red on `main`, non-required, blocking nothing but obscuring CI). Three independent root causes, each fixed at the correct layer. The honesty guard (gate #1) is the most important: it closes a cardinal-sin-class bug where the python realizer silently emitted wrong code.

### Gate #1 — python realize: sidecar type-threading + honesty guard (`6eed384c9`)
- **Bug**: the python lift dropped `:int`/`->int` annotations from the bind-lift-entry, so the realizer got no types; for `factorial` the recursive call's return type was unknown, the enclosing `mul` couldn't match its signature-keyed template, and the conditional refused. **And** with bindings absent the realizer *silently emitted wrong code* (`if (x) < (x)` for `if x < 0`).
- **Fix (CID-invisible)**: types travel through the realize **sidecar** (`realize_param_types`/`realize_return_type`): lift emits → bind forwards + **strips from the canonical lift term** → `merge_realize_sidecar` injects into the realize spec. The federation invariant is preserved — python bind CID stays `a212…` (the naive "put types in the bind-entry" fix moved it to `34b9…` and broke `seam4_federation_rust_vs_python_lift_bind_byte_identity`).
- **Honesty guard**: the realizer now REFUSES (`missing body-template entry`) when a term-shape has unbindable empty leaves and no bindings, instead of mis-emitting. Discrimination test `test_term_shape_without_bindings_refuses_rather_than_misemits`.
- `python_emit_compile_run_conformance`: 2 passed.

### Gate #2 — lower: two-named-term reconciliation (`2d0218a9c`)
- The rust lift emits two named terms sharing a `term_shape_cid` (a sugar-binding carrier + the body); the python lower rejected >1. `reconcile_single_realizable_term` collapses to the realizable body when one is the sugar-binding; genuinely distinct terms still refuse loudly.
- `trinity_citation_comments_exhibit`: 3 passed.

### Gate #3 — realize-rust: round-trip source visibility (`1562efa3e`)
- Root cause was NOT rustfmt: the realizer dropped `pub`, emitting a private `fn`, so the D7 source round-trip diverged. `emit_from_resolved_with_visibility` (panic-safe RAII restore) threads source visibility through. Boy-scout-fixed the same bug in `d7_v4_widen` + `d7_v7_module_sweep`. 8 gold receipts regenerated (verdict stays BYTE_IDENTICAL; only the raw pre-rustfmt source changed: `pub` restored + `Arc < Value >`→`Arc<Value>`).
- `d7_v1_source_round_trip::value_null...`: 1 passed.

## Test plan
- [x] `cargo build --workspace` clean
- [x] gates #1/#2/#3 green (2 / 3 / 1 passed)
- [x] honesty discrimination test green
- [x] #1441 body-discharge / consensus probes intact (`cmd_verify_integration` 3, `witness_consensus_test` 2)
- [x] One workspace failure (`lower_claim_bind_result::...erased_signature_defaults`) confirmed pre-existing-identical on clean base `95b9632fd`; not in the slow lane

Ref: #1436. Note: aligns with #1446 (the realizer honesty + type-threading complements the div/mod faithful-modeling work).

## Release Notes

* **New Features**
  * Added support for preserving source visibility modifiers when regenerating code.
  * Introduced new API for explicit control over visibility during code emission.

* **Bug Fixes**
  * Prevents incorrect code generation when operand bindings are incomplete; now surfaces clear error messages instead.

* **Tests**
  * Added regression test ensuring proper error handling for unbindable operand placeholders.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1448?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TSavo:
## Independent review (fresh eyes) — slow-lane closure (1562efa3e)

**Verdict: APPROVE.** All four CID-sensitive / cardinal-sin-class cruxes check out on direct read + empirical probe. The verifier spine discharge logic is untouched (merge-base diff to `provekit-verifier`/`wp.rs`/`claim-envelope::lib.rs` is empty). Note this branch sits on the pre-#1441 base `95b9632fd`, so the #1441 body-discharge closure isn't on this branch to re-verify -- the spine here is unregressed at its own level.

### (A) seam4 federation byte-identity -- the python bind CID is UNCHANGED at `a212`

Ran `seam4_federation_rust_vs_python_lift_bind_byte_identity` on the PR branch AND on a clean base worktree of `95b9632fd` (with the prerequisite binaries built so both reach the CID assertion). Byte-for-byte identical on both sides:

```
base 95b9632fd:  left(rust)=blake3-512:a51e5f78...  right(python)=blake3-512:a21284436f...
PR   1562efa3e:  left(rust)=blake3-512:a51e5f78...  right(python)=blake3-512:a21284436f...
```

The python bind `.to` CID is `a212...` on both -- the sidecar strip works: `realize_param_types`/`realize_return_type` are added to `BindLiftEntry`, written into the CID-invisible realize sidecar, and removed (both snake_case AND camelCase forms) by `strip_realize_sidecar_from_lift_term` (`bind.rs`) before the CID is taken. They never enter the canonical term, so A9/#1075's type-erasure invariant holds and the naive `34b9` move was avoided.

The `a51e != a212` divergence is a **pre-existing rust-vs-python federation gap** (rust `left` is `a51e` on base too, byte-identical), NOT introduced by this PR and NOT masking a python CID move. The agent's attribution is correct: this PR preserves the invariant it touches and does not regress the gap it does not claim to close. (The seam4 test is red on base and red on PR with identical CIDs both sides.)

### (B) Honesty guard is GENERAL, not fixture-specific (cardinal-sin-class)

`_term_shape_has_unbindable_empty_leaf` checks the STRUCTURAL property "any empty operand-slot leaf (`{}`, no `kind`/`value`)", recursing under cataloged ops -- not a string match against `if x < 0`. I probed the paths the fixture does not cover (real realizer calls):

| input (abs_value shape, unbindable empty leaves) | result |
|---|---|
| `operand_bindings=None` | REFUSED (new guard -> None) |
| `operand_bindings=[]` (empty list, NOT None) | REFUSED via `OperandBindingMisalignmentError` (loud) |
| partial bindings (2 of 4 leaves bound) | REFUSED, cites missing `[[1,0],[2]]` |
| empty leaf nested under a `kind`-bearing non-concept node | guard detects (`True`), emission REFUSED |
| FULL bindings (sanity, no over-refusal) | EMITS `if (x) < (0): return -(x) else: return x` (the `0` literal preserved, NOT the wrong `if (x) < (x)`) |

The `is None` guard plus the pre-existing misalignment check cover all binding-absence shapes. No degenerate input slips through to silent mis-emission; legitimate full-binding emission is not over-refused. python realizer suite: 50/50 green.

### (C) Gold receipts legitimately regenerated, not rubber-stamped

`emit_from_resolved_with_visibility` uses a real `impl Drop` (`VisibilityGuard`) that captures the prior `CURRENT_VISIBILITY` and restores it on drop -- **panic-safe**: an unwind between set and emit still restores via Drop; it restores to the captured previous value (not unconditionally `""`), so reentrant/nested calls are correct. No state leak across emissions.

The d7 source-round-trip suites (`d7_v1`/`d7_v4`/`d7_v7`) **PASS on the PR branch**, including `value_null_source_round_trip_receipt_is_complete` (which failed pre-existing on base in prior reviews). The fix threads source visibility into the realizer (`visibility_of_fn_slice` -> `emit_from_resolved_with_visibility`) so `pub` round-trips. All 8 regenerated receipts changed ONLY `step_3_4_regenerated_source` (`Arc < Value >` -> `Arc<Value>` spacing to match the now-correct realizer output); `step_7_byte_identical_post_rustfmt` stays `true` on all 8 -- verdicts NOT weakened.

### (D) Lower reconciliation does NOT over-collapse

`reconcile_single_realizable_term` collapses to the realizable body ONLY when (1) every term shares the SAME non-empty `term_shape_cid` (strict string equality, `!is_empty()` guards against spurious empty-CID matches) AND (2) exactly one term carries non-empty `fn_name_sugar`. Genuinely-distinct terms (different `term_shape_cid`) skip the collapse and hit the loud `expected exactly one realizable named term` refusal; two real bodies sharing a shape -> `bodies.len()==2` -> also refuses. The multi-function error still fires. The sidecar-type merge is also honesty-correct: `has_real_type` requires a non-empty type, so an all-empty (unannotated) vector does NOT clobber the `int` reconstruction heuristic.

### (E) The slow-lane failures + a fourth pre-existing one

The three the PR claims to close: d7 round-trips (PASS), `python_emit_compile_run_conformance` 2/2 (PASS). Confirmed closed.

Separately: `bind_result_claim_lower_reconstructs_erased_signature_defaults` FAILS on the PR branch -- but it fails IDENTICALLY on clean base `95b9632fd` (same test, same panic line 268, same 5-passed/1-failed). It is a FOURTH pre-existing failure this PR does not claim to fix and does not regress (it shares `lower_plugin.rs` with the (D) change, so I verified it on base specifically). Out of scope, flagging for the team's tracking.

**Could not run** `trinity_citation_comments_exhibit`: it needs the Java realize jar, whose Maven build fails on `javaparser` dependency resolution in my sandbox (no network). The PR touches ZERO Java files, so this is environmental, not a PR concern.

---

**Bottom line:** spine untouched; python bind CID byte-identical at `a212` (sidecar strip correct, seam4 red is a pre-existing federation gap, not a regression); honesty guard refuses-not-misemits across all degenerate inputs; gold receipts legitimately reflect the fixed `pub`-round-trip behavior with verdicts unchanged; reconciliation refuses loudly for genuinely-distinct terms. **APPROVE.** Worktree left clean (probe removed, base worktrees removed, no push).

PR #1449, opened 2026-05-24
Fix erased-signature int-reconstruction (conformance gate, layer 2)

TSavo:
## Summary

Closes `lower_claim_bind_result::bind_result_claim_lower_reconstructs_erased_signature_defaults`, the next failure in the Cross-language conformance gate's `make test-rust` step (surfaced after #1448's d7 fix peeled off the first layer).

**Confirmed pre-existing, not a regression**: the identical failure (`left: []`, `right: ["int"]`) reproduces on pristine `95b9632fd` — it predates this session's work, and #1448's sidecar change is a no-op on this erased path.

**Root cause**: `realize_signature_from_named_term` (lower_plugin.rs) gated the "erased signature" case on `term.params.is_empty()`. But A9 federation byte-identity strips declared *types*, not *params* — a function with real params round-trips as `params=["x"]`, `param_types=[]`, `return_type="()"`. That case was excluded by the `is_empty()` check, so the lower returned the empty `param_types` verbatim (`[]`) instead of reconstructing the `int` default.

**Fix**: the erased signal is now "no declared `param_types` AND return is absent-or-unit (`""`/`"()"`)" — independent of params. When erased, reconstruct one `int` per declared param (`int` return, `()` for unit concepts). Feeds only the realize REQUEST, never the bind CID.

## Test plan
- [x] `lower_claim_bind_result`: 6 passed (erased-defaults reconstructs `["int"]`; fully-typed path still green)
- [x] `python_emit_compile_run_conformance`: 2 passed (no regression)
- [x] seam4 federation byte-identity: python bind CID unchanged at `a212…` (heuristic feeds the realize request only)

## Note
The conformance gate is **not** fully green after this: the next failure underneath is `provekit_lift_processes_walk_crate_to_proof_catalog` (a provekit-lift self-lift name collision on an `expect` contract), which passes on `95b9632fd` and fails on current main — introduced in the `95b9632fd..HEAD` range, under separate investigation.

Ref: #1436.

## Summary by CodeRabbit

* **Refactor**
  * Improved type handling logic for type erasure scenarios in the verification system. The system now more accurately determines when type information is erased and reconstructs appropriate type signatures accordingly, resulting in more robust type processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1449?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/implementations/rust/libprovekit/src/core/lower_plugin.rs
+++ b/implementations/rust/libprovekit/src/core/lower_plugin.rs
@@ -1192,18 +1192,25 @@ fn sidecar_entry_for_term<'a>(sidecar: &'a Value, term: &NamedTerm) -> Option<&'
 }
 
 fn realize_signature_from_named_term(term: &NamedTerm) -> (Vec<String>, String) {
-    // Erasure heuristic: legacy bind payloads sometimes omit type info
-    // entirely. Distinguish "no types declared" from "types declared as
-    // empty/unit". `return_type == ""` is absence; `return_type == "()"`
-    // is explicit unit. Functions with declared parameters or explicit
-    // return type are NOT erased.
-    let erased = term.param_types.is_empty()
-        && term.params.is_empty()
-        && term.return_type.trim().is_empty();
+    // Erasure heuristic: A9 strips declared types from the canonical lift term
+    // for seam-4 federation byte-identity, so a bind payload that round-trips
+    // through that path carries NO param_types and a unit/absent return even
+    // when the source function had real parameters (e.g. `identity(x)` lifts to
+    // params=["x"], param_types=[], return_type="()"). The erased signal is
+    // therefore "no declared param_types AND return is absent or unit" —
+    // INDEPENDENT of whether params exist. (Sidecar-carried real types override
+    // this guess later in merge_realize_sidecar; this is the fallback when the
+    // payload truly has nothing.)
+    let return_trimmed = term.return_type.trim();
+    let erased =
+        term.param_types.is_empty() && (return_trimmed.is_empty() || return_trimmed == "()");
     if !erased {
         return (term.param_types.clone(), term.return_type.clone());
     }
 
+    // Reconstruct one `int` per declared parameter; a zero-param function yields
+    // an empty vec (correct). The return is `()` only when the concept is unit;
+    // any other erased concept reconstructs to `int`.
     let param_types = term
         .params
         .iter()


TSavo:
## Independent review (fresh eyes) — erased-signature int-reconstruction (eae880aa0)

**Verdict: APPROVE.** Minimal, correct, CID-invisible, and does not over-reconstruct. All four checks pass empirically.

### 1. No over-reconstruction (typed path still uses real types)
`realize_signature_from_named_term` returns early with the REAL types when `!param_types.is_empty()` (`if !erased { return (term.param_types.clone(), term.return_type.clone()); }`). The int-default only fires on the truly-erased branch. `bind_result_claim_lower_uses_named_term_realize_request` (the fully-typed path) stays green -- the heuristic does not clobber declared types.

### 2. seam4 byte-identity holds -- the heuristic is CID-invisible
`realize_signature_from_named_term` is called at exactly ONE site (`lower_plugin.rs:1081`, inside `realize_spec_from_named_term` -- the realize REQUEST builder), NOT on any bind-CID path. Confirmed empirically: seam4 federation still shows python `right=blake3-512:a21284436f...` -- byte-identical to base, unchanged by this PR. The reconstruction feeds the realize request only; it does not leak into the CID. (The `a51e != a212` seam4 red is the same pre-existing rust-vs-python federation gap from #1448, not this PR's concern.)

### 3. Test suites green
`lower_claim_bind_result` 6/6 (the previously-failing `bind_result_claim_lower_reconstructs_erased_signature_defaults` now PASSES; the typed-path, fast-path, fallback, refusal, and target-neutral tests all stay green). `python_emit_compile_run_conformance` 2/2.

### 4. The erased-detection condition is sound (probed the unit-return edge)
The new signal `param_types.is_empty() && (return == "" || return == "()")` is correct, and the unit-return edge is NOT a misclassification:
- A9 strips declared *types* (both param_types AND the return) from the canonical lift term. So a legitimately-unit-returning function with real params arrives here as `params=["x"], param_types=[], return="()"` -- the SAME shape as any erased function. Reconstructing one `int` per declared param (and preserving `()` for the unit return) is the correct fallback for that erased state, not an error.
- The only way to reach this function with REAL param types is the non-erased branch (`!param_types.is_empty()`), which short-circuits to the declared types. There is no path where a function with surviving real param_types gets int-reconstructed.
- And `merge_realize_sidecar` overrides the `int` guess with the sidecar's real types when the payload carries them (the #1448 honesty path: only a non-empty type overrides). So the `int` default is a true last-resort fallback, correctly subordinate to any real type evidence.

The old gate (`term.params.is_empty()`) was the bug: it excluded params-present-but-types-erased functions (the A9 round-trip shape), so they returned `[]` instead of reconstructing. The fix correctly makes erasure independent of param presence.

---

**Bottom line:** the fix detects A9-erased signatures by the right signal (no declared types + unit/absent return, independent of params), reconstructs `int`-per-param only as a last-resort fallback subordinate to real types and the sidecar, and never touches the bind CID. seam4 `a212` unchanged, no over-reconstruction, suites green. **APPROVE.** Worktree left clean (no commits, no push).

PR #1453, opened 2026-05-24
[inventory only] --no-fail-fast to enumerate full conformance-gate failure set

TSavo:
Temporary inventory PR: runs the conformance gate's make test-rust with --no-fail-fast so it reports EVERY failing test binary at once (instead of fail-fast masking all but the first). Used to enumerate the complete CI-Linux failure set for the conformance-gate-green effort. NOT for merge.


TSavo:
Inventory served its purpose (enumerated the full conformance-gate failure set via --no-fail-fast). The real fixes landed separately (#1455, #1456, #1457). Closing the throwaway.

Issue #1454, opened 2026-05-24
Conformance gate: env-dependency provisioning for trinity / verify_artifact / verifier tests


TSavo:
## Context

During the conformance-gate-green effort, a `--no-fail-fast` CI run revealed CORE tests that fail purely on missing environment deps (NOT logic bugs, NOT session regressions, NOT the retired capability). These are deferred from the bug-fix PRs and tracked here for deliberate CI/env provisioning.

## Env-dependent failures

1. **`trinity_*_byte_identical` (12 tests)** — these are the dispatcher-contract tests (Rust `realize_for_bind` output vs Java realize plugin output, byte-identical). They panic at `realize_for_bind failed` BEFORE the byte-identity assertion because **no Java realize JAR is present** (no comparand). `Makefile: test-rust: build-java` should provide it; verify the JAR is reliably built AND discovered at the path the tests expect on the CI runner. (If build-java succeeded but the test looks elsewhere, fix the discovery path; the test also self-builds via Maven `ensure_jar_built`, which needs Maven/JDK on the runner.)
2. **`verify_artifact_and_policy_runs_both_rails`, `verify_artifact_rejects_binary_cid_mismatch`** — `verify` CLI emitted non-JSON stdout → parse panic (supply_chain_rails_cli). Likely a fixture/env gap.
3. **`verifier_requires_empirically_witnessed_cross_language_consensus`** — needs the python sqlite realize kits (sqlite3/aiosqlite) pip-installed/available on the runner.

## Why separate from the bug-fix work

These are CORE tests (verifier spine + dispatcher contract), but their failures are environmental, not logic — they'd pass on a fully-provisioned runner. Fixing them is CI/env plumbing (ensure the Java realize JAR, python sqlite kits, and verify fixtures are provisioned in the conformance-gate job), distinct from the pre-existing logic bugs (visibility default + thread-local leak, bind_lift param_types erasure, lower_cli refusal) being fixed directly.

## Related

- Pre-existing logic bugs being fixed in a separate PR (visibility/leak, bind_lift, lower_cli).
- #1450 (dogfood test debug-binary skip) is a sibling "test doesn't actually run in CI" gap.
- Consider making the gate's `make test-rust` run `--no-fail-fast` permanently so the full failure set is always visible (fail-fast masking is what hid this chain across multiple merges).

TSavo:
[...]

PR #1455, opened 2026-05-24
Conformance gate (pre-existing): realize-rust-core visibility leak + stale test fixtures

TSavo:
## Summary

Fixes three PRE-EXISTING conformance-gate failures (all fail identically on base `95b9632fd`; zero session regressions — verified). None are the retired ProofIR-maximalism; all are core. Env-dependent failures (trinity Java JAR, verify fixtures, python sqlite kits) are deferred to #1454. The `bind_lift` federation-vs-boundary-types conflict is a separate design decision (not in this PR).

### Fix 1 — realize-rust-core visibility default + thread-local leak (`d065c7b8d`) — the real one
The realize-rust-core emit cluster (~13 tests: `emit_from_resolved_*`, `emits_value_*`, `renders_*`, `falls_back_*`) was **flaky**: it passed in the full suite only via thread-local *pollution* (an earlier test leaked `CURRENT_VISIBILITY="pub"` through an unguarded dispatch-path write) and failed in isolation / perturbed order.
- `function_source` now defaults unset `CURRENT_VISIBILITY` to `pub` (public-API intent); an explicit `"private"` sentinel still overrides to a bare `fn` so a private source slice round-trips private.
- Promoted `VisibilityGuard` to a module-level RAII helper and wrapped the leaking dispatch-path write — the thread-local can no longer leak across realizations/tests.
- Result: **order-independent** (realize-rust-core lib 36/36, identical single-threaded). #1448's d7 private round-trip stays BYTE_IDENTICAL (explicit private overrides the default).

### Fix 2 — stale `kind` discriminator in lower refusal test (`3900adcc2`) — test-only
`lower_cli_missing_body_template_refuses_without_partial_output` wrote `"kind": "NamedTermDocument"`; the parser gates on canonical `"named-term-document"` (what the CLI emits), so input-parse died before the realizer and the refusal receipt never surfaced. One-token fix; refusal behavior was always correct (verified on base). No production code.

### Fix 3 — 3 stale realize-rust-core tests vs universal @sugar (`1495f0479`) — test-only
`@sugar` attribute emission is universal (the production bridge + Go authoring surface depend on the concept binding). The 3 tests predate it. Updated to expect the attr (one also had stale type/paren/return assertions, corrected to the real emission). Production emission untouched.

## Test plan
- [x] realize-rust-core lib 36/36, order-independent (leak killed)
- [x] #1448 d7 round-trip 3/3 BYTE_IDENTICAL (private→private preserved)
- [x] #1441 body-discharge 5/5, #1443 production bridge 3/3, #1449 erased-signature 6/6

Ref: #1436. Deferred: #1454 (env provisioning). Separate design call: `bind_lift` federation type-erasure vs Java boundary emit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of private function visibility during code regeneration to ensure private functions are preserved byte-identically rather than defaulting to public.
  * Enhanced thread-local visibility state management to prevent contamination across concurrent operations.
  * Corrected generic type whitespace formatting in generated code.

* **Tests**
  * Updated test cases to reflect proper visibility handling and formatting changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1455?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/implementations/rust/libprovekit/tests/d7_v1_source_round_trip.rs
+++ b/implementations/rust/libprovekit/tests/d7_v1_source_round_trip.rs
@@ -159,7 +159,7 @@ fn term_summary(fixture: &JsonValue, resolved: &ResolvedTerm) -> String {
 }
 
 /// Extract the source-language visibility (`pub`, `pub(crate)`, `pub(super)`,
-/// or `"private"` for private) that precedes the `fn` keyword in a function source
+/// or `""` for private) that precedes the `fn` keyword in a function source
 /// slice. The D7 round-trip must thread this into the realizer so the
 /// regenerated signature reproduces the original visibility byte-for-byte
 /// instead of defaulting to a private `fn`.
@@ -177,10 +177,11 @@ fn visibility_of_fn_slice(slice: &str) -> String {
             "pub".to_string()
         }
     } else {
-        // Private/inherited source. The realizer defaults an unset visibility to
-        // `pub`, so a private source must thread the explicit "private" sentinel
-        // to override that default and regenerate a bare `fn` byte-identically.
-        "private".to_string()
+        // Private/inherited source. Threaded as the explicit empty string, which
+        // the realizer treats as PRESENT-but-private (`Some("")` => bare `fn`),
+        // distinct from an absent visibility (which defaults to `pub`). This is
+        // the same encoding `bind` uses for a private source fn.
+        String::new()
     }
 }
 
--- a/implementations/rust/libprovekit/tests/d7_v1_source_round_trip.rs
+++ b/implementations/rust/libprovekit/tests/d7_v1_source_round_trip.rs
@@ -159,7 +159,7 @@ fn term_summary(fixture: &JsonValue, resolved: &ResolvedTerm) -> String {
 }
 
 /// Extract the source-language visibility (`pub`, `pub(crate)`, `pub(super)`,
-/// or `""` for private) that precedes the `fn` keyword in a function source
+/// or `"private"` for private) that precedes the `fn` keyword in a function source
 /// slice. The D7 round-trip must thread this into the realizer so the
 /// regenerated signature reproduces the original visibility byte-for-byte
 /// instead of defaulting to a private `fn`.
@@ -177,7 +177,10 @@ fn visibility_of_fn_slice(slice: &str) -> String {
             "pub".to_string()
         }
     } else {
-        String::new()
+        // Private/inherited source. The realizer defaults an unset visibility to
+        // `pub`, so a private source must thread the explicit "private" sentinel
+        // to override that default and regenerate a bare `fn` byte-identically.
+        "private".to_string()
     }
 }
 
--- a/implementations/rust/libprovekit/tests/d7_v4_widen.rs
+++ b/implementations/rust/libprovekit/tests/d7_v4_widen.rs
@@ -195,10 +195,11 @@ fn visibility_of_fn_slice(slice: &str) -> String {
             "pub".to_string()
         }
     } else {
-        // Private/inherited source. The realizer defaults an unset visibility to
-        // `pub`, so a private source must thread the explicit "private" sentinel
-        // to override that default and regenerate a bare `fn` byte-identically.
-        "private".to_string()
+        // Private/inherited source. Threaded as the explicit empty string, which
+        // the realizer treats as PRESENT-but-private (`Some("")` => bare `fn`),
+        // distinct from an absent visibility (which defaults to `pub`). This is
+        // the same encoding `bind` uses for a private source fn.
+        String::new()
     }
 }
 
--- a/implementations/rust/libprovekit/tests/d7_v4_widen.rs
+++ b/implementations/rust/libprovekit/tests/d7_v4_widen.rs
@@ -195,7 +195,10 @@ fn visibility_of_fn_slice(slice: &str) -> String {
             "pub".to_string()
         }
     } else {
-        String::new()
+        // Private/inherited source. The realizer defaults an unset visibility to
+        // `pub`, so a private source must thread the explicit "private" sentinel
+        // to override that default and regenerate a bare `fn` byte-identically.
+        "private".to_string()
     }
 }
 
--- a/implementations/rust/libprovekit/tests/d7_v7_module_sweep.rs
+++ b/implementations/rust/libprovekit/tests/d7_v7_module_sweep.rs
@@ -204,10 +204,11 @@ fn visibility_of_fn_slice(slice: &str) -> String {
             "pub".to_string()
         }
     } else {
-        // Private/inherited source. The realizer defaults an unset visibility to
-        // `pub`, so a private source must thread the explicit "private" sentinel
-        // to override that default and regenerate a bare `fn` byte-identically.
-        "private".to_string()
+        // Private/inherited source. Threaded as the explicit empty string, which
+        // the realizer treats as PRESENT-but-private (`Some("")` => bare `fn`),
+        // distinct from an absent visibility (which defaults to `pub`). This is
+        // the same encoding `bind` uses for a private source fn.
+        String::new()
     }
 }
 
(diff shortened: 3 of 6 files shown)


TSavo:
## Independent review (fresh eyes) — visibility default + thread-local leak (1495f0479)

**Verdict: REQUEST-CHANGES** (posted as a comment; gh blocks formal request-changes on a self-authored PR). The thread-local leak fix is genuinely correct and the test-only fixes are honest. But the new `pub` default introduces **silent over-promotion of private functions to `pub` in the production lift -> bind -> lower -> realize pipeline**, because bind.rs encodes private as `visibility: ""` and this PR makes `""` mean `pub` -- without updating bind. One focused blocker; the rest is clean.

### BLOCKER -- private functions over-promote to `pub` in production

This PR flips the unset-visibility default from `""` -> bare `fn` to `""` -> `pub`, with `"private"` as the new explicit-private sentinel. The d7 TESTS were patched to thread `"private"`, but the **production path was not**, and bind.rs's contract says `""` IS private.

**Empirical probe** (direct `dispatch(provekit.plugin.invoke)`, this commit):

```
visibility: ""        (bind.rs's PRIVATE encoding)  ->  `pub fn secret(...)`   <- OVER-PROMOTED
visibility: "private" (the new sentinel)            ->  `fn secret(...)`       <- bare, correct
visibility: "pub"                                   ->  `pub fn secret(...)`   <- correct
```

**The contract conflict:** `bind.rs:158-159` (and `:343-344`) document the field as `"pub"`, `"pub(crate)"`, **or empty for private**; bind constructs `NamedTerm` with `visibility: String::new()` (= private) at `bind.rs:2287`, `:2331`, and propagates `entry.visibility` at `:490`. **Data flow:** `realize_spec_from_named_term` (`lower_plugin.rs:1088`) passes `term.visibility` straight into the realize spec's `"visibility"` field; the dispatch path reads it via `unwrap_or("")` and feeds `function_source`, which now maps `"" => "pub "`. So a private function lifted from real source -> bind emits `""` -> realize emits `pub fn`. The regenerated/materialized source exposes what the source hid.

This is fidelity-loss / API-surface inversion in materialized output (not the inverted-proof cardinal sin -- the over-public code is plainly visible, not a false witness), but it breaks the materialize contract that regenerated source reproduces source-language visibility byte-for-byte. For a substrate whose product surface IS the regenerated code, that's a real defect. Root cause: the `pub` default over-fits the realizer's own crate fixtures (every `emit_stub`/value/trinity fixture legitimately wants `pub`), but is wrong in the federation context where bind's contract is `""`=private.

**Minimal fix (recommend b -- no ripple into bind's contract):**
- (a) Update bind.rs to emit `"private"` (not `""`) for private functions at `:2287`/`:2331`/`:490` + the `:158-159` doc-comment.
- (b) In the dispatch path, thread `Option<&str>` instead of `unwrap_or("")` so `function_source` distinguishes **field absent** (default `pub`, the behavior the PR wants for direct callers) from **field present-but-empty** (bind's private signal -> bare `fn`).

**Missing regression:** a test that drives the lift -> bind -> realize PIPELINE for a private function and asserts the regenerated source has bare `fn`. The current dispatch tests only assert `pub fn` for absent/`""` input (validating the new default works for the realizer's own fixtures, not that bind-emitted-`""` doesn't over-promote). Without it the hole reappears.

### The rest -- verified, clean

- **Thread-local leak IS killed (order-independent).** All `CURRENT_VISIBILITY` writes now go through the module-level `VisibilityGuard` RAII (set at 328 / restore-on-drop at 336); the leaking dispatch-path write is wrapped (`VisibilityGuard::set(&visibility)`). realize-rust-core lib: 36/36 green **both** default-parallel AND `--test-threads=1`. The single-threaded pass is the decisive proof the suite no longer depends on cross-test pollution.
- **d7 private round-trip GREEN** (v1/v4/v7) -- the test side threads `"private"`, so private sources still regenerate bare `fn` byte-identically. #1448's invariant holds *for the test path* (this is exactly why the production path needs the same fix).
- **Test-only `3900adcc2` honest:** the canonical kind the codebase emits is `"named-term-document"` (`bind.rs:511`, `cmd_lower.rs:329/453`); the old `"NamedTermDocument"` was stale and would never parse, so correcting it makes the input actually reach the realizer and exercise the real refusal.
- **Test-only `1495f0479` honest:** the updated assertions (`Arc<Value>` post-rustfmt spacing, universal `#[provekit::sugar]` attr) match ACTUAL emission -- proven by the 36/36 pass (a weakened-to-pass assertion would assert a wrong result; equality with real output is the check).
- **No regression:** #1441 body-discharge 5/5, #1449 erased-signature 6/6.

---

**Bottom line:** the leak fix and test corrections are solid. The blocker is narrow and the fix is small: the `pub` default is correct for the realizer's direct callers but over-promotes private functions in the federation path, because bind's `""`=private contract was left unreconciled. Close the absent-vs-empty distinction (or update bind to emit `"private"`), add a pipeline-level private-round-trip test, and this is an APPROVE. Worktree left clean (probe removed, no push).

TSavo:
## Independent review (fresh eyes) — re-confirm of the over-promotion blocker (ffefc13d4)

**Verdict: APPROVE.** The blocker I raised is closed. The `Option<String>` distinction (absent vs present-empty) is exactly the option-b fix and it works empirically. CID-safe, no over-correction, and the new regression catches the exact hole.

### 1. Over-promotion is CLOSED (re-ran my exact probe)

Same `dispatch(provekit.plugin.invoke)` probe, now on ffefc13d4:

```
visibility: ""        (bind's PRIVATE encoding, PRESENT)  ->  `fn secret(...)`        <- bare, FIXED (was `pub fn`)
absent field          (direct public-API surface)          ->  `pub fn secret(...)`    <- preserved
visibility: "pub"                                           ->  `pub fn secret(...)`    <- verbatim
visibility: "pub(crate)"                                    ->  `pub(crate) fn secret`  <- verbatim
```

`CURRENT_VISIBILITY: Option<String>` with `function_source` matching `None => "pub "`, `Some("") => ""` (bare fn), `Some(other) => "{other} "`. The dispatch path threads `params.get("visibility").and_then(as_str)` (Option, not `unwrap_or("")`) via `set_optional`, so a spec that OMITS visibility defaults to `pub` while a spec carrying `""` (how bind encodes private) emits bare `fn`. The root-cause divergence is gone.

The new regression `dispatch_with_present_empty_visibility_emits_private_fn_not_pub` drives the genuine production RPC shape (`"visibility": ""` PRESENT, full invoke request) and asserts bare `\nfn secret` AND `!pub fn secret` -- it catches the exact over-promotion, not a test-only path.

### 2. CID-safe
The commit touches ONLY `realize-rust-core/src/lib.rs` (internal Option threading) + the three d7 test files + `lower_claim_bind_result.rs` (test). It does NOT touch `bind.rs`, any spec, or any CID surface. bind's emitted private encoding is still `""` (unchanged) -- so seam4 stays `a212` and pinned fixtures are untouched. The fix is confined to realize-rust-core's internal threading exactly as claimed; the test side and the production bind side now AGREE on `Some("")` = private (eliminating the divergence that was the bug).

### 3. No over-correction
- realize-rust-core lib **38/38** green (the 36 prior + 2 new regression tests), and passes both default-parallel AND `--test-threads=1` -- the absent->`pub` stub/value/trinity tests still hold (public-API surface preserved), and the leak fix's order-independence is intact.
- d7 private round-trip **BYTE_IDENTICAL** v1/v4/v7 -- private sources still regenerate bare `fn` (now threaded as `String::new()` / `Some("")`, matching bind's encoding).
- `Some("pub")` -> `pub fn`, `Some("pub(crate)")` -> `pub(crate) fn` verbatim.

### 4. No regression
#1441 body-discharge 5/5, #1449 erased-signature 6/6.

(The leak fix and the two test-only fixes -- `3900adcc2` discriminator, `1495f0479` assertions -- were verified good in my prior pass and are unchanged here.)

---

**Bottom line:** the production lift -> bind -> realize pipeline no longer over-promotes a private fn to `pub` (bind's `Some("")` -> bare `fn`), direct callers still default to `pub`, d7 byte-identity holds, and there's a pipeline-shape regression test guarding it. The fix is confined to realize-rust-core internal threading -- CID-safe. **APPROVE.** Worktree left clean (probe removed, no push).

PR #1456, opened 2026-05-24
Python onboarding to Go-parity: verify-discharge through the spine + @boundary authoring surface

TSavo:
## Summary

Brings **Python** to parity with the Go onboarding (#1445): a python library declares a boundary, GETS a verifiable contract (discharged through the spine via z3, with a signed witness), and materializes back into python. Built on the keystone spine (#1441), production bridge (#1443), and the python realize fixes from this session (#1448/#1451). Mirrors the proven Go procedure.

### STEP 1 — the python lift→verify gap (empirical)
The real python lifter emits a body-derived `function-contract`, but its shape diverged from the verifier's dischargeable expectation on **four** axes (Go diverged on one): result var `return_value` vs `result`; a `python:return` wrapper; **namespaced ops** (`python:mul`, like Go's `go:mul`); and `Value` sorts (the lifter erases `: int`) vs `Int`. All handled by a verify-facing dialect transform (`verify_dialect.py`), the python analog of Go's `NormalizeCoreArith`. The round-trip dialect is preserved.

### Div soundness (the cardinal axis — Go lesson applied preemptively)
`python:div`/`python:floordiv`/`python:mod` are left **namespaced/uninterpreted** — division goes Undecidable via the wp-refusal path, NEVER false-discharges. Empirically (real CLI):
| case | exit | status | witness |
|---|---|---|---|
| `double` x*2, `double(3)==6` | 0 | discharged (z3) | minted |
| `double` x*3 (broken) | 1 | unsatisfied | none |
| `halve` x//2, `halve(-7)==-4` (py-false) | 3 | undecidable | none |
| `halve` x//2, `halve(-7)==-3` (py-true) | 3 | undecidable | none |
(One mid-build correction: an early dialect refused the whole contract on a division op, dropping the bridge into an unproven path; fixed to keep the op namespaced so the bridge is written and wp refuses — identical to Go's stance.)

### STEP 2 — production-bridge verify-discharge
`cmd_verify_python_production_bridge.rs` (3) + `cmd_verify_python_division_unsound.rs` (2). Real lifter → `provekit mint` (tool-written bridge, asserted via `pool.bridges_by_symbol["double"]`) → `provekit verify`. All 5 green. Positive witness `blake3-512:5b9e0369…`.

### STEP 3 — `@provekit.boundary` authoring surface + closed loop
[...]

PR #1457, opened 2026-05-24
Restore seam4 federation byte-identity (type-leak + term_shape alignment)

TSavo:
## Summary

Restores the **federation byte-identity invariant** (`seam4_federation_rust_vs_python_lift_bind_byte_identity`), which has been RED on main since `48b343a43` (2026-05-22) — rust and python lifts of the same construct produced *different* bind CIDs, so contracts didn't actually federate across languages. This closes both layers of the divergence. seam4 is now GREEN: rust bind CID == python bind CID == `a212…`, byte-identical.

This is the "social" half of the substrate: code from different languages must arrive at the *same* CID for the same contract, or they don't recognize each other at the hub. That recognition was broken; this repairs it.

### Layer 1 — type-leak (`74f80c421`)
`48b343a43` put boundary signature types (`param_types`/`return_type`/`original_param_types`) as **bare keys on the CID-bearing `bind-lift-entry`** (to feed the Java boundary emitter), leaking them into the federated bind CID. Rust emits `["i64","i64"]`, untyped python emits none → divergent CIDs. Fix: migrate them to the CID-invisible `realize_*` sidecar (the channel `strip_realize_sidecar_from_lift_term` already strips before hashing — the same pattern #1448 uses). Java boundary consumer (`cmd_lower.rs` `emit_java_wrapper`) updated to read the sidecar. Java keeps its types; they just no longer touch federation.

### Layer 2 — term_shape alignment (`860472991`, `d49b18913`)
A second, deeper divergence: the rust lifter embedded operand names in leaves (`{"kind":"symbol","text":"x"}`) while python emitted empty `{}` leaves + an `operand_bindings` sidecar. **Operand names are sugar** — the term_shape is structural identity, so `f(x)=x*2` and `f(y)=y*2` must share a shape. Verified (not assumed) that python's empty-leaf form is canonical. Fix: a *scoped* param/local reference (`Expr::Path` in `shape_of_expr`) now emits an empty leaf (name travels on `operand_bindings`); *free* identifiers (None/Some/enum paths) keep their symbol leaf. Discharge is unaffected because the realize path already reads `operand_bindings` before `kind=symbol`.

### Discharge intact (the delicate part)
#1441 body-discharge **5/5** green throughout — `double(3)==6` still discharges + mints its witness after the term_shape change; operand info is read from the CID-invisible sidecar.

### Corrected mid-flight (documented honesty)
An initial wire-clear was over-broad and cleared signature *types* from the legacy bind-result lower path, regressing `i64`→`int` inference (`bind_result_claim_lower_uses_named_term_realize_request`). Narrowed to display-only metadata (visibility/generic_params/doc_lines) — seam4 green AND legacy signature fidelity preserved.

### Benign documented side-effect
`{ let q = a+b; q }` now lifts as `concept:assign(q, add)` rather than `seq([assign, q-leaf])` (the trailing bare scoped-var return is an empty leaf, dropped by the pre-existing `collapse_operation_shapes`). A direct consequence of names-are-sugar (`{let q=…}` and `{let z=…}` now share a term_shape). Discharge 5/5 and both production bridges round-trip unaffected; the two term_shape tests that pinned the old shape were updated with names-are-sugar comments + a positive assertion that operand names live on the sidecar.

## Test plan
- [x] **trinity census slow-lane 9/9 — seam4 federation GREEN (rust CID == python CID == `a212`)**
- [x] #1441 body-discharge 5/5; #1443 rust + #1445 Go production bridges round-trip
- [x] `bind_lift_erases_signature_types` passes; legacy bind-result lower path fidelity preserved
- [x] libprovekit (23 suites), walk_rpc (49), provekit-cli lib (36); `cargo build --workspace` green
- [x] No pinned CID literals on the broken rust forms; shape-pinned tests updated to canonical

Ref: #1436. Root regression: `48b343a43`.

### DIFF
--- a/implementations/rust/libprovekit/src/core/bind.rs
+++ b/implementations/rust/libprovekit/src/core/bind.rs
@@ -635,6 +635,25 @@ pub fn strip_realize_sidecar_from_lift_term(term: Term) -> Term {
                 object.remove("realizeParamTypes");
                 object.remove("realize_return_type");
                 object.remove("realizeReturnType");
+                object.remove("realize_original_param_types");
+                object.remove("realizeOriginalParamTypes");
+                // #1075/A9 federation: the bind-lift-entry is the cross-language
+                // boundary surface and must hash to the SAME bytes whether
+                // lifted from typed Rust or untyped Python. The Python lifter
+                // emits only {kind, param_names, term_shape, term_shape_cid,
+                // operand_bindings, realize_*, source_function_name, witnesses};
+                // Rust additionally carries visibility/generic_params/doc_lines
+                // for the Java boundary realize path. Those are realize-only
+                // metadata (read off the UN-stripped lift IR by cmd_lower, never
+                // off this hashed term) so they ride CID-invisible here too,
+                // scoped to bind-lift-entry to leave sugar-entry CIDs untouched.
+                if object.get("kind").and_then(Json::as_str) == Some("bind-lift-entry") {
+                    object.remove("visibility");
+                    object.remove("generic_params");
+                    object.remove("genericParams");
+                    object.remove("doc_lines");
+                    object.remove("docLines");
+                }
             }
         }
     }
--- a/implementations/rust/libprovekit/src/core/bind.rs
+++ b/implementations/rust/libprovekit/src/core/bind.rs
@@ -543,6 +543,24 @@ fn bind_payload_wire_named_term_document(named: &NamedTermDocument) -> NamedTerm
         term.function.clear();
         // fn_name_sugar is preserved: it carries the source fn name as a
         // non-CID-affecting annotation on the citation (Option C sugar layer)
+        //
+        // #1075 federation: the wire op-tree is arg[1] of the federated
+        // concept:bind-result payload (the cross-language CID). Source-language
+        // realize-only metadata — visibility, generic_params, doc_lines, and the
+        // signature types — must NOT ride it, or typed-Rust (`pub fn add(x: i64,
+        // y: i64) -> i64`) and untyped-Python (`def add(x, y)`) bind to different
+        // CIDs. These fields are NOT lost: the full NamedTermDocument (with them
+        // intact) is addressed separately as the bind claim's `artifacts[0]`
+        // (named_cid) and is the canonical realize-input channel; cmd_lower's
+        // production path reconstructs from the ir-document, never from this wire
+        // op-tree. Parallel to the bind-lift-entry strip in
+        // strip_realize_sidecar_from_lift_term.
+        term.visibility.clear();
+        term.generic_params.clear();
+        term.doc_lines.clear();
+        term.param_types.clear();
+        term.return_type.clear();
+        term.original_param_types.clear();
     }
     wire
 }
--- a/implementations/rust/libprovekit/src/core/bind.rs
+++ b/implementations/rust/libprovekit/src/core/bind.rs
@@ -546,21 +546,29 @@ fn bind_payload_wire_named_term_document(named: &NamedTermDocument) -> NamedTerm
         //
         // #1075 federation: the wire op-tree is arg[1] of the federated
         // concept:bind-result payload (the cross-language CID). Source-language
-        // realize-only metadata — visibility, generic_params, doc_lines, and the
-        // signature types — must NOT ride it, or typed-Rust (`pub fn add(x: i64,
-        // y: i64) -> i64`) and untyped-Python (`def add(x, y)`) bind to different
-        // CIDs. These fields are NOT lost: the full NamedTermDocument (with them
-        // intact) is addressed separately as the bind claim's `artifacts[0]`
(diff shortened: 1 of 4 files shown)


TSavo:
## Independent review (fresh eyes) — restore seam4 federation byte-identity (d49b18913)

**Verdict: APPROVE.** The federation invariant is genuinely restored: rust and python now produce a byte-identical bind CID. Discharge survives, the scoped/free split is sound, legacy fidelity holds, the benign side-effect is a correct names-are-sugar consequence, and nothing on this branch regressed. The verifier discharge logic is untouched (merge-base diff to `body_discharge.rs`/`wp.rs`/`resolve_target.rs` is empty) -- this is a lift/bind representation change, exactly as intended.

### 1. seam4 GENUINELY green -- byte-identical (THE acceptance criterion)

Ran `seam4_federation_rust_vs_python_lift_bind_byte_identity` (--features slow-tests) and captured both CIDs:

```
SEAM4_RUST_CID   = blake3-512:a21284436fbb527924093effd0355ad00855acfcdd5b17333e3680153d53a125ddd56e135accb221af2547155ea28648da0042bd1df866dc40bd78f9cba7be19
SEAM4_PYTHON_CID = blake3-512:a21284436fbb527924093effd0355ad00855acfcdd5b17333e3680153d53a125ddd56e135accb221af2547155ea28648da0042bd1df866dc40bd78f9cba7be19
```

Equal, byte-for-byte. In my #1448 review this was RED (`rust a51e != python a212`); the rust CID has now MOVED from `a51e` to `a212`, converging on python's canonical empty-leaf form -- so the fix corrected the rust side toward the canonical shape rather than papering over the divergence. Federation is restored. The full trinity census (all four federation seams 1/3/4/6) is 9/9 green.

### 2. Discharge intact -- the delicate part
#1441 body-discharge 5/5: `double(3)==6` STILL discharges + mints its witness after operand names moved out of the leaves into `operand_bindings`. This survives because the realize binary's `term_shape_leaf_expression` checks `operand_bindings.get(position)` BEFORE the `kind=symbol` leaf, so it never read the scoped-name leaves -- emptying them is invisible to discharge. Confirmed empirically, not just from the code comment.

### 3. Scoped-vs-free split is correct
Both the operand-binding emission (`bindings_of_expr`, walk_rpc.rs:2588) and the leaf-shape emission (walk_rpc.rs:3359) gate on the SAME `ctx.scoped_names.contains(&...)` set. So a scoped param/local gets BOTH an empty leaf AND an operand_bindings entry (in sync), while a FREE identifier (None/Some/enum paths) gets neither -- it keeps its `{kind:symbol,text}` leaf so deeper consumers can lower it without position-threading. No scoped name is emptied without its binding; no free identifier is wrongly emptied. The distinction is driven by one set, so the two halves cannot drift.

### 4. Legacy bind-result fidelity preserved
`bind_result_claim_lower_uses_named_term_realize_request` green (typed signatures stay typed -- the narrowed wire-clear no longer regresses `i64`->`int`). `lower_claim_bind_result` 6/6. The L1 erasure is verified by `bind_lift_erases_signature_types_from_bind_ir_entries` (bare `param_types`/`return_type` absent from the CID-bearing entry; `realize_param_types`/`realize_return_type` carry them on the CID-invisible sidecar) -- and seam4 stays green WITH the narrowed clear (both held simultaneously).

### 5. The benign side-effect is actually benign
`{ let q = a+b; q }` lifting as `concept:assign` (trailing bare scoped-var return -> empty leaf -> dropped by `collapse_operation_shapes`) is covered + correct: `term_shape_let_binding_preserves_assignment_boundary` and `term_shape_top_level_operator_differs_from_let_assignment_boundary` are green (walk_rpc 49/49). It's a correct names-are-sugar consequence (`{let q=…}` == `{let z=…}` shape), and discharge 5/5 + both available production bridges round-trip unaffected -- not a silent loss of a real operation.

### 6. No regression + CID-migration
- #1443 rust production bridge 3/3; #1445 Go production bridge 3/3.
- walk_rpc unit suite 49/49 -- the shape-pinned tests were UPDATED to the canonical empty-leaf form (e.g. `term_shape_simple_add_is_canonical_gamma_literal`, the `{}` leaf assertions), NOT deleted, and no stale pins on the broken rust CIDs remain.
- bind_kit 8/8.

Scope note: the python production bridge (#1456) is NOT on this branch -- it branched from `c43aa0e70`, before #1456 merged, so `cmd_verify_python_production_bridge` does not exist here (it's not a test target). That regression check is N/A on this base; the two compose after both reach main. I verified what exists on this branch (rust + Go bridges, discharge, the full federation census).

---

**Bottom line:** rust bind CID == python bind CID, byte-identical at `a212` (was `a51e != a212`); discharge survives the operand-name move (realize reads bindings before leaves); the scoped/free split is driven by one consistent `scoped_names` set; legacy typed signatures stay typed under the narrowed clear; the assign-vs-seq shift is a correct, tested names-are-sugar consequence. The core federation property is restored. **APPROVE.** Worktree left clean (temporary CID-print reverted, no push).

PR #1458, opened 2026-05-24
Materialize via .proof + RPC: kill canonical-bodies JSON (jackson/gson template) + kit-owned assembly

TSavo:
## Summary

Establishes the architecture where **the `.proof` is the only substrate-written disk artifact and every language-specific act is the kit speaking RPC** — and migrates the first two shims (Jackson, Gson) off their canonical-bodies JSON to prove it. This is the template the remaining ~21 library-shim kits follow.

Before: `provekit materialize` realized by having `SugarRealizer` read `java-canonical-bodies-<tag>.json` **from disk**, and the same-language `--out-dir` path assembled the Java compilation unit **in the rust CLI** (emitting `__fragment_imports__` as a comment, no helper hoist, no classpath → javac failed). Both violate the rule: the substrate did language-specific work and persisted non-`.proof` JSON.

### PIECE 1 — `.proof`-load via RPC (`6f73906c1`)
- New `RealizeRequest.body_templates` (`bodyTemplates`) RPC param. `cmd_materialize` reads the shim's `.proof` (manifest `sugar_proof` pointer → glob the shim dir's `*.proof`), extracts `library-sugar-binding-entry` records for the target tag, and attaches them to the realize spec; child realize calls inherit them.
- Factored `binding_entry_to_template_entry` (incl. the `${paramN}` placeholder rewrite) — shared by the disk-write projector (`cmd_mint`) and materialize.
- `SugarRealizer` prefers RPC-provided templates (`currentBodyTemplates` ThreadLocal, never statically cached), falls back to disk only when absent — so other kits are untouched.
- **`java-canonical-bodies-jackson.json` and `…-gson.json` deleted.** Bodies come from `JsonShim.java`/`GsonShim.java`'s `@ProveKitSugar` via the shim `.proof`. Proven: deleting the JSON pre-change → both sites refuse; post-change → resolve + compile.

### PIECE 2 — kit-owned assembly (`5fc88502f`)
- Same-language `--out-dir` now routes through the Java kit's `assemble` RPC (`dispatch_assemble` + `run_compile_check_with_classpath`), same as the cross-language path. The kit emits package + real `import`s + hoisted helper field + class wrap; the CLI writes back what the kit returns. **javac exits 0** for jackson and gson.
- The legacy CLI carrier-assembly path is removed for assemble-capable kits; kits without the RPC fall back unchanged via `AssembleError::MethodNotSupported`. `--write`/in-place + dry-run paths untouched.

## Test plan
- [x] jackson + gson: materialize emits faithful library code (`MAPPER.readTree`/`writeValueAsString` vs `JsonParser.parseString`/`GSON.toJson`), assembled by the kit, **javac 0**, templates loaded from the shim `.proof` (JSON-moved-aside test passes), the two JSON files deleted
- [x] No regression: `cmd_materialize_integration` (16, incl. TS-sqlite `--out-dir`), `library_tag_dispatch_test` (2), `go_realize_materialize` (2), `registry_authority_dispatch_test` (7) green
- [x] Other kits keep their disk JSON + fallback (no kit deleted out from under itself)

## Per-kit fan-out (follows this PR)
1. Lift the shim's `@ProveKitSugar` into a `.proof`; 2. add `sugar_proof` to the realize manifest; 3. kit's realize RPC prefers `bodyTemplates`; 4. kit implements the `assemble` RPC; 5. delete that kit's `<lang>-canonical-bodies-<tag>.json`. Safe/incremental: a kit keeps working on disk-fallback until all five are done.

Non-blocking follow-ups: same-language signature derivation from the shim's `return_sort_cid` morphism (demo consumer uses `Object` supertype today); `extract_package_hint` should defer to the kit.

Ref: #1436. Goal: no shim authored in JSON; `.proof` the only substrate disk artifact; all else RPC.

## Summary by CodeRabbit

* **Refactor**
  * Replaced on-disk canonical body template files with RPC-based proof templates for JSON serialization/deserialization.
  * Updated `provekit materialize` to route output assembly through kit-owned RPC when available, with fallback to legacy substrate concatenation.
  * Improved sugar proof integration for library-specific template selection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1458?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TSavo:
## Independent review (fresh eyes) — .proof-load-via-RPC + kit-owned assembly (5fc88502f)

**Verdict: APPROVE-WITH-NITS.** The design pattern is sound and the surrounding evidence is clean: the deleted JSON is genuinely unreachable, the `.proof`-load wiring is real and additive, the kit owns assembly (CLI doesn't, with a clean legacy fallback), the template factoring is shared (not duplicated), and no other kit's behavior changed. Two honest caveats: I could not independently reproduce the jackson/gson end-to-end (Java jar build constraint in my sandbox), and the foundational load path lacks unit coverage.

### 1. The `.proof`-load is REAL and the deleted JSON is unreachable
Both `java-canonical-bodies-{jackson,gson}.json` are genuinely deleted (not on disk, not in index, not at HEAD). Grep finds ZERO references to `canonical-bodies-jackson`/`gson` anywhere in `implementations/rust` or `menagerie`, and no stale copy or `readTree`/`parseString` string in `target/` build artifacts -- so they cannot be silently re-resolved. The templates MUST come from `body_templates_from_shim_proof` (cmd_materialize.rs:2006, called at :1818) over the RPC `bodyTemplates` param, or materialize emits stubs. The wiring is well-designed: `RealizeRequest.body_templates` is `skip_serializing_if = Vec::is_empty` (absent for non-migrated kits -> no wire/CID pollution); child realize calls inherit it (no mid-tree disk fallback); SugarRealizer.entries() prepends RPC entries (first-match-wins -> RPC authority) then `addAll(diskEntries())`, blank -> disk fallback.

**Caveat (trust-on-author for me):** I could NOT independently reproduce `materialize --library jackson`/`gson` end-to-end -- the Java realize jar won't build in my sandbox (offline maven cannot resolve `javaparser`, the same constraint I hit in #1448). The PR's `examples/json-shim-demo-client/CAPTURED-OUTPUT.txt` (showing `MAPPER.readTree` / `JsonParser.parseString` emitted with the JSON deleted) is the AUTHOR's capture, not my reproduction. I verified everything around the headline; the jackson/gson byte-emission itself is the author's evidence.

### 2. Byte-equivalence pre/post
Same constraint -- I cannot run the materialize to diff the emitted bodies. The claim (JSON was a byte-identical projection of the shim `.proof`, so output is identical) is structurally plausible because point 5 confirms the SAME `${paramN}`-rewrite function feeds both paths, but I did not empirically diff. Trust-on-author for this reviewer.

### 3. Kit-owned assembly is correct -- CLI does NOT assemble (clean fallback)
`cmd_materialize.rs:1036-1095`: `dispatch_assemble` Ok -> writes the kit's returned files VERBATIM (target-owned assembly), collects kit-declared classpath, sets `wrote_assembled` so the `continue` at :1084 skips the legacy concat. `Err(MethodNotSupported)` -> falls THROUGH to the UNCHANGED legacy concat path (:1087-1094) -- the legacy carrier-assembly is preserved verbatim, NOT a forked branch with behavior drift. `Err(Failed)` -> warns + also falls through. This is the right additive escape hatch; the design won't drift across the 21-kit fan-out.

### 5. `binding_entry_to_template_entry` factoring is shared, not duplicated
Defined ONCE (`cmd_mint.rs:1439`, `pub(crate)`), IMPORTED by `cmd_materialize.rs:2011`. One function, two consumers (cmd_mint's disk projector + materialize's `.proof`-load), so the `${paramN}` rewrite is guaranteed identical between projector output and the `.proof` path -- no synchronization bomb. `provekit-cli` lib 36/36 green (covers the shared fn + projector).

### 4. Other kits untouched (additive)
All named suites green with the exact counts: `cmd_materialize_integration` **16** (incl. TS-sqlite), `library_tag_dispatch_test` **2**, `go_realize_materialize` **2**, `registry_authority_dispatch_test` **7**. These exercise the non-migrated kits (TS/Go), proving SugarRealizer's blank-`bodyTemplates`->disk fallback keeps them working in production. No other kit's behavior changed.

### NIT (track for the 21-kit fan-out, not a blocker)
The foundational load path `body_templates_from_shim_proof` has NO rust unit test, and the jackson/gson verification is a MANUAL capture (`CAPTURED-OUTPUT.txt`), not an automated test. For the PR that establishes the migration pattern ~21 kits will follow, the load path + at least one migrated kit's byte-equivalence deserve automated coverage (a unit test on `body_templates_from_shim_proof` asserting the projected entries match the projector's, and an integration test driving jackson/gson materialize behind the existing `if java_available()` guard like the other gauntlet tests). Without it, a regression in the load path would only surface as a manual demo drift. The integration suites cover the FALLBACK (disk) kits but nothing covers the new RPC-authority path automatically.

---

**Bottom line:** the design is the right shape -- `.proof` is the authority via RPC, the deleted JSON is unreachable, the kit owns assembly with a clean legacy fallback, the template factoring is shared, and non-migrated kits are demonstrably unaffected. The pattern will fan out cleanly. **APPROVE-WITH-NITS:** add automated coverage of the `.proof`-load path before/as the 21-kit migration proceeds, and note that the jackson/gson byte-emission headline is the author's captured evidence (I verified everything around it but could not reproduce the Java-jar end-to-end in my sandbox). Worktree left clean (no push).

PR #1459, opened 2026-05-24
Kill canonical-bodies JSON: all 4 remaining Java library shims → .proof-load

TSavo:
## Summary

Migrates the remaining four Java library shims off their canonical-bodies JSON, following the #1458 template (Java kit already loads `bodyTemplates` from the shim `.proof` via RPC + assembles via RPC). **All four `java-canonical-bodies-{bouncycastle,java-io,provekit-rfc8785-jcs-java,sqlite-jdbc}.json` deleted (1,172 lines of hand-authored JSON gone).** Bodies now come from each shim's `@ProveKitSugar` `.proof`. Base `java-canonical-bodies.json` (core ops) untouched.

Per shim: native `@ProveKitSugar` shim (all existing — `provekit-shim-{blake3-java→bouncycastle, stdio-java→java-io, rfc8785-jcs-java, java-sqlite-jdbc}`), `sugar_proof` manifest pointer, JSON deleted, `cmd_materialize_proof_load` test copied. Materialize-from-`.proof` verified for all; javac-0 for three.

### Notable
- **Substrate-honesty win**: the deleted `sqlite-jdbc` JSON carried a pre-existing `java.${param1}.Connection` placeholder bug; the lifted `.proof` produces correct imports. Killing the hand-authored JSON fixed a latent bug.
- **Resolver fix**: `resolve_library_surface` unconditionally stripped the `{target}-` prefix, truncating the literal tag `java-io`→`io` (no manifest → silent stub-refuse). Now manifest-aware (`realize_tag_exists`), prefers the full tag when a realize manifest declares it. Only affects tags literally starting with `{target}-`; jackson/gson/etc. unaffected.
- **sqlite-jdbc compile-check deferred** (not a regression): the shim propagates checked `java.sql.SQLException` via `throws`; the Java realize kit doesn't emit `throws` clauses yet, so the assembled unit doesn't javac-compile. Realize-kit follow-up; its test asserts bodies-from-`.proof` + a `no java.${` discriminator instead of compiling.

## Test plan
- [x] All 4 JSONs deleted; materialize emits correct library bodies from the `.proof` (move-aside → refuse, proving `.proof`-load)
- [x] javac-0 for bouncycastle / java-io / rfc8785-jcs (sqlite-jdbc deferred per above)
- [x] `cmd_materialize_proof_load` (6), `cmd_materialize_integration` (16), `library_tag_dispatch_test` (2), `cargo build --workspace` green
- Pre-existing (not this PR): `JavaNullBoundaryRealizerTest...arithmeticSequence` fails identically on base; `sqlite-jdbc` throws-clause kit gap tracked.

Ref: #1436, #1458 (template). Goal: no shim authored in JSON; `.proof` the only substrate disk artifact.

## Summary by CodeRabbit

* **New Features**
  * Added example demo clients demonstrating Blake3 hashing, JSON canonicalization, SQLite database operations, and standard I/O capabilities.

* **Tests**
  * Expanded proof-loading test coverage to verify multiple library implementations via runtime RPC delivery.

* **Chores**
  * Migrated proof material delivery from disk-based JSON templates to runtime RPC loading.
  * Updated manifest configurations to reference new proof sources.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1459?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


TSavo:
## Independent review (fresh eyes) — 4 Java shim migrations off canonical-bodies JSON (37e1ab539)

**Verdict: APPROVE-WITH-NITS.** Follows the #1458 template cleanly: the 4 JSONs are genuinely deleted + unreachable, the resolver fix is sound and non-regressive, and the sqlite-jdbc import-fix + deferral are honest by the test's logic. Same caveat as #1458: I could not independently reproduce the per-shim end-to-end (Java jar won't build in my sandbox), so the headline body-emission is the author's captured evidence. No spine touched.

### 1. `.proof`-load setup genuine -- the 4 JSONs are deleted + unreachable
All four (`bouncycastle` 29, `java-io` 74, `rfc8785-jcs` 78, `sqlite-jdbc` 991 lines) deleted from disk. The ONLY remaining reference is the test helper `assert_no_canonical_bodies_on_disk(tag)` (which asserts they're GONE) -- no production code in `provekit-cli`/`libprovekit`/`menagerie` re-resolves them. So a passing body assert genuinely == RPC-authority path from the shim `.proof`.

**Caveat (trust-on-author, same as #1458):** the Java realize jar will not build in my sandbox (offline maven cannot resolve `javaparser`), so ALL 6 `cmd_materialize_proof_load` tests SKIP via their `if !java_realize_jar().exists()` guard -- they report `ok` but are guard-skipped, not executed. I therefore did NOT independently reproduce `materialize --library {bouncycastle,java-io,rfc8785-jcs,sqlite-jdbc}` emitting correct bodies. The test ASSERTIONS are correct as written (each pins library-specific bodies + JSON-deleted); the end-to-end execution is the author's evidence.

### 2. sqlite-jdbc import-fix is real (verified by the test's discriminator logic)
The test asserts the CORRECT literal imports (`import java.sql.Connection;` + `import java.sql.PreparedStatement;`) AND the discriminator `!emitted.contains("java.${") && !emitted.contains("javax.${")` -- proving the deleted JSON's buggy `java.${param1}.Connection` placeholder form is gone, replaced by the `.proof`'s correct literal imports. Plus `!emitted.contains("${param")` (all substituted). This is the right check for "the `.proof` fixes the JSON's placeholder bug, not a regression." (Verified by reading the assertion logic; not executed -- Java-jar caveat.)

### 3. Resolver fix is sound + non-regressive
`resolve_library_surface` now prefers the FULL tag (`java-io`) only when `realize_tag_exists(project_root, target, library)` confirms a manifest declares exactly that tag; otherwise it falls through to the existing `{target}-` strip. So:
- `java-jackson`/`java-gson`: no manifest with literal tag `java-jackson` (tag is `jackson`) -> `realize_tag_exists` false -> strips to `(java, jackson)`. UNCHANGED.
- `java-io`: manifest declares tag `java-io` -> kept full -> `(java, java-io)`. Fixed (was truncated to `io`).
- neither matches -> bare strip (back-compat for unregistered tags).
Only tags literally starting with `{target}-` that ALSO have a matching manifest change resolution. Empirically non-regressive: `cmd_materialize_integration` 16/16, `library_tag_dispatch_test` 2/2, `registry_authority_dispatch_test` 7/7 all green.

### 4. sqlite-jdbc compile-check deferral is honest
The test uses `materialize_client_no_compile_check` -- it asserts bodies-from-`.proof` (correct + placeholder-free) but does NOT run javac, explicitly because the shim throws checked `SQLException` and the kit doesn't emit `throws` clauses yet (a documented real kit gap). "Deferred" genuinely means "not compiled" (compile verification left for when the kit grows throws support), NOT "compiled-wrong-and-ignored." Honest posture.

### 5. The arithmeticSequence failure is pre-existing, not introduced
This PR touches NO Java realizer source (only deleted JSONs, realize manifests, demo clients, `cmd_materialize.rs` resolver, and the test file). So `JavaNullBoundaryRealizerTest...arithmeticSequence` cannot be introduced here -- it's pre-existing on base.

### NIT (track for the fan-out)
Same coverage gap as #1458: the per-shim `.proof`-load verification depends entirely on the Java jar being built, with no rust-side unit test of the `body_templates_from_shim_proof` extraction for these tags (the kind that would run in a jar-less environment). For the migration fan-out, a rust-level assertion that the deleted JSON's content matches the shim `.proof`'s projected entries (jar-independent) would let CI/reviewers verify the dead-equivalence without the full Java toolchain. The resolver fix also has no dedicated unit test (covered only via integration).

---

**Bottom line:** the 4 JSONs are genuinely deleted and unreachable, the resolver fix is surgical and non-regressive (jackson/gson untouched), the sqlite-jdbc import-fix discriminator is the right check and the deferral is honest, and the arithmeticSequence failure is pre-existing. The only thing I could not run is the jar-gated per-shim end-to-end -- the author's `CAPTURED-OUTPUT`/skipped tests are the evidence there. **APPROVE-WITH-NITS** (Java-jar reproduction caveat + jar-independent coverage nit). Worktree left clean (no push).

PR #1460, opened 2026-05-25
Migrate path loads body-templates from the shim .proof via RPC (shared enabler)

TSavo:
## Summary

The shared enabler for killing every **migrate-consumed** shim's canonical-bodies JSON. PR #1458 wired `.proof`-load into `materialize`; this does the same for the `bind`/migrate path, which previously dispatched realize without the shim templates (so a migrate into a target library still depended on that library's on-disk JSON).

- Factored the `.proof`-load attach logic out of `cmd_materialize.rs` into shared `pub(crate)` helpers: `body_templates_from_shim_proof` (exposed) + new `attach_body_templates_from_shim_proof(spec, project_root, target_lang, library_tag)`. `cmd_materialize` calls the new helper; behavior unchanged.
- Wired it into `cmd_bind_migrate` at the single choke point `realize_probe_via_path` (before RPC dispatch). One attach point covers every migrate realize site: `probe_realize_binding`, `realize_callsite_bodies` (via `realize_request_for_callsite`), `MigrateKit::realize_target`, the d5 harness.

## Verification
- Unit test `migrate_realize_spec_carries_shim_proof_body_templates` runs against the **on-main java `sqlite-jdbc` shim `.proof`** — whose canonical-bodies JSON was deleted in #1459, so a non-empty template result can *only* come from the `.proof`. Asserts: helper extracts non-empty templates for `(java, sqlite-jdbc)`; the attach point inserts them under `bodyTemplates`; empty `library_tag` is a no-op (disk fallback preserved).
- No regression: `cmd_materialize_integration` 16/16, `cmd_materialize_proof_load` 6/6, `cross_platform_point_query_receipt_test` 6/6, all 154 provekit-cli bin unit tests (incl. both d5 migrate-realize harness tests), `cargo build --workspace` clean.
- **Back-compat verified, not assumed**: the d5 harness now gets `bodyTemplates` attached for python-sqlite3 and its output is unchanged — empirically confirming the python kit on main ignores them and falls back to its disk cache (until the python kit's prefer-RPC lands).

No JSON deleted (enabler only — per-kit deletions follow, rebased onto this).

Ref: #1436, #1458. Goal: no shim authored in JSON; `.proof` the only substrate disk artifact; all exchange RPC.

### DIFF
--- a/implementations/rust/provekit-cli/src/cmd_bind_migrate.rs
+++ b/implementations/rust/provekit-cli/src/cmd_bind_migrate.rs
@@ -1050,8 +1050,22 @@ fn realize_probe_via_path(
     repo_root: &Path,
     language: &str,
     tag: &str,
-    spec: JsonValue,
+    mut spec: JsonValue,
 ) -> Result<libprovekit::core::RealizedSource, String> {
+    // `.proof`-load-via-RPC for the MIGRATE path: this is the single choke
+    // point every migrate realize dispatch funnels through (the binding probe,
+    // per-callsite body realization, and the MigrateKit's `realize_target`).
+    // Attaching the target shim's signed binding-entry templates here mirrors
+    // what `cmd_materialize` does for the materialize path, so a migrate into a
+    // target library no longer depends on that library's on-disk
+    // `<lang>-canonical-bodies-<tag>.json`: the @ProveKitSugar shim `.proof` is
+    // the authority, carried over RPC. Absent a `sugar_proof` pointer (or for a
+    // kit not yet preferring RPC), this is a no-op and the disk cache remains
+    // the fallback. `repo_root` is the project root; `(language, tag)` are the
+    // target lang/library tag.
+    crate::cmd_materialize::attach_body_templates_from_shim_proof(
+        &mut spec, repo_root, language, tag,
+    );
     let mut inputs = HashMapInputCatalog::default();
     let input_cid = inputs.insert(Input::Spec(spec));
     let kit_name = format!("lower-{language}");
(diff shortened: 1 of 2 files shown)


TSavo:
## Independent review (fresh eyes) — migrate `.proof`-load enabler (6162f9bfd)

**Verdict: APPROVE.** The factoring is behavior-preserving, the choke point is complete (all migrate realize sites funnel through it), the unit test genuinely proves migrate `.proof`-load, and back-compat holds. Spine untouched (only `cmd_materialize.rs` + `cmd_bind_migrate.rs` changed).

### 1. cmd_materialize factoring is behavior-preserving
The extracted `attach_body_templates_from_shim_proof` is logically identical to the old inline block:
- old `if !templates.is_empty()` -> new `if templates.is_empty() { return; }` (same gate, inverted-early-return).
- old `if let Some(obj) = spec.as_object_mut() { obj.insert("bodyTemplates", ..) }` -> verbatim.
- the empty-`library_tag` guard moved INTO the helper; the materialize call site keeps its outer `if !carrier_library.is_empty()`, so it's doubly-guarded (harmless). `body_templates_from_shim_proof` itself is unchanged (only made `pub(crate)`). No logic dropped or reordered. materialize 16/16 + proof_load 6/6 confirm.

### 2. The choke point genuinely covers ALL migrate realize sites
`realize_probe_via_path` attaches templates to `mut spec` BEFORE the single `Input::Spec(spec)` dispatch (line 1070) and the `lower-{language}` kit construction (line 1071) -- both of which exist ONLY inside this function. Every production migrate realize dispatch routes here:
- `probe_realize_binding` -> `realize_probe_via_path` (line 1040)
- `realize_callsite_bodies` / `realize_request_for_callsite` -> (line 1536)
- `MigrateKit::realize_target` -> (line 2807)
No realize path bypasses the funnel, so no site silently lacks `bodyTemplates`.

### 3. The unit test genuinely proves migrate `.proof`-load
`migrate_realize_spec_carries_shim_proof_body_templates` (green, in the bin target) uses the real on-main java `sqlite-jdbc` shim whose canonical-bodies JSON was DELETED in #1459 -- so the non-empty `body_templates_from_shim_proof(java, sqlite-jdbc)` extraction can ONLY come from the shim `.proof`. It then asserts the shared attach inserts `bodyTemplates` with matching count into a migrate-probe-shaped spec. Real `.proof`-load, not a stale source.

### 4. Back-compat holds (empty/ignored templates are a no-op)
Negative control in the test: empty `library_tag` -> no `bodyTemplates` attached (disk-cache fallback). And `cross_platform_point_query_receipt_test` 6/6 with the python-sqlite3 JSON present: the d5 harness now gets `bodyTemplates` attached for python-sqlite3, but the on-main python kit ignores them (disk fallback), so output is UNCHANGED. Genuine back-compat -- ignored templates are a no-op, not a masked break.

### 5. No regression
provekit bin unit suite **154/154**; `cmd_materialize_integration` 16; `cmd_materialize_proof_load` 6; `library_tag_dispatch_test` 2; `cross_platform_point_query_receipt_test` 6; `cargo build --workspace` 0 errors.

---

**Bottom line:** the factoring is a clean extraction (verbatim attach logic, shared `pub(crate)` helper), the migrate attach is wired at the one funnel every realize dispatch passes through, the test proves real `.proof`-load against a JSON-deleted shim, and not-yet-migrated kits keep working via the disk fallback. The attach point + factoring are the right foundation for the python/rust/ts migrate-consumed deletions to rebase onto. **APPROVE.** Worktree left clean (no push).

PR #1461, opened 2026-05-25
Kill 4 dead Rust canonical-bodies JSONs + rust kit .proof-load-via-RPC foundation

TSavo:
## Summary

Deletes 4 Rust canonical-bodies JSONs (verified DEAD — none reachable by any live route) and adds the rust kit's `.proof`-load-via-RPC foundation (mirroring #1458's Java SugarRealizer). Base `rust-canonical-bodies.json` (core ops) untouched.

### The 4 deletions — all verified safe/dead
- `serde_json.json` — actually carried a **misfiled** `concept:rfc8785-jcs-encode` (served by the live rfc8785-jcs shim); serde-json's real `json-parse`/`json-serialize` live in its shim `.proof`.
- `blake3.json` / `std::io.json` — all concepts present in `provekit-shim-{blake3-rust,stdio-rust}/.proof`.
- `libprovekit-rpc-cross-platform.json` — pure orphan (no shim/manifest/route; only a test-fixture `@sugar` label).
- **Decisive**: `materialize --library serde_json` *refuses* — `serde_json`/`std::io` are structurally invalid routing tags (`[a-z][a-z0-9-]*`), and no `library-bindings` surface routes these concepts. Nothing live reached these disk caches.

### Prefer-RPC foundation (`28f4f1ae4`)
Rust kit prefers RPC `bodyTemplates` over disk, RAII `BodyTemplatesGuard` against thread-local leak, disk fallback intact. Mirrors SugarRealizer (#1458).

## Test plan
- [x] `cargo build --workspace` clean
- [x] seam4 federation census 9/9, `cmd_verify_rust_division_unsound` 4/4 (deletions don't touch the dogfood/federation)
- [x] `cmd_materialize_integration` 16/16, `cmd_materialize_proof_load` 2/2, production bridges 6/6, migrate async 1/1
- [x] New regression `shim_proof_carries_rust_bodies_after_disk_deletion.rs` (4) — asserts shim `.proof` carries the concepts the deleted JSONs held, pins disk-absence; re-introducing a JSON fails it
- [x] No migrate consumption of any deleted tag

## Follow-up (not a blocker, tracked)
The rust prefer-RPC path has no live route yet: rust manifests lack `sugar_proof`, and shim `.proof` `target_library_tag` (`serde_json`) mismatches manifest `library_tag` (`provekit-shim-*-rust`). Wiring a live rust consumer (manifest `sugar_proof` + tag alignment) is a separate enabler.

Ref: #1436, #1458.


TSavo:
## Independent review (fresh eyes) — rust track: 4 orphan JSON deletions + prefer-RPC foundation (fbf0c859f)

**Verdict: APPROVE.** The 4 deletions are genuinely dead/unreachable, the dogfood federation is undisturbed, and the prefer-RPC code mirrors the reviewed #1458 pattern with the #1455 leak guard. Notably this track has a JAR-INDEPENDENT regression test that actually runs (the coverage gap I flagged on #1458/#1459 is closed here). No spine touched.

### 1. The 4 deletions are genuinely DEAD (not live-but-broken)
`validate_library_tag` (kit_dispatch.rs:1017) enforces `[a-z][a-z0-9-]*` on every manifest `library_tag` (called at load, :932/:1224). So:
- `serde_json` (underscore) and `std::io` (colons) CANNOT be valid routing tags -- they fail validation, and `resolve_library_surface` refuses them. No live route could reach those caches.
- The actual rust realize manifests use VALID tags (`rust-shim-blake3`, `rust-shim-serde-json`, `rust-shim-stdio`, ...) and carry their concepts in the shim `.proof`, NOT the deleted JSONs (which held the wrong/orphan tags `blake3`, `serde_json`, `std::io`).
- `libprovekit-rpc-cross-platform` is an orphan (test-fixture label only); `blake3`/`serde_json`/`std::io` are superseded by the `rust-shim-*` tags.
- The `serde_json` JSON literally carried `concept:rfc8785-jcs-encode` (a misfiled concept), confirming it was never serde-json's real authority.
Bounded grep confirms the only remaining reference to any of the 4 is the regression test asserting they're GONE -- no production re-resolution path.

### 2. No regression to dogfood/federation
The deletions sit near the substrate's own rust shims, so I checked the federation invariants directly: seam4 trinity census **9/9** (byte-identity intact) and `cmd_verify_rust_division_unsound` **4/4** (the rust division-soundness guard, the analog of the Go/Python cardinal-sin guards, undisturbed). The deletions touched neither.

### 3. Prefer-RPC mirrors #1458 (with the #1455 leak guard)
`operator_body_template_for` tries `rpc_body_template_entries()` FIRST (authority); on no match, falls through to the disk cache (back-compat) -- the same shape as java `SugarRealizer.entries()`. `CURRENT_BODY_TEMPLATES` is written ONLY through `BodyTemplatesGuard` (RAII), whose `Drop` restores the prior value (panic-safe) -- explicitly modeled on `VisibilityGuard` to prevent the cross-request thread-local leak that would corrupt seam4 byte-identity. The base `rust-canonical-bodies.json` (core ops, routed via `body_template_for`) is UNTOUCHED -- only the 4 tag-specific files deleted. realize-rust-core lib **38/38** green both parallel AND `--test-threads=1`, confirming the guard introduces no order-dependence/leakage. Build clean.

### 4. Regression test genuinely proves the dead-equivalence -- and is jar-independent
`shim_proof_carries_rust_bodies_after_disk_deletion` **4/4** (executed, NOT jar-skipped): `serde_json`/`blake3`/`stdio` shim `.proof` supplies the bodies the deleted JSONs held (read via the EXACT rust decode path cmd_materialize uses), and `libprovekit_rpc_cross_platform_orphan_cache_stays_deleted` pins the orphan deletion. Re-introducing any of the 4 fails the test. This runs WITHOUT the Java toolchain -- exactly the jar-independent dead-equivalence coverage I noted was missing on the Java tracks (#1458/#1459). Good model for the fan-out.

### The #1462 follow-up framing is honest
The prefer-RPC path has no LIVE route yet (rust manifests lack a `sugar_proof` pointer + the tag alignment), so this PR is genuinely dead-JSON-cleanup + foundation, NOT "live route broken." The test header documents this explicitly (lines 31-41), and points 1+4 confirm it: the deleted JSONs really were dead (invalid/orphan tags), so deleting them changes no live behavior; the prefer-RPC code is dormant until #1462 wires the route.

---

**Bottom line:** the deletions are verified-dead (structurally-invalid + orphan tags, no live route), seam4 9/9 and rust division 4/4 hold, the prefer-RPC mirrors #1458 with the #1455 RAII leak guard and an untouched base JSON, and the regression test is jar-independent + actually runs. Cleaner than the Java tracks. **APPROVE.** Worktree left clean (no push).

PR #1531, opened 2026-05-26
fix: ETXTBSY race + bundle SQLite (greens main; CICP removal lands next)

TSavo:
## Why main is red

The revert (#1530) backed out the LSP ports cleanly, but main's conformance gate stayed **red** on a single failure:

```
test python_mint_auto_writes_body_discharge_bridge ... FAILED
  stderr: kit transform failed: spawn ["/tmp/provekit-lift-python-verify-43642.sh", "--rpc"]: Text file busy (os error 26)
==== test-all FAIL: test-rust ====
```

`make conformance` (C1-C8) is **green**; java/ts/ruby/C all green. Only this one test fails, intermittently.

## Root cause

`cmd_verify_python_production_bridge.rs` (3 tests) and `cmd_verify_python_division_unsound.rs` (2 tests) each run as parallel threads of **one** process. `build_python_lift_verify()` keyed the wrapper-script path on `process::id()` alone — so all parallel tests wrote and exec'd the **same** `/tmp/provekit-lift-python-verify-<pid>.sh`. One test exec'ing it while another truncates it for write => **ETXTBSY**.

## Fix

- Append a process-wide `AtomicU64` counter to the path → unique per call, no cross-test collision.
- Adopt the `sync_all`+drop-before-chmod write already proven in `cli_surface.rs::write_executable` → closes the open-writer-fd half of the race.

Test-only; no production code touched.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for Python lift verification by enhancing temporary script generation to prevent collisions during concurrent test execution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1531?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/examples/provekit-shim-rusqlite/Cargo.toml
+++ b/examples/provekit-shim-rusqlite/Cargo.toml
@@ -9,7 +9,12 @@ keywords = ["sqlite", "rusqlite", "provekit", "proofchain", "boundary-realizatio
 categories = ["database"]
 
 [dependencies]
-rusqlite = "0.39"
+# `bundled`: compile SQLite from the amalgamation and link it statically, so
+# the build never depends on a system `libsqlite3.so` being discoverable by the
+# linker. The self-hosted runner fleet is heterogeneous (lld doesn't always get
+# `/usr/lib/x86_64-linux-gnu` on its search path), which intermittently broke
+# `-lsqlite3` resolution in build-rust. Bundled makes the build runner-independent.
+rusqlite = { version = "0.39", features = ["bundled"] }
 provekit = { package = "provekit-sugar", path = "../../implementations/rust/provekit-sugar" }
 
 [dev-dependencies]
--- a/examples/rusqlite-demo-consumer/Cargo.toml
+++ b/examples/rusqlite-demo-consumer/Cargo.toml
@@ -10,4 +10,5 @@ name = "rusqlite-demo"
 path = "src/main.rs"
 
 [dependencies]
-rusqlite = "0.39"
+# bundled: static SQLite, no system libsqlite3.so link dependency (runner-independent).
+rusqlite = { version = "0.39", features = ["bundled"] }
--- a/implementations/rust/provekit-cli/tests/cmd_verify_python_division_unsound.rs
+++ b/implementations/rust/provekit-cli/tests/cmd_verify_python_division_unsound.rs
@@ -83,17 +83,29 @@ fn unique_dir(suffix: &str) -> PathBuf {
 }
 
 fn build_python_lift_verify() -> PathBuf {
+    use std::io::Write as _;
+    use std::sync::atomic::{AtomicU64, Ordering};
+    // Unique per call: parallel tests in this binary share one process id, so a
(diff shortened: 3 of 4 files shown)


TSavo:
## Second fix added to this branch

While CI ran the first fix, the conformance gate surfaced a **separate** flake: `build-rust` died with `rust-lld: error: unable to find library -lsqlite3` — even though ci.yml installs `libsqlite3-dev`.

Root cause: the self-hosted runner fleet is heterogeneous, and rust's `-fuse-ld=lld` (via the `gcc-ld` wrapper) doesn't reliably get `/usr/lib/x86_64-linux-gnu` on its linker search path. The revert run linked on a runner where it did; this run didn't. That fragility is why the gate has been intermittently red.

Fix (commit 5b6250ef8): enable rusqlite's `bundled` feature in the shim crate (pulled into `provekit-realize-rust-core`) and the demo consumer. `libsqlite3-sys` then compiles the SQLite amalgamation and statically links it — no system `libsqlite3.so`, no `-lsqlite3` search-path dependency, build is runner-independent. `tokio-postgres` is pure-Rust; no other crate declares `links =`.

Both fixes are independently understandable; together they take the conformance gate from flaky-red to deterministically green.

TSavo:
## Folded CICP in (per T): the green build should mint the witness, not throw it away

`make conformance` already passed on the bundled-SQLite run (build-rust links statically now — the `-lsqlite3` fix is proven). Rather than green this without CICP and then rebuild from scratch to mint the first witness, CICP is now on this branch (commit 20a696365, codex ba187d26's reuse-loop fix):

- `cmd_ci.rs`: `ci accept` admits a real candidate (replaces a refused witness).
- `ci.yml`: `cicp-precheck` computes reuse admission; the conformance gate `needs` it with `if: always() && !cancelled()` (fails **open** — precheck fail => full run); every build/test step gated `if: CICP_REUSE_HIT != 'true'`; staleness smoke now **hard-fails** on stale witnesses; `cicp-refresh` writes the witness back.

No duplicate jobs (verified against main's existing CICP wiring). Fail-open means worst case is a normal full build. When this goes green it runs `ci accept` and captures the witness for its blast-radius CID — so a subsequent no-change run reuses it and skips the build. That's the payoff in one shot.

TSavo:
Un-bundled the CICP commit per T's call to **remove CICP entirely**. This PR is now just the two flake fixes (ETXTBSY + bundled SQLite) — `make conformance` already passed with both, so this greens main. The complete CICP excision (code + ci.yml jobs + witness store + protocol spec de-catalog/re-sign + conformance vectors + docs) lands as a focused PR on this green baseline. codex's reuse-loop fix is preserved on `codex/close-cicp-reuse-loop` but moot now.

TSavo:
Subsumed by the omnibus PR #1532 (chore/remove-cicp), which bundles both flake fixes (ETXTBSY + bundled SQLite) plus the full CICP + rpc-cross-platform removal. Branch preserved as fallback. Closing to free CI fleet capacity for #1532.

PR #1564, opened 2026-05-27
fix(solvers): quote/sanitize special-char symbols across smt-lib, coq, and lean backends

TSavo:
Lifted ctor names like `go:call` / `go:slice-literal` contain `:`, which is illegal in an SMT-LIB *simple symbol* — z3 rejected them ("invalid expression, unexpected keyword"), which surfaced as a spurious `undecidable`. They now emit as quoted symbols (`|go:call|`) at both the ctor application and its `declare-fun`, so z3 parses and reasons about them as uninterpreted functions.

Verified: the byte-order composition obligation `post(serialize) ⟹ canonical(out)` now parses and z3 returns `sat` on its negation — the obligation is genuinely *not valid* (a real seam, not a syntax error). smt-lib compiler tests green.

`generated.rs` is CDDL-generated; mirror in `tools/generate-from-cddl.py` on regeneration.

## Summary by CodeRabbit

**Bug Fixes**
- Compiler backends now properly handle special characters in identifiers across Coq, Lean, and SMT-LIB targets
- Improved symbol compatibility to ensure correct translation to each target verification system

[Review Change Stack](https://app.coderabbit.ai/change-stack/TSavo/provekit/pull/1564?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/implementations/rust/provekit-ir-compiler-coq/src/generated.rs
+++ b/implementations/rust/provekit-ir-compiler-coq/src/generated.rs
@@ -27,12 +27,12 @@ pub fn emit_term(term: &Term) -> String {
         }
         Term::Ctor { name, args, .. } => {
             if args.is_empty() {
-                return name.clone();
+                return coq_ident(name);
             };
             let args_str = args.iter();
             let args_str = args_str.map(emit_term);
             let args_str: Vec<String> = args_str.collect();
-            format!("({} {})", name, args_str.join(" "))
+            format!("({} {})", coq_ident(name), args_str.join(" "))
         }
         Term::Lambda {
             param_name,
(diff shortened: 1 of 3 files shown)


TSavo:
Broadened scope: the same special-char-symbol bug existed in the **coq** and **lean** backends (T flagged "what about coq and lean and vampire?").

- **vampire / tptp**: routes through the SMT-LIB emitter, so already covered by the original quoting fix.
- **coq**: lifted ctor/predicate names like `go:call` / `go:slice-literal` contain `:`/`-`, which Coq rejects. Added `coq_ident` (→`_`), applied at both Ctor arms, the uninterpreted-predicate fallback, and the `Parameter` declaration. Verified: `Goal ((result = (go_slice_literal hi lo)) -> ((go_call "...CanonicalByteOrder" result) = true)).`
- **lean**: `lean_ident` already sanitized every *use* site but the four free-symbol binder *declarations* still emitted raw names (declared `(go:call : ...)`, used `go_call`). Sanitized the declarations to match.

Coq (3+15) and Lean (1+1+4) suites green, no regression.

PR #1695, opened 2026-05-30
fix(rust): lift test assertions as postconditions

### LINKED ISSUE #1593: rust-tests adapter promotes asserts to post (not inv) so discharge is substantive
Today provekit-lift-rust-tests mints each ContractDecl with `pre: None`, `post: None`, and the asserted formula in `inv` (lib.rs:70-72 documents this explicitly).

When the verifier discharges a callsite at f(x), it asks 'does the callee's pre hold?'. With `pre: None`, the discharge is vacuously satisfied — class=vacuous, solver=none. z3 is never invoked. **240/240 callsites in the snake-eats-tail empirical run discharged vacuously for exactly this reason.**

**Substrate-honest path:** promote each `assert_eq!(actual, expected)` to `post: actual == expected` on the contract that names the call. Then implications form properly — caller's pre + callee's post → callee's contract holds at the callsite, and z3 has a real obligation to discharge.

This is the gap that makes self-verification go from 'structurally enumerable' to 'structurally substantive.'

Touches: implementations/rust/provekit-lift-rust-tests/src/{lib.rs, layer2.rs} (the three ContractDecl::push sites at 419, 601, 773).

Related: PR #1591 (implication lifter is the consumer of these contracts), PR #1587 (snake-eats-tail run that surfaced this gap empirically).

TSavo:
Closes #1593.

## Summary
- promote rust-tests lifted assertions into `post` instead of `inv`
- route Layer 0, evidence, bounded-loop, helper-inline, and characterization declarations through one contract-shape helper
- keep `pre` and `inv` empty so test evidence participates as substantive postconditions instead of vacuous invariants

## Verification
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-lift-rust-tests lifts_simple_assert_eq_as_postcondition_not_invariant -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-lift-rust-tests -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-cli --test cmd_verify_rust_production_bridge rust_production_path_double_discharges_and_mints_witness -- --nocapture`
- `cargo test --manifest-path implementations/rust/Cargo.toml -p provekit-cli --test cmd_verify_rust_production_bridge rust_production_path_refuses_planted_contradictory_implication -- --nocapture`
- `git diff --check`

### DIFF
--- a/implementations/rust/Cargo.lock
+++ b/implementations/rust/Cargo.lock
@@ -1914,6 +1914,7 @@ name = "provekit-realize-rust-core"
 version = "0.1.0"
 dependencies = [
  "base64",
+ "proc-macro2",
  "provekit-canonicalizer",
  "provekit-ir-types",
  "provekit-proof-envelope",
@@ -1924,8 +1925,10 @@ dependencies = [
  "provekit-shim-rusqlite",
  "provekit-shim-serde-json-rust",
  "provekit-shim-stdio-rust",
+ "quote",
  "serde",
  "serde_json",
+ "syn",
 ]
 
 [[package]]
--- a/implementations/rust/provekit-lift-rust-tests/src/layer2.rs
+++ b/implementations/rust/provekit-lift-rust-tests/src/layer2.rs
@@ -80,7 +80,8 @@ use provekit_ir_symbolic::{
 
 use crate::{
     callsite_contract_name_pub, is_assertion_macro_pub, lift_assertion_macro_at_callsites_pub,
-    lift_assertion_macro_pub, path_to_string_pub, translate_term_pub, BoundCall, LiftWarning,
+    lift_assertion_macro_pub, lifted_assertion_contract_decl, path_to_string_pub,
+    translate_term_pub, BoundCall, LiftWarning,
 };
 use syn::spanned::Spanned;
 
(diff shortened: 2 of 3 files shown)


TSavo:
Closing per review. This diff relabeled rust-tests witnessed facts from `inv` to `post` without creating a dischargeable target precondition, so it did not make the obligation substantive. The useful direction is either keep witnessed test facts honest or design a distinct witnessed-evidence slot; production-function pre/post discharge belongs in the rust-fn-contracts path that already landed.

TSavo:
Closed by decision. Rationale for the record:

- This does not make discharge substantive: pre stays None, so a bridge to these contracts has no obligation to prove and still vacuous-passes. Moving the closed fact from inv to post relabels it, it does not discharge it.
- It mislabels test evidence: a point-test witness (e.g. square(3)==9 for input 3) is a closed observed fact, not a guarantee the function makes on every return. post reads as the latter.
- #1593's real fix already shipped in #1687 via the rust-fn-contracts surface, which lifts production functions' derived pre/post. That is where the actual z3/cvc5 discharges come from.

If a witnessed fact needs a precise home, that is a distinct evidence slot, not a reuse of post. The shared lifted_assertion_contract_decl helper can land separately if useful, without the inv->post change.

PR #1992, opened 2026-06-08
Behavior identity is reformat-invariant (canonicalize at hash time, bounded blast radius)

TSavo:
## What

`cargo sugar check` / `sugar diff` cried wolf on a behavior-preserving refactor: `fn double(x){ let n = x; n*2 }` got a *different* CID than `fn double(x){ x*2 }`, so an internal reformat looked like a behavior change. This closes that — the behavior **identity** is now invariant under reformatting (and bound-variable renaming), while a *real* change (`x*3`) still moves it.

Two halves, both CLI-side per the substrate law (**kits emit ProofIR; the rust CLI computes over it**):

1. **Lifter emits a faithful `let` (`sugar-walk`).** The rust syn lifter leaked a local binding name into the postcondition: `let n = x; n*2` emitted `result = n*2` with `n` a *free* variable (top-level lets were never bound into `LiftCtx`'s scope stack). `wrap_leading_lets` now re-attaches the leading immutable, liftable `let`s as a faithful `IrTerm::Let` over the result term, with the source's surface names — the kit emits *data*, it does not pre-resolve. A transitive-reference fixpoint drops lets the tail never touches, so a function with an unrelated local is byte-unchanged.

2. **Canonicalize at hash time (`sugar-claim-envelope`).** `contract_cid` (= the contract `header.cid`, which is exactly what `sugar diff` compares) and `contract_property_hash` now run pre/post/inv through `canonicalize_formula` (pure-let inlining + binder alpha-renaming) before hashing. The faithful `let` inlines to `result = x*2`, so the two surface shapes share one identity.

The verifier is unaffected: `emit_term` already lowers `IrTerm::Let` to a native SMT-LIB `(let (...) body)`, so the *stored* faithful form discharges. Canonicalization runs only at hash time; discharge sees the faithful term.

## Blast radius is bounded by construction

`canon_formula_value` returns the **original bytes** when canonicalization is a no-op (`if canon == formula { return v.clone() }`). So introducing this step **cannot move the content address of any contract without a `let`/binder**. The regime change is confined to exactly the contracts whose identity it must fix.

One pinned hash legitimately moved — a `forall n. n>0` precondition now alpha-normalizes (correct: `forall n. n>0` and `forall m. m>0` are the same proposition and must share an identity). That pin is re-blessed, and a new `property_hash_alpha_invariant_under_pre_rename` test captures the actual intent.

## Validation (all on battleaxe via bcargo)

- **`sugar-walk`**: 332 + 167 + integration tests green. New: `leading_let_reformat_shares_behavior_identity`, `unreferenced_leading_let_is_not_wrapped`.
- **`sugar-claim-envelope`**: 94 green. New keystone (deterministic, no lifter/solver): `reformat_with_local_shares_contract_cid_real_change_does_not` — at the exact layer `sugar diff` compares, `let n=x; n*2` shares the inline `x*2` CID and `x*3` does not. Plus `let_free_formula_is_byte_transparent`.
- **Full rust workspace**: 886 passed. (One daemon-socket IO flake under parallel load, `test_daemon_polyglot_smoke`, passes 5/5 in isolation — not a regression.)
- **numpy showcase**: PASS, and the regenerated mint CID is byte-identical with and without the canonicalization step (numpy contracts have no binders; numpy is python-kit-lifted, so the rust lifter change can't touch it).
- **rust-test-assertion-consistency** + **tokio-effect-consistency**: PASS — discharge survives the faithful-`let` emission (consistency discharged, contradiction refused, witness discharged).

## Note for review

This is a canonicalization-profile change to the content-address regime. Committed `.proof` baselines and the self-application that pin CIDs for binder/let-bearing contracts may need a re-bless on merge; the bounded-blast-radius guard keeps that set to exactly the binder/let contracts, and the full workspace + showcases above show nothing else moves.

## Release Notes

* **Bug Fixes**
  * Contract identity is now computed using canonical formula representation, ensuring that reformatting contract conditions (such as introducing variable bindings) does not change the contract's identity when behavior remains unchanged.
  * Variable renaming in quantified formulas no longer affects contract property hash.

* **Tests**
  * Added tests for behavior-versioning semantics and alpha-invariance.

### DIFF
--- a/implementations/rust/sugar-claim-envelope/Cargo.toml
+++ b/implementations/rust/sugar-claim-envelope/Cargo.toml
@@ -8,6 +8,7 @@ description = "Sugar: claim/memento envelope minters (contract, bridge, implicat
 
 [dependencies]
 sugar-canonicalizer = { path = "../sugar-canonicalizer" }
+sugar-ir-types = { path = "../sugar-ir-types" }
 sugar-proof-envelope = { path = "../sugar-proof-envelope" }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/implementations/rust/sugar-claim-envelope/src/lib.rs
+++ b/implementations/rust/sugar-claim-envelope/src/lib.rs
@@ -886,6 +886,14 @@ fn json_to_cvalue(v: &JsonValue) -> Arc<Value> {
 /// (a leaked `let` binding, a renamed bound variable) must hash to the same
 /// contract identity. If the value is not a parseable `IrFormula`, it is hashed
 /// unchanged.
+///
+/// BLAST-RADIUS BOUND: when canonicalization is a no-op (the formula has no
+/// `let` and no binder to normalize), the ORIGINAL `Arc<Value>` is returned
+/// byte-for-byte. Only formulas the canonicalizer actually rewrites get the
+/// re-serialized value. So introducing this step cannot move the content
+/// address of any contract that does not contain a `let`/binder -- the regime
+/// change is confined to exactly the contracts whose identity it must fix, and
+/// no CID drifts on a serde round-trip artifact.
 fn canon_formula_value(v: &Arc<Value>) -> Arc<Value> {
     let json: JsonValue = match serde_json::from_str(&encode_jcs(v)) {
         Ok(j) => j,
@@ -896,6 +904,10 @@ fn canon_formula_value(v: &Arc<Value>) -> Arc<Value> {
         Err(_) => return v.clone(),
     };
     let canon = sugar_ir_types::canonicalize_formula(&formula);
+    // No-op canonicalization -> preserve the original bytes exactly.
+    if canon == formula {
+        return v.clone();
+    }
     match serde_json::to_value(&canon) {
         Ok(j) => json_to_cvalue(&j),
         Err(_) => v.clone(),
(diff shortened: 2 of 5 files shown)


TSavo:
**Java validated** (it was missing from the original validation set; the envelope canonicalization is language-blind, so Java contracts flow through it too).

Ran on battleaxe (JDK 21), all green:
- `java-contract-breadth` — JSR-380 `@Max/@Size/@NotNull/@Min+@Max` consumers each **discharged** + witness discharged (these contracts mint through the canonicalized `contract_cid`).
- `java-implication-edge` — post(producer)→pre(consumer) edge **discharged** + witness discharged.
- `java-test-assertion-consistency` — junit consistency **discharged** + witness discharged.

Note: Java contracts are annotation/assertion-derived (JSR-380, JUnit), not lifted from arbitrary method bodies — so the rust lifter's leaked-local shape (the `let n=x; n*2` bug this PR fixes) doesn't arise on the Java path. The relevant Java question was *no regression from the language-blind canonicalization*, and these confirm it: every Java contract here is binder/let-free, so byte-transparent through `canon_formula_value`, and discharge is unaffected.

Full language coverage now: rust + python + java green.

TSavo:
**Request changes — not ready to merge.** This is a content-address-regime change (contract identity now routes through `canonicalize_formula`), so the bar is the highest. CI is red and there are two valid soundness findings in the new lifter code.

### What's right (the shape is sound)
- Correct separation: discharge sees the **faithful** term (`Let` lowers to SMT `(let …)`); canonicalization runs **only at hash time**, so it can never produce a false *discharge* — it only affects *identity*. Right place for it.
- **Bounded blast radius is verified-by-construction**: `canon_formula_value` returns the original bytes when `canon == formula`, so no let/binder-free contract can have its CID move.
- `canonicalize_formula` is pre-existing in `sugar-ir-types`; the PR wires it, doesn't invent it. The wiring is clean.

### Blockers
1. **CI is red (`make ci` ×2) while the body claims "886 passed."** Reconcile before merge. The "Note for review" admits `.proof` baselines + self-application pins "may need a re-bless," and only one pin was re-blessed — the red is most likely the rest of the binder/let pin set across the conformance/self-application gates. Confirm the red is *only* benign pin-rebless (not a real break) and complete the re-bless. A red `make ci` is not mergeable.

2. **`return`-form lets are not wrapped (incomplete fix).** `wrap_leading_lets` is called only in the tail-expr path (`lift_tail_expr_to_result_term`), **not** in `lift_return_stmt_postcondition`. So `fn f(x){ let n = x; return n*2; }` still leaks the free variable `n` — the exact bug this PR fixes, left unfixed for explicit `return`. Wrap leading-lets in the return path + add a regression test alongside `explicit_return_derives_result_postcondition`.

3. **Dependency-fixpoint shadowing bug.** The backward walk unions `fv(bound_term)` into `needed` but never removes the binding's own name. It should be `needed = (needed \ {b.name}) ∪ fv(bound_term)`; otherwise shadowed bindings are mis-retained and the canonical form is wrong. Fix + test.

### Extra-care (the reason this PR matters more than a test)
Contract identity now flows through canonicalization. The CID **is** the identity — a let-handling bug here is not a failed test, it's a potential **false-merge** (two different behaviors sharing one CID), which is worse than the false-alarm it fixes, and the multi-point conjoin work depends on identity being exactly right. Please add a discrimination test that `canonicalize_formula` inlines **only pure lets** — a side-effecting / non-pure let must keep a **distinct** CID, not silently merge. (The `x*3 ≠ x*2` keystone covers behavior-different; this covers impure-let non-inlining.)

Recommendation: fix (2) and (3) with regression tests, add the pure-let-only discrimination test, get `make ci` actually green (complete pin re-bless, confirm red was pins not a real break), then re-review.

TSavo:
Addressed all four review findings (de60a449a) — agreed with every one; they were real soundness holes in the happy-path-only first cut. Each has a regression test.

1. **Explicit `return` not covered** (CodeRabbit Major / Codex P1) — `wrap_leading_lets` now takes the dominating prefix slice directly, and `lift_return_stmt_postcondition` routes through it. `let n=x; return n*2;` no longer leaks `n`. Test: `explicit_return_reformat_shares_behavior_identity`.
2. **Chained lets → parallel-SMT unsoundness** (Codex P1) — now emits **nested** single-binding lets (`let a in (let b in ..)`) instead of one multi-binding `(let ((a x)(b (+ a 1))) ..)`, which would leave `b`'s `a` free under SMT-LIB parallel `let`. Test: `chained_leading_lets_emit_nested_single_binding_lets` asserts the nesting.
3. **Shadowing by mutable/unliftable rebind** (Codex P1) — any name bound more than once across the prefix now **bails** (refuses to wrap), so a tail name is never bound to an earlier dead binding. `let n=x; let mut n=x+1; n*2` is (x+1)*2 and must not collapse to x*2. Test: `mutable_shadow_is_refused_no_false_identity`.
4. **Backward walk didn't kill the bound name** (CodeRabbit Minor) — now `needed = (needed \ {name}) ∪ fv(bound_term)` via `needed.remove(&b.name)`. The shadow case it cited bails outright (finding 3). Test: `liftable_shadow_is_refused`.

Full sugar-walk suite green (336 + 167 + integration) on battleaxe.

Separately, the **red CI was inherited from main**, not this PR — the 4 `assertion_vocab_lift.rs` `real_*_javap_*` tests hard-`.unwrap()` on `javap`, absent from CI's `test-rust` phase (no JDK), failing `make ci` identically on main since the "learns REAL JUnit via javap" work. Fixed forward in 117bbe6b1: they skip when `javap` isn't on PATH (still run wherever a JDK exists — all 4 pass on battleaxe), restoring a green gate for every PR.

PR #1993, opened 2026-06-08
Remove aspirational user-facing docs; re-pitch README around behavioral versioning

TSavo:
The README opened with `k(I) = t` (the substrate math) as the first substantive section. That's the *mechanism*, not the *pitch*. This re-leads it.

**Changes (top of README only; everything below the fold is untouched):**

- **Value-first lead.** "Nobody depends on your code. They depend on what your code *does*." Sugar turns the surface you already write into a signed, content-addressed `.proof` of the **behavior**. Keeps "Sugar in, `.proof` out" as the spine.
- **New headline section — "Why it matters: the version lied, the behavior moved, Sugar saw it."** SemVer versions the shadow (bytes); Sugar versions the object (behavior). A reformat reads `none`; a malicious patch under a frozen version reads `new`/`lost`. The effect-row argument: you cannot bolt a credential harvester onto a package for free, because new behavior means new effects and effects can't hide in a fingerprint that records them. `--frozen` / `--require`. Named as the shape of the npm/PyPI supply-chain compromises.
- **Honest scope up front.** Works *today* as `cargo sugar` (Rust) and `sugar-check` (Python pre-commit). The npm/JS wedge is in progress — the lifter is missing, not the thesis. Links `docs/explanation/behavioral-versioning.md`.
- **Demote `k(I)=t`** to "How a `.proof` works: the claim, formally" (kept verbatim, just no longer the lead).
- **Add the missing `sugar diff` verb** to the CLI list, and a behavioral-versioning row to the Start-here nav.
- No em-dashes in the new prose.

Doc-only; no code touched. Separate from the canonicalization work in #1992 (which is what makes the body fingerprint reformat-stable enough to back this pitch).

## Summary by CodeRabbit

* **Documentation**
  * Enhanced README introduction with improved behavioral-verification narrative explaining the proof-supply-chain concept
  * Added documentation for the `sugar diff` command with behavior-based comparison and bump enforcement options
  * Introduced new guide on behavioral versioning and its advantages over traditional semantic versioning


TSavo:
Scope grew (on T's call) from "re-pitch the README" to "remove the user-facing documentation that was written ahead of the implementation." Three commits:

1. **README re-pitch** — leads with "Sugar in, `.proof` out" and the behavioral-versioning / supply-chain catch; demotes `k(I)=t` to "how it works."
2. **Delete 12 fabricated per-language tutorials** — they promised `cargo sugar-lift`, pip/npm/gem installers, and per-language walkthroughs that do not exist.
3. **Delete 71 user-facing docs** — `explanation/`, `how-to/`, `reference/`, `reference-contracts/`, `lsp/`, `quickstart-*`, `index.md`, and the misc product dirs.

**Kept:** the After-X papers, `jokes.md`, and all engineering records (audits, security, self-application, plans, postmortems, contributing, research, baselines).

**Why:** the docs described installers and per-language flows that don't ship. Fiction-ahead-of-implementation is the exact thing the product is built to be debunk-proof against, so it's removed rather than patched. Honest usage docs return when there's a path that runs end to end to document. The README's Start-here section now says this plainly.

**Residual (not in this PR):** ~21 surviving files (mostly `contributing/`) still cross-link the deleted dirs; bounded follow-up sweep.

PR #5915, opened 2026-07-19
Drain #5911: qualified/dotted import-binding-resolvable Call shape

TSavo:
Part of #5911

## Mechanism (no new recognizer code)

The shape re-audit at #5252 judged the 130 single-callee tickets filed under #5911 (775 unclassified factory-walk rows) as ONE mechanism: every member is a `Call` whose callee is reachable through a direct import/from-import lexical binding chain to an already-known vendor/stdlib type or function.

The generic machinery already exists and is untouched by this PR:
- `imported_call_identity` (`callee_universe.py`) resolves the binding chain — import/assignment, alias-aware, shadow-aware, rebind-aware.
- `_PROTOCOL_IMPORTED_SUPPORT` is the empty-by-construction kit overlay (#5618 pattern); it is populated only via `load_imported_callee_protocol` / `load_kit_manifest_file`, never a hard-coded vendor string in production.
- `sugar_lift_py_tests/recognition/kit_manifest.py` + the corpus-mint wiring (#5907/#5908) already load a named, sha256-hashed manifest file into every mint child process.

This PR adds only **data**:
1. 125 new `CalleeUniverseSupport` enum members for the #5911 families (two distinct import paths for the identical construction target collapse onto one member: `PANDAS_TIMEDELTA`, `PANDAS_TIMESTAMP`, `PANDAS_PERIOD`, `PANDAS_INTERVAL`, `PANDAS_ISNA`; every other row keeps its own member).
2. `kit_manifests/pandas_import_binding_5911.json` — an `imported_callee` manifest declaring all 130 FQNs from the #5911 board, loadable via `--kit-manifest` / `SUGAR_KIT_MANIFEST`.
3. `tests/test_kit_manifest_5911_qualified_imports.py` — twins proving the warrant is lexical binding provenance, never spelling.

## Twins

Truthful:
- `import pandas as pd; pd.Timedelta(...)` (dotted attribute form)
- `from pandas._libs.tslibs.timedeltas import Timedelta; Timedelta(...)` (bare-name form, deep private import path — same construction target as above, same enum member, proving the two board rows are one mechanism)
- `import datetime; datetime.datetime(...)` (stdlib)
- `import pandas._libs.json as json_lib; json_lib.ujson_loads(...)` (nested dotted module alias)

Lying (all refute under the SAME loaded protocol):
- aliased lookalike import (`import json as pd; pd.Timedelta(...)`)
- shadowed binding (parameter named `pd`)
- late rebind (`pd = replacement` after the call site)
- bare FQN with no import at all — a hard `FactoryPanic` on the unbound Name (never a silent gap, never a green construction), same class as the existing `numpy.issubdtype` lying twin.

**Verified fails-before/passes-after**: reverting only the production enum addition (`callee_universe.py`) while keeping the manifest + test file breaks 9 of 12 new tests (`KitManifestError: unknown CalleeUniverseSupport member ...` and the truthful/lying assertions), proving the enum data — not happenstance — is what authenticates. Re-applying restores all 12 green. `bin/bpytest tests/test_kit_manifest_5911_qualified_imports.py` — 12 passed on battleaxe's python-unit closure.

Ran the full existing universe-coverage suite (`test_builtin_callee_universe_sugar.py`, `test_universe_coverage_gap.py`) with and without this diff stashed: identical 36 pre-existing failures both ways — confirmed unrelated to this change (not introduced by it).

## Measured drain

A pandas-corpus shard run (`corpus_fatal_triage.py pandas --file-timeout 30 --compact --shard-count 12 --shard-index 0`) was launched on battleaxe to measure real before/after `R_factory_walk_unclassified` with `--kit-manifest kit_manifests/pandas_import_binding_5911.json`. Battleaxe is running two other heavy corpus jobs concurrently per fleet capacity limits, so this is a sharded (not full-corpus) measurement; will post the actual row-count delta as a PR comment once the run completes rather than block this PR on it, per fix-forward practice — the mechanism and its twins are proven at unit level now.

## Member tickets

All 130 FQNs from the #5911 table are declared in the manifest (list in `kit_manifests/pandas_import_binding_5911.json`). Any row whose call site does not structurally resolve through `imported_call_identity` (shadowed, rebound, unbound, or genuinely dynamic binding) stays loud as `FactoryPanic` / unclassified — named, not counted as drained.

`R_vendor_special_case`: unchanged — no vendor string match, name whitelist, or inline isinstance/AST matcher was added; every coordinate arrives only through the kit-manifest contract, which the vendor floor does not scan (same class as #5902-#5907).

Do not merge — coordinator merges after soundness read.

## Summary by CodeRabbit

* **New Features**
  * Expanded recognition support for qualified imports across pandas, NumPy, datetime, regular expressions, and related libraries.
  * Added manifest-based mappings for 130 imported call targets.
  * Improved handling of valid dotted and deep-import forms while rejecting aliases, shadowed names, rebinding, and unimported references.

* **Tests**
  * Added comprehensive coverage for manifest loading, recognition, rejection cases, and clean protocol reset.

> [!NOTE]
> ### Add kit manifest support for 130 qualified/dotted import-binding-resolvable call shapes
> - Adds a [kit manifest JSON file](https://github.com/TSavo/sugar/pull/5915/files#diff-be6c9152154ebc71e5e54abc7400a393f608967819ded09de6238490e1f74bdd) mapping 130 fully-qualified import targets (pandas, numpy, stdlib datetime, dateutil, internal pandas._libs paths) to `CalleeUniverseSupport` enum members.
> - Extends the [`CalleeUniverseSupport` enum](https://github.com/TSavo/sugar/pull/5915/files#diff-72e9f915afa350390348a1c277d0c2047b26a0182bb4780c57baf73f30fdcaa8) with new members for each manifest-driven imported callee identity; distinct import paths can collapse to the same member (e.g. `pandas.Timedelta` and `pandas._libs.tslibs.timedeltas.Timedelta` both map to `PANDAS_TIMEDELTA`).
> - Production tables never embed vendor FQNs directly; recognition resolves them via the manifest-driven `imported_callee` protocol at runtime.
> - Adds a [comprehensive test suite](https://github.com/TSavo/sugar/pull/5915/files#diff-9a27b1c9bd7b861c0b87da96f54f12857af373bd3952232f4f1bb4809cc319ab) covering truthful dotted-attr and bare-name imports, negative cases (aliased lookalikes, shadowed parameters, late rebinds, missing imports), and manifest unload behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 60a151d.</sup>
> 
>


TSavo:
Baseline measured (no manifest loaded): pandas corpus shard 0/12, `--file-timeout 30 --compact` → `R_factory_walk_unclassified = 300` for this shard. With-manifest run (same shard, `--kit-manifest kit_manifests/pandas_import_binding_5911.json`) is running now on battleaxe; will post the delta as soon as it lands.

PR #6292, opened 2026-07-25
Substitution is a discharge: the node a rewrite replaces answers the roll call

TSavo:
## The defect

A node the rewrite replaces constructs nothing of its own, so it sat enrolled on the roster with **no answer at all** — which the minority report read, correctly, as a silent unaccounted construction.

## The fix

Substitution *is* that node's discharge. It answers `present_inert` (showed up, built nothing) and whatever replaced it answers separately for what it builds. **The roster does not shrink.**

Five seams answer this way:

- the two `Param`s their `FormalRef`s replace,
- the two body `Name`s replaced by their bound values,
- the callee `Name` a call site absorbs as a spelling (both the spread and call-site branches).

## The tooth

The test pins the roster **by exact coordinate, not by count** — a "fix" that shrank the minority by shrinking the roster would satisfy a count check and destroy the accounting. It also pins the substituted set itself: remove the mechanism and the minority goes `1 -> 6` with exactly

`Param a(1,8)`, `Param b(1,11)`, `Name(2,11)`, `Name(2,15)`, `Name(6,11)`

coming back, into a roster that is still 16 entries. Any *other* six is a different bug and must not pass here.

## The remaining minority entry

`Module` — the presently unwritten, **explicitly accounted** frontier (`Module.sugar` raises `SugarNotWritten` by design). Commented as such: when `Module.sugar` lands this expectation should turn red, and that red is the good kind.

## Receipts

Own-baseline A/B. **Not gated on CI** — trunk is broadly red independently. Base is stated, not implied current: `origin/main` moved several times while this was measured, and a delta is valid at the base it was measured on.

### `sugar-source-tree` — the package the change lives in

| half | tree | result |
|---|---|---|
| A | mechanism reverted, new test deselected | 583 passed, 5 skipped, 0 failed |
| B | patched | **584 passed, 5 skipped, 0 failed** |

Delta exactly +1, zero regressions.

### `sugar-lift-py-tests` — the consuming suite, no-regression check

Two fresh detached worktrees, stamps pre-resolved under `SUGAR_BINARY_ALLOW_BUILD=0` before either half started.

| | A | B |
|---|---|---|
| `git rev-parse HEAD` | `2566183571cba676…` | `fde816e9bf9d3511…` |
| sourceStamp (before / after) | `blake3-512_8fea7f12…` / identical | `blake3-512_a2daaa35…` / identical |
| package / profile | `sugar-cli` / release | `sugar-cli` / release |
| passed | 1089 | 1090 |
| failed | 129 | 128 |
| skipped / errors | 12 / 5 | 12 / 5 |
| **collected** | **1235** | **1235** |

`1089 + 129 + 12 + 5 = 1235` both sides.

The single differing node — `test_callsite_term_cycle_key_volume.py::test_callsite_add_on_deep_terms_does_not_rebuild_cid_keys` — is a **wall-clock budget assertion** (`budget_seconds = 0.15`), not a behavioural one. It fails on **both bases under both conditions**: inside the sweep, in isolation, at node scope and at file scope. Measured elapsed clusters just above the budget on this machine — `0.189s` in the A sweep, `0.171s` isolated, against a `0.150s` budget.

The lone pass observed in the B sweep is **unexplained and tracked separately**. The attribution does not depend on explaining it: there is no condition in which this test separates the two trees.

**Claim: zero attributable regressions and zero attributable improvements across 1235 collected.** The change is confined to the roll-call accounting proven by the `sugar-source-tree` tooth.

`black --check` clean on both touched files.

### DIFF
--- a/implementations/python/sugar-source-tree/src/sugar_source_tree/nodes.py
+++ b/implementations/python/sugar-source-tree/src/sugar_source_tree/nodes.py
@@ -892,13 +892,30 @@ class Node(Typed):
                     "_substitute_body (a block), never _substitute_children -- "
                     "give the containing node a block-aware substitute"
                 )
+            if new is not value:
+                value.discharge_by_substitution()
             return new, new is not value
         items = tuple(value)
         new_items = tuple(
             item.substitute(scope) if isinstance(item, Node) else item for item in items
         )
         changed = any(new is not old for new, old in zip(new_items, items))
+        if changed:
+            for new, old in zip(new_items, items):
+                if new is not old and isinstance(old, Node):
+                    old.discharge_by_substitution()
         return (new_items if changed else value), changed
+
+    def discharge_by_substitution(self) -> None:
+        """Answer the roll call for a node the rewrite replaced.
+
+        Substitution IS this node's discharge: it constructs nothing of its own,
+        so it answers ``present_inert`` -- showed up, nothing built -- and the
+        node that replaced it answers separately for what it constructs. Without
+        this the replaced node stays registered with no answer at all, which the
+        minority report reads (correctly) as a silent unaccounted construction.
+        """
+        self.reporter.present_inert(self)
 
     def _substitute_children(self, scope: BindingMap) -> "Node":
         """The structural recurse a NON-binding compound opts into: substitute
@@ -2116,6 +2133,10 @@ class FunctionDef(Statement):
         from .shadow import ShadowNode
 
         formal = self._formal_coordinate(parameter, ordinal)
+        # The FormalRef stands in the Param's place for the whole body, so the
+        # Param's discharge is this substitution -- otherwise it registers and
+        # never answers.
+        parameter.discharge_by_substitution()
         return materialize(
             self.unit,
             ShadowNode("FormalRef", parameter.span, (("coordinate", Leaf(formal)),)),
@@ -6722,6 +6743,10 @@ class Call(Expression):
                 )
                 for kw in self.keywords
             )
+            if isinstance(self.func, Name):
+                # Absorbed as a spelling, never constructed -- see the
+                # call-site branch below.
+                self.func.discharge_by_substitution()
             callee_name = self._spread_callee_name(self.func)
             return SpreadCallSugar(
                 callee_name=callee_name,
@@ -6857,6 +6882,11 @@ class Call(Expression):
             )
         if isinstance(self.func, Name):
             from sugar_lift_py_tests.sugar.call_site_sugar import CallSiteSugar
+
+            # The call-site coordinate absorbs the callee's spelling instead of
+            # its sugar, so the callee never constructs. That absorption IS its
+            # discharge; without it the Name registers and never answers.
+            self.func.discharge_by_substitution()
 
             contract_ref = None
             contract_resolution_gap = None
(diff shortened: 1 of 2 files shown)


TSavo (commented):
## Semantic review: substitution-as-discharge (#6292)

**Verdict: mechanism is sound; tooth is almost strong enough; one conservation gap in the instrument.**

### What the law claims
A node the rewrite **replaces** constructs nothing of its own. Leaving it enrolled with no answer is correctly read as silent unaccounted construction. Substitution **is** that node's discharge (`present_inert`); the replacement answers separately. The roster does not shrink.

### Seams reviewed (code)
| Seam | Site | Assessment |
|------|------|------------|
| Generic single-child replace | `_substitute_field` / children path when `new is not value` | Correct: only discharges the **old** node |
| Tuple children | zip old/new when changed | Correct |
| Param → FormalRef | `FunctionDef` materialize | Correct and necessary |
| Body Name → bound value | via generic substitute path | Covered by generic arm |
| Callee Name → CallSite spelling | both spread and non-spread `Call` branches | Correct; Attribute/computed callees correctly **not** discharged here |

Double-discharge: spread and call-site branches are mutually exclusive — OK.

### Tooth (test) — what is pinned well
- **Roster by exact coordinate** (kind, name, line, col) — not count. Prevents "fix R by deleting from roster."
- **Minority is exactly Module** for the all-written fixture — explicit frontier, not permanent zero.
- **Substituted set** listed: Param(1,8), Param(1,11), Name(2,11), Name(2,15), Name(6,11).
- A/B claim: delta +1 twin, zero regressions, revert restores those five to minority.

### Tooth gap — discharge *kind* is not instrumented
`CollectingReporter.present_fact` and `present_inert` both append to the **same** `present` list. The report has no `present_inert` view.

Therefore the test proves:

- those five coordinates **answered present** (not minority), and
- roster length/coords are conserved,

but it does **not** prove they answered via `present_inert` rather than some other present path (e.g. accidental `present_fact` on the old node). The comment about "remove discharge_by_substitution → minority 1→6 with exactly this set" is the right discrimination, but it lives outside the committed test as an A/B procedure.

**Recommended strengthen (non-blocking for land if A/B is re-run on `a0d38afca`):**
1. Track discharge kind on CollectingReporter (`present_inert_cids` vs `present_fact_cids`), **or**
2. Commit an automated revert-check in-test (monkeypatch `discharge_by_substitution` off) that asserts the minority becomes exactly Module ∪ those five coordinates.

Without one of those, roll-call conservation under "substitution is discharge" is still one step short of machine-checked.

### Interaction risks
- **#6291** (panic enum) — no seam overlap; PR claims re-baseline A/B after it. Re-run A/B on **`a0d38afca`** (floors + EffectBoundaryWith + CI durability) before merge.
- **Roll-call suite identity / #6297** — roster must remain coordinate-stable across CI identity; this PR does not shrink roster, which is the right polarity.
- **Module frontier** — when Module.sugar lands, the Module-only minority assert **must** go red. Comment already says so; keep it as a deliberate tripwire, not soft-assert.

### Land criteria I would require
1. A/B re-run on `a0d38afca` (not only `063df6c` / `df40810`): still +1, still zero regressions, still exact five-coordinate set under revert.
2. Optional but preferred: instrument inert-vs-fact or in-test revert discrimination.
3. Do **not** merge on green counts alone.

### Not claimed
- No ΔR on pandas construction/desugar from this PR.
- Does not discharge Module, Delete, or other unwritten constructs.

**Recommendation:** approve after re-baseline A/B on current main tip; prefer the inert-channel tooth before treating roll-call conservation as closed.

PR #7422, opened 2026-08-18
Brand opaque manager producer authority

### LINKED ISSUE #7420: Producer-authority sentinels are bare object(), so 74 frontier files cannot canonicalize
## What the frontier says

Of the 317 construction panics on the sealed pandas board (run `31979878975`, main `eeff6262e`), **115 are canonicalization refusals**, and grouping them by *refused-at field path* rather than by category shows they are not scattered. Two field paths hold 98 of them:

```
  74  .contract_ref._authority of ...with_opaque_cited_manager_sugar.WithOpaqueCitedManagerSugar
  24  .expected_source_call_frame_owner of ...call_site_sugar.CallSiteSugar
```

This issue is the **74**. The `_Handle` escape behind the other 24 is filed separately.

The refusal, verbatim:

> `ConstructedValueCategoryGap: unclassified constructed value category builtins.object: ConstructedValueV2 names its categories and will not invent a preimage by reflection. Name the category, or keep the coordinate loud.`

## The category is `builtins.object` because the value is a bare `object()`

`context_manager_resolution.py:367`

```python
_OPAQUE_CITED_MANAGER_AUTHORITY = object()
```

`context_manager_resolution.py:406`

```python
_authority: object = field(init=False, repr=False, compare=False, default=None)
```

and `__post_init__` immediately closes it:

```python
if self._authority is not _OPAQUE_CITED_MANAGER_AUTHORITY:
    raise ContractRefProtocolError(
        "opaque-cited context-manager ref lacks producer authority"
    )
```

So the field **holds exactly one value, always, enforced at construction** — and is declared `object`, carrying that one value as an anonymous instance of `builtins.object`. The canonicalizer is right to refuse: there is no category to name, and inventing one by reflection is exactly what it must not do.

The invariant is already closed at runtime. It is only the *declaration* that is open. That is the detectable/unrepresentable split: the producer-authority claim is enforced by a guard when it should be carried by a type.

## Why it costs 74 files

`WithOpaqueCitedManagerSugar` is the seat for the off-population `with` head — the shape that says *the citation is authenticated, the semantics are authenticated as absent from our knowledge*. That seat is common in pandas. Every file that reaches it and then reaches canonicalization stops on this field. 74 distinct files, one per row.

Nothing about pandas is responsible for these 74. The refusal is measuring our own declaration.

## The idiom is systemic

`_OPAQUE_CITED_MANAGER_AUTHORITY` is not alone. The same producer-authority sentinel appears as a bare `object()` in at least ten places:

```
conservation_mint.py:23                _WITNESS_AUTHORITY
context_manager_resolution.py:80       _IMPORT_CALL_VALUE_SUBSUMPTION_AUTHORITY
context_manager_resolution.py:367      _OPAQUE_CITED_MANAGER_AUTHORITY
import_binding.py:121                  _IMPORT_AUTHORITY
floor/import_member_value.py:22        _IMPORT_MEMBER_AUTHORITY
floor/decorated_class_value.py:9-12    _APPLICATION_/_PUBLICATION_/_MEMBER_/_METACLASS_AUTHORITY
compose_control_effect_board.py:307    _RELATION_MEMBERSHIP_AUTHORITY
```

Only one of them is on the frontier today. The rest will refuse identically the moment their seats reach canonicalization. Whatever is decided here should be decided for the idiom, not for the one instance.

## The ladder

Rungs, weakest to strongest — the ruling is which rung, not whether:

1. **Brand it.** Give the sentinel a named type whose sole instance is the authority, and declare `_authority: _OpaqueCitedManagerAuthority`. The category becomes nameable; the guard stays.
2. **Remove the seam.** `_authority` is `init=False` and always the same value; at runtime it carries zero information. It exists to prove the producer minted the ref. If the ref type is only constructible through the producer door, the field is redundant and the guard has no subject — delete both.

Rung 2 is the honest one if the producer door can be made the only door. Rung 1 is correct and cheap if it cannot. I have not established which, and that is the first question for whoever takes this.

## Falsifiability

Whatever lands must come with a lying twin: a ref carrying an authority value that is *not* the producer's, which must still be refused after the change. A branded type that any caller can instantiate has moved the category without moving the guarantee, and the suite will go green on it.

## Not in scope

The 193 `With._construct_sugar` "no contract ref" rows are a different wall and stay separate. The 3 `builtins.complex` root refusals are a genuine third thing (a complex literal whose category is unnamed) and are small enough to fold into whichever fix names categories.

## Provenance

Board artifact from run `31979878975`; taxonomy computed over `constructionPanics` (317 rows, one per file). Field declarations read at main `eeff6262e`. Counts are measured, not recalled.

TSavo:
Fixes #7420.

## Ruling

Take rung 1: brand the sentinel and retain the identity guard.

`OpaqueCitedContextManagerRefV1` exposes a generated dataclass constructor. That constructor is refused today only because its non-producer `_authority` reaches the identity check in `__post_init__`; deleting the field and guard would make direct construction succeed. Repository-wide symbol search found one legitimate minter, but Python's public constructor and `object.__new__` mean the producer door is not structurally the only door. Rung 2's premise is therefore not established.

## Change

- Replace the bare `object()` sentinel with a frozen, identity-only `_OpaqueCitedManagerAuthority` category.
- Type `_authority` as that category. The public constructor receives a fresh same-category lie; only the producer overwrites it with the accepted singleton.
- Preserve the runtime identity guard.
- Add a real `WithOpaqueCitedManagerSugar` canonicalization tooth and a same-category non-producer lying twin.

This shape generalizes to the other bare producer-authority sentinels listed in #7420: name each category, preserve singleton identity, and retain its guard unless that ref's constructor is first made structurally producer-only. This PR converts only the opaque-cited manager seat.

## Delta-epsilon

- Predicted epsilon: remove the `builtins.object` constructed-value category gap at `.contract_ref._authority` for this seat.
- Observed focused result: the new canonicalization test first failed at exactly that path, then passed after branding.
- No live pandas corpus recensus was run, so this PR does not claim a current file-count delta.
- Floors preserved: non-producer refs still refuse; direct construction still refuses; no other authority idiom was softened.

## Lying twin

`test_ref_refuses_a_nonproducer_value_of_the_authority_category` constructs a complete ref carrying a fresh value of the new authority category. With the guard present it raises `ContractRefProtocolError`. Temporarily deleting only the guard made the test fail with `DID NOT RAISE`; the guard was then restored and the focused file rerun green.

## Validation

Baseline failing test names before change: `<none>`.

Focused failing test names after change: `<none>`.

The focused file's 16 zero-argument tests were executed serially through a direct Python harness with the three required source roots. The repository's authenticated `bpytest` runner was attempted first, but failed before collection with exit 125 because Docker received a WSL UNC path (`\\wsl.localhost\\...`) rather than an absolute Linux path; that infrastructure failure is not reported as a test failure.

Additional checks:

- `black==26.5.1 --check` on both changed files
- `python3 -m py_compile` on both changed files
- `git diff --check`

A targeted Pyright probe was not a clean receipt: it reported ten diagnostics outside the changed lines, including unresolved `pytest` in the ad-hoc environment and existing attribute/override diagnostics. Per the brief, no full suite or corpus sweep was run locally; CI is the remaining broad telemetry.

> [!NOTE]
> ### Brand opaque manager producer authority with a named dataclass type
> Replaces the bare `object()` singleton used as producer authority in `OpaqueCitedContextManagerRefV1` with a dedicated frozen dataclass `_OpaqueCitedManagerAuthority` (eq=False), preserving identity-based checks while giving the authority a named type.
>
> - The module-level singleton `_OPAQUE_CITED_MANAGER_AUTHORITY` is now an instance of `_OpaqueCitedManagerAuthority` rather than a plain `object()`.
> - `OpaqueCitedContextManagerRefV1._authority` field type is updated to `_OpaqueCitedManagerAuthority`; `__post_init__` still enforces identity against the singleton.
> - New tests verify that a non-singleton instance of the same category type is rejected with `'lacks producer authority'`, and that the authority node produces a blake3-512 CID prefix.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3760db7.</sup>
> 
>

### DIFF
--- a/implementations/python/sugar-lift-py-tests/src/sugar_lift_py_tests/context_manager_resolution.py
+++ b/implementations/python/sugar-lift-py-tests/src/sugar_lift_py_tests/context_manager_resolution.py
@@ -364,7 +364,12 @@ class ContextManagerContractRefV1:
     semantics: ContextManagerSemanticsV1
 
 
-_OPAQUE_CITED_MANAGER_AUTHORITY = object()
+@dataclass(frozen=True, eq=False)
+class _OpaqueCitedManagerAuthority:
+    """Named category whose one producer instance is checked by identity."""
+
+
+_OPAQUE_CITED_MANAGER_AUTHORITY = _OpaqueCitedManagerAuthority()
 
 
 @dataclass(frozen=True)
@@ -404,7 +409,15 @@ class OpaqueCitedContextManagerRefV1:
     target_name: str
     roster: OpaqueSourceCallRosterV1
     citation_cid: str
-    _authority: object = field(init=False, repr=False, compare=False, default=None)
+    _authority: _OpaqueCitedManagerAuthority = field(
+        init=False,
+        repr=False,
+        compare=False,
+        # The public dataclass constructor receives a same-category lie, so it
+        # still reaches the identity guard and refuses. Only the mint door
+        # overwrites this field with the producer singleton below.
+        default_factory=_OpaqueCitedManagerAuthority,
+    )
 
     #: Named so a reader of one row can see the claim without the docstring.
     #: This is testimony, never a bucket key.
--- a/implementations/python/sugar-lift-py-tests/tests/test_opaque_cited_manager_ref.py
+++ b/implementations/python/sugar-lift-py-tests/tests/test_opaque_cited_manager_ref.py
@@ -38,6 +38,7 @@ from sugar_lift_py_tests.sugar.sugar_base import Sugar
 from sugar_lift_py_tests.sugar.with_opaque_cited_manager_sugar import (
     WithOpaqueCitedManagerSugar,
 )
+from sugar_source_tree.binding_state import constructed_value_cid_v2
 
 SOURCE_CID = "blake3-512:" + "aa" * 64
 OWNER_CID = "blake3-512:" + "bb" * 64
@@ -233,6 +234,28 @@ def test_ref_cannot_be_constructed_without_producer_authority():
     assert "lacks producer authority" in str(raised.value)
 
 
+def test_ref_refuses_a_nonproducer_value_of_the_authority_category():
+    """Naming the category must not let a caller mint producer testimony."""
+    from sugar_lift_py_tests.context_manager_resolution import (
+        _OPAQUE_CITED_MANAGER_AUTHORITY,
+    )
+
+    ref = _ref()
+    clone = object.__new__(OpaqueCitedContextManagerRefV1)
+    for name, value in (
+        ("use_site", ref.use_site),
+        ("target_name", ref.target_name),
+        ("roster", ref.roster),
+        ("citation_cid", ref.citation_cid),
+        ("uncited", ref.uncited),
+        ("_authority", type(_OPAQUE_CITED_MANAGER_AUTHORITY)()),
+    ):
+        object.__setattr__(clone, name, value)
+    with pytest.raises(ContractRefProtocolError) as raised:
+        clone.__post_init__()
+    assert "lacks producer authority" in str(raised.value)
+
+
 def test_ref_refuses_a_roster_naming_another_call():
     """Citation and roster must name ONE call; cross-wired testimony refuses."""
     from sugar_lift_py_tests.context_manager_resolution import (
@@ -286,6 +309,11 @@ def test_context_manager_edge_wire_refuses_a_citation():
     with pytest.raises(ContextManagerEdgeTransportError) as raised:
         ContextManagerEdgeDtoV1.from_resolved(ref, ref.use_site)
     assert "authenticated derived ref" in str(raised.value)
+
+
+def test_producer_authority_has_a_named_constructed_value_category():
+    """The real With seat canonicalizes without inventing a reflected category."""
+    assert constructed_value_cid_v2(_node()).startswith("blake3-512:")
 
 
 # --------------------------------------------------- absence is not unknown


TSavo:
Coordinator verification, run independently of the author's evidence.

I ran both sides of the discriminator against the production `constructed_value_cid_v2` door, freshly minting a real `WithOpaqueCitedManagerSugar` seat on each side:

```
SIDE A: branded (this PR)
  authority category: sugar_lift_py_tests.context_manager_resolution._OpaqueCitedManagerAuthority
  receipt-cid bypass: False
  OK cid: blake3-512:d524bbe212807e8...

SIDE B: bare object() sentinel, fresh mint
  authority category: builtins.object
  receipt-cid bypass: False
  REFUSED ConstructedValueCategoryGap: unclassified constructed value category builtins.object
```

The refusal on side B is verbatim the message the sealed board reports at `.contract_ref._authority` for 74 files, and neither side takes the construction-event-receipt bypass. So `test_producer_authority_has_a_named_constructed_value_category` is reaching the real refusal site, not a simulation of it, and the branding is load-bearing.

One trap worth recording for the next person: my first attempt reverted `_authority` on an already-constructed node and got an identical CID on both sides, which reads as a tautology. That was the `constructed_value_cid_v2` memo returning the cached value. The sentinel must be reverted at module scope and the node re-minted, or the negative side is silently vacuous.

The lying twin is genuinely same-category (a fresh `_OpaqueCitedManagerAuthority()`, not the singleton), so naming the category did not weaken the guarantee.

`recensus-path-smoke (PATH integrity)` is red here and is NOT from this diff: it has failed on main on every run back to 2026-08-13, including at this PR's exact base `eeff6262e`. Verified, not assumed.

Not established, and not claimed by the author: the corpus delta. One seat canonicalizing does not prove all 74 files clear, since anything else blocking those files is still in front of them.

PR #7423, opened 2026-08-18
Close raw call-definition handle escape

### LINKED ISSUE #7421: Raw cpython_adapter._Handle escapes into constructed values at 39 frontier files
## What the frontier says

Sibling of #7420, from the same taxonomy of the 317 construction panics on the sealed pandas board (run `31979878975`, main `eeff6262e`).

**39 rows — 39 distinct files — refuse canonicalization because the value at the coordinate is a `sugar_source_tree.cpython_adapter._Handle`**: our own raw parser handle, sitting where a semantic node belongs.

By refused-at field path:

```
  24  .expected_source_call_frame_owner of ...call_site_sugar.CallSiteSugar
   7  .formal_function_sugar.statements[N]...
   4  .source_call_frame.body.statements[N]...
   2  .expected_definition_ref of ...guarded_binding_read_sugar.GuardedBindingReadSugar (nested)
   1  .expected_definition_ref of ...generator_with_sugar.GeneratorWithSugar
   1  .state.when_true.args[0]...
```

Verbatim refusal:

> `ConstructedValueCategoryGap: unclassified constructed value category sugar_source_tree.cpython_adapter._Handle: ConstructedValueV2 names its categories and will not invent a preimage by reflection. Name the category, or keep the coordinate loud.`

`_Handle` is defined at `cpython_adapter.py:62` as a `BackendNode` and is minted throughout the adapter (lines 117, 125, 256, 259, 384, 400, 435, 446, 507, 523). It is a backend detail. It should not be reachable from a constructed value at all.

## This is already known, and the projection already exists

`test_cpython_call_handle_testimony.py:699` says it outright:

> ``expected_definition_ref=bound_frame.owner.ref`` -- a raw parser handle,

and the test then exercises a projection that turns that handle into a `ClassDef` (`:707-716`). So there is a door that converts handle → semantic node, and it is tested. **These 39 files are the seats where that projection does not run.** The question is not how to build the projection; it is why these coordinates bypass it.

## The declaration lets it through

`call_site_sugar.py`

```python
contract_ref: Any = dataclass_field(default=None, compare=False)
expected_source_call_frame_owner: Any = dataclass_field(default=None, compare=False)
expected_definition_ref: object | None = dataclass_field(default=None, compare=False)
```

`Any` and `object | None` admit both the projected node and the raw handle, so nothing at the boundary can tell "not yet projected" from "legitimately absent". That is absence and lookup-failure sharing a representation, in type form.

## And `desugar` reads the difference back out by isinstance

`call_site_sugar.py:200-213`

```python
if isinstance(
    self.expected_definition_ref,
    (FunctionDef, AsyncFunctionDef, ClassDef),
):
    owner_matches = (
        self.source_call_frame.owner.ref is self.expected_definition_ref.ref
    )
else:
    owner_matches = (
        self.source_call_frame.owner.ref is self.expected_definition_ref
    )
```

The comment above it records the history honestly: without the first arm it "fell to the raw-handle branch and compared a typed node against a ref, which never matches." That is the membrane — `desugar` branching on runtime type to recover a distinction the field refused to carry. The `else` arm exists *only* to serve unprojected handles. Delete the handles and the branch has no subject.

## Shape of the fix

Close the field. `expected_definition_ref` / `expected_source_call_frame_owner` should be a two-member reconciling set, not `Any`:

- **projected** — a semantic occurrence (`FunctionDef | AsyncFunctionDef | ClassDef`), which answers `.ref`
- **absent** — a named positive claim that no definition was reached, with its reason

A raw `_Handle` is then unrepresentable at the field, and the 39 seats fail at the *producer* — loudly, naming the coordinate that skipped projection — instead of surviving to canonicalization as an unnameable category. The `isinstance` branch in `desugar` collapses to a match over the closed set.

That reordering is the point: it converts 39 silent late refusals into a small number of loud early ones that name which producers are not projecting.

## Falsifiability

The guard that lands must have a subject that fails only it. Specifically: a `CallSiteSugar` minted with a raw `_Handle` at `expected_definition_ref` must be refused at construction after the change. If the only test is that the 39 files now canonicalize, the check is a tautology — they could canonicalize because the projection ran, or because the category got named, and those are different outcomes with different truth.

Run both sides.

## Provenance

39 rows measured over `constructionPanics` from run `31979878975`; source read at main `eeff6262e`. See #7420 for the other 74 canonicalization rows, which are a distinct cause (bare `object()` authority sentinels) sharing this symptom.

TSavo:
## Summary

- closes both `CallSiteSugar` definition-reconciliation coordinates over projected `FunctionDef` / `AsyncFunctionDef` / `ClassDef` occurrences or a typed positive-absence claim
- moves reconciliation into `Call._construct_sugar` before the value is minted, and retires the shallow late testimony-projection hook
- keeps raw `cpython_adapter._Handle` values loud at the constructor, including the primary `expected_source_call_frame_owner` coordinate
- rejects false positive-absence claims when a source frame or frame table is present, with defensive testimony refusals behind the constructor

The bypass was a shallow post-construction rewrite: `Node.sugar()` invoked `Call._project_constructed_value_for_testimony` only after `CallSiteSugar` had already admitted open `Any` / `object | None` fields. That rewrite handled `expected_definition_ref` and `source_call_frame.owner`, but omitted `expected_source_call_frame_owner` and could not reach raw handles nested inside already-constructed values. Construction now carries the typed semantic occurrence itself, or a named absence reason, so the compatibility branch is gone.

Closes #7421.

## Predicted Epsilon R (construction / wall lanes)

| bucket | predicted Δ | why |
| --- | ---: | --- |
| `constructed` | unmeasured | no pandas frontier recensus was run or claimed; this PR measures early category closure, not canonicalization |
| `mandatory_panics` | unmeasured | remaining raw-handle producers are expected to become fewer, earlier coordinate-named failures |
| `suppressed_descendants` | 0 | no suppression policy changed |
| `typed_effects` | 0 | no effect family changed |
| `silent` | 0 | floor: raw handles cannot become accepted testimony |

## Validation

- Authenticated focused baseline on `eeff6262e`: 23 collected, 11 passed, 12 failed.
- Authenticated focused final: 26 collected, 14 passed, 12 failed.
- The exact failing names are identical before and after:
  - `test_a_cross_file_enrolled_callee_is_not_an_identity_fault`
  - `test_an_unauthenticated_callee_unit_still_refuses_by_its_own_name`
  - `test_an_opaque_callee_keeps_its_own_separate_name`
  - `test_the_frame_site_term_is_keyed_on_the_definitions_unit`
  - `test_the_third_authority_accepts_the_true_cross_file_callee`
  - `test_a_shadowed_same_name_definition_is_REFUSED`
  - `test_a_class_callee_carries_a_derivable_constructor_law`
  - `test_the_third_authority_accepts_the_true_allocation_callee`
  - `test_a_shadowed_same_name_CLASS_is_REFUSED`
  - `test_a_class_resolving_to_a_function_of_the_same_name_is_REFUSED`
  - `test_an_opaque_allocation_callee_still_refuses_by_its_own_name`
  - `test_an_allocation_handle_is_projected_to_its_typed_class_occurrence`
- All 12 are pre-existing `BareConstructionDoor` failures; both new raw-handle teeth and the false-absence twin pass.
- Required lying twin: removing the `expected_definition_ref` constructor guard made `test_call_site_constructor_refuses_raw_definition_handle` fail with `DID NOT RAISE TypeError`; restoring it passed.
- Positive-absence lying twin: before the coherence guard, `test_source_backed_call_refuses_a_false_definition_absence` failed with `DID NOT RAISE TypeError`; after the guard it passed.
- Lift-side blast radius: 88 collected, 84 passed, with the same four pre-existing failing names on base and head.
- `test_module_prefix_definition_reporter.py`: 7 passed.
- Reviewer reproduction confirmed the authenticated method frame CID remains byte-identical after moving CallSite-only reconciliation below the method return.
- `py_compile` on all four changed files and `git diff --check` passed.

## Floors preserved

- `silent=0`: raw parser handles and false absence claims remain loud.
- No test was deleted or weakened to obtain green.
- No corpus canonicalization delta is claimed without a census.
- No secrets or data-path changes.

> [!NOTE]
> ### Close raw call-definition handle escape in `CallSiteSugar` and `nodes.Call`
> - Introduces `DefinitionOccurrenceAbsentV1` and `DefinitionOccurrenceAbsenceReasonV1` in [call_site_sugar.py](https://github.com/TSavo/sugar/pull/7423/files#diff-d7d594636ed207e69db76dd7b89c48760b0d3de93da23981143b3c45da88a3c7) to represent positive absence of a definition occurrence, with strict type validation at construction time.
> - `CallSiteSugar` now rejects raw/unreconciled parser handles for `expected_definition_ref` and `expected_source_call_frame_owner`, and raises `TypeError` if a positive absence is claimed while a corresponding source frame or table is present.
> - `nodes.Call` in [nodes.py](https://github.com/TSavo/sugar/pull/7423/files#diff-de87dc5ad0986c49884d4bc2021f9de6119f8eab49814fe74b7390e1b80214b0) pre-projects the callee's definition occurrence via the new `_project_call_definition_occurrence` helper before minting `CallSiteSugar`, raising `BackendDefect` for untyped or failed projections.
> - The generic `_project_constructed_value_for_testimony` hook on `Node` is removed; projection is now the responsibility of node-specific construction.
> - Behavioral Change: raw parser handles that previously reached `CallSiteSugar` now cause a hard failure at construction time rather than being reconciled later during testimony.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 38b5d66.</sup>
> 
>

### DIFF
--- a/implementations/python/sugar-lift-py-tests/src/sugar_lift_py_tests/sugar/call_site_sugar.py
+++ b/implementations/python/sugar-lift-py-tests/src/sugar_lift_py_tests/sugar/call_site_sugar.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import builtins
 from dataclasses import dataclass, field as dataclass_field
-from typing import Any
+from enum import Enum
+from typing import TYPE_CHECKING, Any, TypeAlias
 
 from sugar_lift_py_tests.context_manager_resolution import SourceFragmentCoordinateV1
 
@@ -13,6 +14,48 @@ from sugar_lift_py_tests.sugar.sugar_base import (
     require_constructed_term_sugar,
 )
 from sugar_lift_py_tests.sugar.witnesses import _call_pair
+
+
+class DefinitionOccurrenceAbsenceReasonV1(str, Enum):
+    """Why a call producer positively claims no source definition occurrence."""
+
+    NOT_SOURCE_RESOLVED = "not-source-resolved"
+    NO_LEXICAL_SOURCE_CALL_ROW = "no-lexical-source-call-row"
+
+
+@dataclass(frozen=True)
+class DefinitionOccurrenceAbsentV1:
+    """Positive absence arm for a call's definition reconciliation fields."""
+
+    reason: DefinitionOccurrenceAbsenceReasonV1
+
+    def __post_init__(self) -> None:
+        if type(self.reason) is not DefinitionOccurrenceAbsenceReasonV1:
+            raise TypeError(
+                "DefinitionOccurrenceAbsentV1.reason must be "
+                "DefinitionOccurrenceAbsenceReasonV1"
+            )
+
+
+if TYPE_CHECKING:
+    from sugar_source_tree.nodes import AsyncFunctionDef, ClassDef, FunctionDef
+
+    DefinitionOccurrenceReconciliationV1: TypeAlias = (
+        FunctionDef
+        | AsyncFunctionDef
+        | ClassDef
+        | DefinitionOccurrenceAbsentV1
+    )
+else:
+    DefinitionOccurrenceReconciliationV1 = object
+
+
+_DEFINITION_NOT_SOURCE_RESOLVED = DefinitionOccurrenceAbsentV1(
+    DefinitionOccurrenceAbsenceReasonV1.NOT_SOURCE_RESOLVED
+)
+_NO_LEXICAL_SOURCE_CALL_ROW = DefinitionOccurrenceAbsentV1(
+    DefinitionOccurrenceAbsenceReasonV1.NO_LEXICAL_SOURCE_CALL_ROW
+)
 
 
 @dataclass(frozen=True)
@@ -48,11 +91,13 @@ class CallSiteSugar(ConstructedTermSugar):
     source_call_frame: Any = dataclass_field(default=None, compare=False)
     source_call_frame_table: Any = dataclass_field(default=None, compare=False)
     source_call_frame_coordinate: Any = dataclass_field(default=None, compare=False)
-    expected_source_call_frame_owner: Any = dataclass_field(default=None, compare=False)
+    expected_source_call_frame_owner: DefinitionOccurrenceReconciliationV1 = (
+        dataclass_field(default=_NO_LEXICAL_SOURCE_CALL_ROW, compare=False)
+    )
     formal_function_sugar: Any = dataclass_field(default=None, compare=False)
     formal_coordinate_cids: tuple[str, ...] = dataclass_field(default=(), compare=False)
-    expected_definition_ref: object | None = dataclass_field(
-        default=None, compare=False
+    expected_definition_ref: DefinitionOccurrenceReconciliationV1 = dataclass_field(
+        default=_DEFINITION_NOT_SOURCE_RESOLVED, compare=False
     )
     native_operation_formal_coordinates: tuple = dataclass_field(
         default=(), compare=False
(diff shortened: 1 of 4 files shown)


TSavo:
Coordinator verification, run independently of the author's evidence.

The author lists `test_an_allocation_handle_is_projected_to_its_typed_class_occurrence` as a pre-existing failure. That is exactly the test covering the projection hook this PR retires, so "pre-existing" is the claim most worth attacking. I checked it on a pristine detached worktree at the base commit rather than trusting the report.

**Base `eeff6262e`, clean checkout:** `12 failed, 11 passed`
**This branch:** `12 failed, 14 passed`

Same twelve names on both sides, all failing with `BareConstructionDoor: ... the tree has none -- it was opened through the bare door`, which has nothing to do with handle escape. Three net new passing tests, zero new failures. The baseline claim holds.

One thing that falls out of this and is worth its own attention: **twelve of the twenty-three tests in `test_cpython_call_handle_testimony.py` are dead on main** — they refuse at the construction door before reaching any assertion. Among the dead is the only test that covers handle→typed-occurrence projection. So this PR retires the late projection hook while its named coverage is unable to run, and the actual evidence for the retirement is the new direct teeth the author added and mutation-proved, not that test. That is acceptable, but it should be stated rather than implied — filing the dead-test file separately.

Nothing here blocks the change.

